PR #3, opened 2015-07-31
combine if check when possible

samsonasik:


ezimuel:
@samsonasik I think my solution is more readable. Thanks anyway for your contribution!

samsonasik:
ok :+1:

Issue #6, opened 2015-08-01
What is container?


bakura10:
Hi,

What is the goal of those lines: https://github.com/zendframework/zend-expressive/blob/master/src/App.php#L93

It seems that the App has no container. Judging from the exception it seems you are using the Interop interfaces. Maybe the container could set as a dependency in constructor?

samsonasik:
@bakura10 I think that indeed a container interop interface. As for zf, we can consume this in new skeleton and inject with ServiceManager? /cc @ezimuel

weierophinney:
@samsonasik This is separate from the zf2 mvc layer.

bakura10:
This is what I understand @weierophinney , but I REALLY love the lightweight architecture in that kind of library, and I'd love to actually use that. However as said on Twitter, I'm really a bit lost about where all this is going, mainly by the lack of visibility of where the Zend team is leading, and how everyting should interact together.

I can definitely see the advantages of that kind of project (Zend-Expressive...) for API driven proejcts, where the Zend\Mvc is really more complex that waht I need. But for this to work it would be awesome to have some more complex example, and how it could be integrated inside the existing ZF modules. For instance: how could I use Zend\Expression with a module like SlmQueue that rely on the ZF config?

I'd love to have the general goal of all this. As I see how it evolves (and tbh I LOVE how it is evolving - aka smaller packages with a very focused goal), the mid-term goal is to kinda deprecate usage of Zend\Mvc (and the router), and use a third-party router (I don't see Dasprid those times so I don't really know where his Dash router is going), and create a new, minimal, decoupled Zend\ModuleManager that could work with it.

If we take advantage of the Container Interop, PSR7... all of this could work smoothly, and I'd be happy to help to write some code.

samsonasik:
we don't know instead of guess-ing as the repository already public, what the plan for its usage? it's regarding with zf3 implementation?

weierophinney:
In the ZF3 roadmap post, we indicated we'd provide a middleware-based
application solution in *parallel* with the current MVC. This is the start
of that. *It doesn't replace the MVC.* (And it will potentially require its
own skeleton, which was the first of my original reply to you.)

I'll provide a more full explanation next week so you don't have to keep
guessing. In the mean time, consider it a work in progress, and maybe sit
back to see how we flesh it out before diving too deep into the code. (I
have a considerable refactor that's based around use cases almost ready to
post, and it will likely shed more light on our intentions.)

[...]

Issue #4, opened 2015-07-31
Use PSR7 interfaces in match


bakura10:
Hi,

I've just had a quick look at the code, and it seems the `match` could accept a ServerRequestInterface instead (https://github.com/zendframework/zend-expressive/blob/master/src/Router/RouterInterface.php#L19). That would make things easier and a great way to take advantage of those interfaces from start to end. It seems that next version of Aura.Router that you rely on supports this directly: https://github.com/auraphp/Aura.Router/blob/3.x/src/Matcher.php#L110

For adapter that do not support it (like FastRoute), a transformation should be done before passing the info.

ezimuel:
@bakura10 thanks for your comment! The goal of the route adapter is to offer an easy way to integrate external routing system, even if not PSR-7 compliant.  I think the interface `match($path, $params)` is more general then `match(ServerRequestInterface $request)` and offers an easy way to integrate also PSR-7 using the following code:
```php
$path =  $request->getUri()->getPath();
$params = $request->getServerParams();
```

bakura10:
Hi,

Seen otherwise, the transform could be made in the match for library that do not support it (untested code, just the idea):

Example in Aura that supports passing PSR7 interface:

```php
/**
     * @param  string $patch
     * @param  array $params
     * @return boolean
     */
    public function match(ServerRequestInterface $request)
    {
        $this->route = $this->router->match($request);
        return (false !== $this->route);
    }
```

For those that do not support it: 

```php
/**
     * @param  string $patch
     * @param  array $params
     * @return boolean
     */
    public function match(ServerRequestInterface $request)
    {
        $request = ServerRequestFactory::fromGlobals();

        $dispatcher = new Dispatcher($this->router->getData());
        $result     = $dispatcher->dispatch($request->params(), $request->path());
        if ($result[0] != Dispatcher::FOUND) {
            return false;
        }
        $this->routeInfo = $result;
        return true;
    }
```

In short terms most library should comply with PSR7 :).

ezimuel:
@bakura10 we discussed with @weierophinney and we decided to use the the PSR-7 for the routing interface. We did an entire refactor of the implementation. You can see it [here](https://github.com/zendframework/zend-expressive/blob/master/src/Router/RouterInterface.php) and thanks for your feedback!

PR #7, opened 2015-08-03
Refactor and merge ideas

weierophinney:
This pull request is the result of discussions with @ezimuel and from a code review of the original ezimuel/zend-stratigility-dispatch code. It accomplishes the following:

Middleware
----------

- Allow re-use of the Application instance as middleware (already part of the original).
- Allows piping other middleware into the Application instance (as post-processing middleware and/or error handling middleware; this feature is new to this branch).
  - This required making the routing aspect of invocation a dedicated middleware method, `routeMiddleware`.

Routing
-------

- Adds a `Route` class to encapsulate the path, middleware, HTTP methods, and any options that might further configure the route when passed to the underlying routing library.
- Refactors the `RouterInterface`:
  - to accept route configuration as `Route` instances.
  - to accept a PSR-7 `ServerRequestInterface` to `match()`, which will allow routers to pull whatever they need from the request environment in order to perform their work.
  - to add statelessness; `match()` now returns a `RouteResult` with information on success/failure of the match operation, the matched path and middleware, and, if failure was due to HTTP negotiation, the allowed HTTP methods.
  - the above required refactoring the Aura and FastRoute implementations.

Emitters
--------

- We now define an `EmitterStack` for providing a stack of Emitters; first one to return a non-boolean false result halts the stack. More on this when I cover `Application::run()`.

Application
-----------

- Renamed `App` to `Application`. We can change this if desired; I started with `Application` to prevent collisions with the original.
- `addRoute()` became `route()`, and accepts the path, middleware, and HTTP methods (which are optional; if not provided, all methods are assumed); it returns a Route instance. You may also pass a Route instance to it.
- `__call()` proxies to `route()`, and raises a `BadMethodCallException` if the HTTP method is unknown. Only a small list of allowed HTTP methods is provided; anything else can be added via the `route()` method (which accepts arbitrary HTTP methods).
- Application composes a "final handler" callable, defaulting to a `Zend\Stratigility\FinalHandler` instance if none is provided. This is used if the `__invoke()` method is called without an `$out` argument.
- Application composes a `Zend\Diactoros\Response\EmitterInterface`, defaulting to an `EmitterStack` with an `SapiEmitter` composed if none is provided.
- `run()` now optionally accepts `ServerRequestInterface` and `ResponseInterface` instances. It will create instances if none are provided.
- `run()` now invokes itself directly, passing itself the request and response instances, instead of composing itself in a `MiddlewarePipe` to allow registering error handlers, and executing via a `Zend\Diactoros\Server` instance.
- `run()` now passes the returned response to the composed emitter to emit it.

The changes to `run()` essentially allow us to have error handling by default, and to allow flexibility in how responses are rendered; a response composing a theoretical `GeneratorStream` or `IteratorStream` might emit the response differently than the default `SapiEmitter` composed in `FinalHandler` would. Adding special-purpose emitters via a stack gives us that flexibility where it is missing in the Diactoros and Stratigility implementations.

## AppFactory and Container\ApplicationFactory

Dependencies such as the router are now required to have a functional Application instance. To provide a solid out-of-the-box experience, this pull requests adds `AppFactory::create()`. The factory method can be called without arguments, in which case it will use the Aura router by default and the ZF2 ServiceManager. You can pass the container and router, respectively, as optional arguments to configure the Application instance.

The `Container\ApplicationFactory` class is invokable, and accepts a container-interop instance. It will attempt to pull instances by service name (using FQCN or fully-qualified interface names) to configure an Application instance, using sane defaults for any missing services. It also provides the ability to pull routing configuration and use that to inject routes into the Application instance.

Final Notes
-----------

The patch provides architectural notes via the [NOTES.md](https://github.com/weierophinney/zend-expressive/blob/feature/refactor/NOTES.md) file, and also provides usage examples via [bookdown documents](https://github.com/weierophinney/zend-expressive/tree/feature/refactor/doc/book).

This pull request should address #4 and #6.


bakura10:
Hi,

That sounds awesome, and I didn't have time yet to fully read your documents, but stupid question (that may be more related to middleware). I can definitely see some analogy between modules and middlewares (from what I understand). A middleware, as a module, provide functionalities.

However, it seems that middleware is more "route oriented". Can a middleware only provide features without giving any route?

My other question is that contrary to modules that seems to "integrate" into the request workflow, it seems that middlewares are completely independant one of each other, and don't know what happen in the outside world. For instance, I could I attach an event executed at the end of the request, or before any middleware execute? Should I create a new middleware and position at the right position in the request flow?

I really like the approach and what I've seen from the examples, it seems to make everything much more simple and natural, and I like the route-oriented approach, but I'm having a hard time understanding how that maps into a bit more complex application, and how an external library could plug into various steps of the requests.

Once again, sorry if you have answered to those questions in a conference/blog post/documentation, I have to admit I didn't really invested time on understanding middleware.

weierophinney:
> I can definitely see some analogy between modules and middlewares (from what I understand). A middleware, as a module, provide functionalities.

You're thinking in terms of ZF2. Middleware is code that sits between a request and a response. Middleware is often nested or executed in a pipeline; as such, you can compose middleware in order to accomplish complex functionality.

It's similar to a ZF2 module, but can be both narrower and broader in scope, based on how you write it.

I suggest reading [a blog post I wrote on the subject earlier this year](https://mwop.net/blog/2015-01-08-on-http-middleware-and-psr-7.html) for background.

zend-expressive does the following:

- It provides a middleware implementation that provides routing. Stratigility, on which it is based, does not provide full routing capabilities (like Connect, on which it was based, it provides purposely *limited* routing, primarily to allow segregating middleware by path, and thus allowing lazy-evaluation).
- It is itself a middleware pipeline (an instance of `Zend\Stratigility\MiddlewarePipe`), allowing composition of middleware. 

Which leads us to:

> However, it seems that middleware is more "route oriented". Can a middleware only provide features without giving any route?

You can pipe non-route-based middleware using the `pipe()` method, which is inherited from `Zend\Stratigility\MiddlewarePipe`. Anything added via `route()` (or the various method overloads that proxy to `route()`) will go through zend-expressive's routing middleware. I actually describe this in the documentation: https://github.com/zendframework/zend-expressive/blob/master/doc/book/usage-examples.md#middleware-that-executes-on-every-request

Right now, we have an order-of-operations problem, however: we pipe the routing middleware at instantiation, which means you can only reliably add error handling middleware via `pipe()` (though other middleware will execute if the routing middleware fails to route).

I'm currently working on a pull request that will make custom pipelines easier.

> My other question is that contrary to modules that seems to "integrate" into the request workflow, it seems that middlewares are completely independant one of each other, and don't know what happen in the outside world. For instance, I could I attach an event executed at the end of the request, or before any middleware execute? Should I create a new middleware and position at the right position in the request flow?

Again, you're thinking along the lines of the ZF2 MVC.

**This is not a replacement for the ZF2 MVC; it is a complementary and parallel offering.**

As such, it's not event-driven like the ZF2 MVC. Your last assumption is correct: if the order in which middleware should execute is important, then you should `pipe()` or compose your middleware accordingly.

Regarding composition, I mean something like this:

```php
$blog = AppFactory::create();
// add routes/middleware to it.

$api = AppFactory::create();
// add routes/middleware to it.

$app = AppFactory::create();
$app->pipe('/api', $api);
$app->pipe('/blog', $blog);
```

In the above, each of `$blog` and `$api` are middleware that could execute on their own; they likely compose other middleware, either via routing or piping. We then *compose* them into our main application, segregating them to sub-routes; inside those nested middleware, routing will be relative to the path provided, so *no changes to routing are required*. As such, we can mix and match middleware into an application, and they'll work no matter how far nested they are.

Hope the above and the provided links help clarify things, @bakura10 !

bakura10:
Thanks for the answer. That definitely makes things clearer and I'll actually try to transform my ZfrOauth to a middleware. What I like from what I understand is that consumer could attach the middleware at the beginning of their application, and having the full Oauth without having to initialize/execute other middleware. Seems awesome for performance ^^.

ojhaujjwal:
@bakura10 +1 about transforming `zfr-oauth2-server` to a middleware. That idea had also popped to my head a few days ago. :D

PR #14, opened 2015-08-04
add exception message to Exception\ContainerNotRegisteredException

samsonasik:


weierophinney:
No need; the exception name is enough.

PR #15, opened 2015-08-04
Delay routing middleware registration to allow pipelines

weierophinney:
This patch alters the workflow of `Application` to delay registration of the routing middleware until the first call to `route()` (which may be via one of the proxy calls via `__call()`). This allows sculpting a full middleware pipeline with early interceptors and/or piping to subpaths.

Additionally, the `Container\ApplicationFactory` implementation was updated to allow specifying middleware for the pipeline. To accomplish this, it uses a new configuration key, `middleware_pipeline`, with up to two subkeys, `pre_routing` and `post_routing`. Each specifies an ordered array of middleware or services resolving to middleware; `pre_routing` middleware are registered prior to any routing middleware, and `post_routing` afterwards.

Finally, if services names are used for `middleware_pipeline` entries, and the service is known by the container, the factory will register a closure:

```php
function ($request, $response, $next = null) use ($services, $middleware) {
    $invokable = $services->get($middleware);
    if (! is_callable($invokable)) {
        throw new Exception\InvalidMiddlewareException(sprintf(
            'Lazy-loaded middleware "%s" is not invokable',
            $middleware
        ));
    }
    return $invokable($request, $response, $next);
};
```

This implements lazy-loading for such services, delaying retrieval from the container until the middleware is actually used. This also means that if the retrieved service is not valid middleware, you will not find out until runtime.

### DIFF
--- a/doc/book/usage-examples.md
+++ b/doc/book/usage-examples.md
@@ -449,6 +449,76 @@ $app->pipe($anErrorHandler);
 $app->run();
 ```
 
+### Using the container to register middleware
+
+If you use a container to fetch your application instance, you have an
+additional option for specifying middleware for the pipeline: configuration:
+
+```php
+return [
+    'routes' => [
+        [
+            'path' => '/path/to/match',
+            'middleware' => 'Middleware Service Name or Callable',
+            'allowed_methods' => [ 'GET', 'POST', 'PATCH' ],
+            'options' => [
+                'stuff' => 'to',
+                'pass'  => 'to',
+                'the'   => 'underlying router',
+            ],
+        ],
+        // etc.
+    ],
+    'middleware_pipeline' => [
+        'pre_routing' => [
+            'middleware services',
+            'to register BEFORE',
+            'routing',
+        ],
+        'post_routing' => [
+            'middleware services',
+            'to register AFTER',
+            'routing',
+        ],
+    ],
+];
+```
+
+The key to note is `middleware_pipeline`, which can have two subkeys,
+`pre_routing` and `post_routing`. Each accepts an array of middlewares to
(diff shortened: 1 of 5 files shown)


doc/book/usage-examples.md
```diff
@@ -422,32 +422,136 @@ along to the next by calling `$next()`.
...
+$app->route(/* ... */);
+$app->pipe($anErrorHandler);
+
+$app->run();
+```
+
+### Using the container to register middleware
+
+If you use a container to fetch your application instance, you have an
+additional option for specifying middleware for the pipeline: configuration:
```
samsonasik: > additional option for specifying middleware for the pipeline. Use the following configuration:

test/Container/ApplicationFactoryTest.php
```diff
@@ -200,4 +200,489 @@ public function testWillUseSaneDefaultsForOptionalServices()
...
+            ->has('Zend\Diactoros\Response\EmitterInterface')
+            ->willReturn(true);
+        $this->container
+            ->get('Zend\Diactoros\Response\EmitterInterface')
+            ->will(function () use ($emitter) {
+                return $emitter->reveal();
+            });
+
+        $this->container
+            ->has('Zend\Expressive\FinalHandler')
```
samsonasik: should be <code>Zend\Stratigility\FinalHandler</code> if #18 will be merged

weierophinney: I suggest leaving as-is until I have a decision on #18. If we decide not to make that change, I don't want to have to revert, and there are reasons we may not make that change due to planned features.

samsonasik: ok :)

PR #17, opened 2015-08-05
Update NOTES.md for SM config usage

samsonasik:


samsonasik:
it should be merge-able ;)

weierophinney:
Those notes are, as noted at the top *outdated*, and really only exist as an historical record of decisions made. I fully intend to remove them before we go stable.

As such, not going to merge; those examples are incorrect anyways.

PR #18, opened 2015-08-05
change service naming: Zend\Expressive\FinalHandler to Zend\Stratigility\FinalHandler

samsonasik:
as <code>FinalHandler</code> class is in <code>Zend\Stratigility</code> components.

### DIFF
--- a/src/Container/ApplicationFactory.php
+++ b/src/Container/ApplicationFactory.php
@@ -23,7 +23,7 @@ use Zend\Expressive\Router\Route;
  *
  * - Zend\Expressive\Router\RouterInterface. If missing, an Aura router bridge
  *   will be instantiated and used.
- * - Zend\Expressive\FinalHandler. The service should be a callable to use as
+ * - Zend\Stratigility\FinalHandler. The service should be a callable to use as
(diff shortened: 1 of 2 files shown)


weierophinney:
I'm going back and forth on this. On the one hand, yes, the class is defined in Stratigility. On the other, we may very well provide our own final handler before we're done (to provide templated error pages). Still mulling it over.

samsonasik:
ok, just note, if we will use own final class, then <code>Application::getFinalHandler()</code> should return <code>Zend\Expressive\FinalHandler</code> too ;)

weierophinney:
> ok, just note, if we will use own final class, then Application::getFinalHandler() should return Zend\Expressive\FinalHandler too ;)

It doesn't need to, actually. It just needs to return a `callable` to use for when the middleware stack is exhausted; `Zend\Stratigility\FinalHandler` is just a default implementation, but it could even be a closure (we use closures in many of the tests). That's the other reason why the service name is largely irrelevant.

weierophinney:
Closing. FinalHandler may be a callable. As such, the service name is not and will not have a 1:1 mapping to an existing class.

PR #20, opened 2015-08-05
Fixed @param typo in __constructor(), fixed parameter check in Application::route()

Stanimirdim92:
If you have a look at Application::route() documentation you will see that the $path can be either a string or an object with Router\Route instance. However, at line 220 the parameter check was only to see if $path has an instance of Router\Route.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -79,7 +79,7 @@ class Application extends MiddlewarePipe
      * to set internal properties. On completion, pipes its own
      * `routeMiddleware()` method to its internal pipeline.
      *
-     * @param Dispatcher $dispatcher
+     * @param Router\RouterInterface $router
      * @param null|ContainerInterface $container IoC container from which to pull services, if any.
      * @param null|callable $finalHandler Final handler to use when $out is not
      *     provided on invocation.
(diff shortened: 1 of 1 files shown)


src/Application.php
```diff
@@ -213,11 +213,11 @@ public function routeMiddleware(ServerRequestInterface $request, ResponseInterfa
...
      * @param null|array $methods HTTP method to accept; null indicates any.
      * @return Router\Route
-     * @throws InvalidArgumentException if $path is not a Router\Route AND middleware is null.
+     * @throws InvalidArgumentException if $path is not a Router\Route|string AND middleware is null.
      */
     public function route($path, $middleware = null, array $methods = null)
     {
-        if (! $path instanceof Router\Route && null === $middleware) {
+        if ((! is_string($path) || ! $path instanceof Router\Route) && null === $middleware) {
             throw new InvalidArgumentException(sprintf(
```
samsonasik: there are options:
- use <code>InvalidArgumentException</code> and add to use statement
- use <code>Exception\InvalidArgumentException</code> as already in Exception namespace

I prefer 2nd approach. Also, it requires additional tests for <code>setExceptedException</code>.

weierophinney: This refactor is incorrect. These are the possible signatures:

- `route(Router\Route $route)`
- `route(string $route, $middleware, array $methods = null)`

This is why the conditional looked for the combination of "not a Route instance AND middleware is missing".

As for the exception when `$path` is not a string, that is raised by the `Router\Route` instance itself when we attempt to instantiate one at [line 279](https://github.com/zendframework/zend-expressive/blob/master/src/Application.php#L279) (see [lines 63-65 of the Route class](https://github.com/zendframework/zend-expressive/blob/master/src/Router/Route.php#L63-L65)).

I also already have tests in place for these behaviors.

weierophinney:
Closing. I've opened #22 for the change to the constructor docblock (more changes were necessary), and the refactor on the `route()` method is unnecessary.

src/Application.php
samsonasik: yes, but the InvalidArgumentException itself detected as never imported in travis after the change. while the change not valid, the InvalidArgumentException usage is need to be fixed.

samsonasik: yes, but the InvalidArgumentException itself detected as never imported in travis after the change. while the change not valid, the InvalidArgumentException usage is need to be fixed.

weierophinney: I have tests in `Zend\Expressive\Router\Route`, but not in `Zend\Expressive\Application`, it seems. I've added some now in #24.

PR #19, opened 2015-08-05
minimum php version requirement in composer.json

weierophinney:
in composer.json, currently there is no minimum php version requirement yet.


weierophinney:
Converted this issue to a PR with the fix.

PR #23, opened 2015-08-05
update InvalidArgumentException usage in Application::route()

samsonasik:
per my comment in https://github.com/zendframework/zend-expressive/pull/20/files#r36340269


samsonasik:
ping @weierophinney

samsonasik:
I've added unit tests for it.

weierophinney:
@samsonasik Saw your comment on #20 first, and created and merged #24 in response already!

samsonasik:
so fast :+1:

PR #26, opened 2015-08-05
Provide documentation around injecting Aura.Router

weierophinney:
Provides documentation around creating your own Aura.Router instance, configuring it, and injecting it into your router instance for zend-expressive. Both programmatic and IoC approaches are detailed.

### DIFF
--- a/doc/book/router.aura.md
+++ b/doc/book/router.aura.md
@@ -36,7 +36,7 @@ The answer, then, is to use dependency injection. This can be done in two ways:
 programmatically, or via a factory to use in conjunction with your container
 instance.
 
-## Programmtic Aura.Router Creation
+## Programmatic Aura.Router Creation
 
 To handle it programmatically, you will need to setup the Aura.Router instance
 manually, inject it into a `Zend\Expressive\Router\Aura` instance, and inject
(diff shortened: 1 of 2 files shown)


doc/book/router.aura.md
```diff
@@ -0,0 +1,172 @@
...
+use that when creating your `Application` instance.
+
+```php
+<?php
+use Aura\Router\RouterFactory;
+use Zend\Expressive\AppFactory;
+use Zend\Expressive\Router\Aura as AuraBridge;
+
+$auraRouter = (new RouterFactory())->newInstance();
+$auraRouter->setSecure(true);
```
bakura10: Are you sure about this? It does not seem to work here. The setSecure is defined in the Aura\Route, not the Router.

doc/book/router.aura.md
```diff
@@ -0,0 +1,172 @@
...
+$auraRouter->addValues([
```
bakura10: Same.

weierophinney: Pretty sure on both: https://github.com/auraphp/Aura.Router#default-route-specifications details using these on the `$router` instance to set defaults.

doc/book/router.aura.md
harikt: Aura.Router have `setSecure` method . See https://github.com/auraphp/Aura.Router#extended-route-specification

doc/book/router.aura.md
harikt: aah @weierophinney have already commented on this. :+1:  .

PR #28, opened 2015-08-06
Ensure that FinalHandler gets original response

weierophinney:
`FinalHandler` allows passing a `ResponseInterface` instance at instantiation. When invoked without an error, it then compares the received response with its original response to determine whether or not to raise a 404; if they are the same, it raises the 404, otherwise, it returns the received response.

`Application` was not passing the response to the `FinalHandler`, which broke this semantic. This patch does the following:

- adds a new optional argument to `getFinalHandler()`, a response instance. If the method creates a `FinalHandler,` it passes that instance to it.
- updates `__invoke()` to pass the response to `getFinalHandler()` if/when it calls that method.

### DIFF
--- a/test/ApplicationTest.php
+++ b/test/ApplicationTest.php
@@ -265,6 +265,35 @@ class ApplicationTest extends TestCase
         $this->assertSame($finalResponse, $test);
     }
 
+    public function testFinalHandlerCreatedAtInvocationIsProvidedResponseInstance()
+    {
+        $routeResult = RouteResult::fromRouteFailure();
+        $this->router->match()->willReturn($routeResult);
+
+        $app = new Application($this->router->reveal());
+
+        $finalResponse = $this->prophesize('Psr\Http\Message\ResponseInterface');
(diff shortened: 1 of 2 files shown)


test/ApplicationTest.php
```diff
@@ -265,6 +265,35 @@ public function testFinalHandlerIsUsedAtInvocationIfNoOutArgumentIsSupplied()
...
     }
 
+    public function testFinalHandlerCreatedAtInvocationIsProvidedResponseInstance()
+    {
+        $routeResult = RouteResult::fromRouteFailure();
+        $this->router->match()->willReturn($routeResult);
+
+        $app = new Application($this->router->reveal());
+
+        $finalResponse = $this->prophesize('Psr\Http\Message\ResponseInterface');
```
bakura10: What about using ::class in your tests too ? :) At least you will get refactoring everywhere

weierophinney: I'm mocking interfaces in most cases, and, as such, don't have to import them as I'm not consuming them directly. :)

PR #29, opened 2015-08-06
Use class

bakura10:


src/Container/ApplicationFactory.php
```diff
@@ -119,16 +122,16 @@ class ApplicationFactory
...
     {
-        $router = $services->has('Zend\Expressive\Router\RouterInterface')
-            ? $services->get('Zend\Expressive\Router\RouterInterface')
+        $router = $services->has(RouterInterface::class)
+            ? $services->get(RouterInterface::class)
             : new AuraRouter();
 
-        $finalHandler = $services->has('Zend\Expressive\FinalHandler')
-            ? $services->get('Zend\Expressive\FinalHandler')
+        $finalHandler = $services->has(FinalHandler::class)
```
weierophinney: In this case, no; the final handler does not need to be an instance of `Zend\Stratigility\FinalHandler`; it can be a simple `callable`. This is the reason why the tests fail currently.

The reason I used `Zend\Expressive` for the namespace is because we have some plans to offer an alternate final handler solution that will provide templated error pages. However, again, we wouldn't want to typehint on that class, as any callable is allowed in the signature that consumes the service.

In other words, it's a service name that doesn't map 1:1 with a class.

bakura10:
Oops... it seems like a good reason to use `Zend\Stratigility\FinalHandler` instead of `Zend\Expressive\FinalHandler` ;).

weierophinney:
@bakura10 You'll need to rebase; I've merged most of your PRs at this point!

bakura10:
grrr, seems I still suck at rebasing...

samsonasik:
@bakura10 you can do following steps:
- git checkout master && git pull git://github.com/zendframework/zend-expressive.git master
- git checkout class && git rebase master

after that, if you have conflict, you can do:
- fix conflict <<<HEAD
- don't add new commit
- git add .
- git rebase --continue

And last, do:
- git push --force origin class

weierophinney:
@bakura10 I took care of the conflicts and got it merged. Thanks!

PR #31, opened 2015-08-06
Implement a ZF2 router implementation

weierophinney:
This patch implements a router that consumes `Zend\Mvc\Router\Http\TreeRouteStack` to provide routing capabilities. HTTP method negotiation is kept internal to the zend-expressive implementation, as the ZF2 router does not differentiate between failure to match the path versus the HTTP method.

Matching marshals a zend-http request using [zend-psr7bridge](https://github.com/zendframework/zend-psr7bridge).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,9 @@
     "require-dev": {
         "nikic/fast-route": "^0.6.0",
         "phpunit/phpunit": "^4.7",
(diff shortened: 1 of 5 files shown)


weierophinney:
@ezimuel I've now implemented the lookup table; good catch!

PR #42, opened 2015-08-07
Windows friendly built-in server command

codeliner:
Using the command from the docs to start the build-in webserver gave me two errors on windows:

first error:
```
$ php -S 0:8080 -t public/
Directory public/ does not exist.
```

second error:
```
$ php -S 0:8080 -t public
<br />
<b>Warning</b>:  Unknown: php_network_getaddresses: getaddrinfo failed: 
Failed to listen on 0:8080 (reason: php_network_getaddresses: getaddrinfo failed: ... )
```

This PR fixes the docs so that windows users can enjoy zend-expressive from the first moment on too ;).

### DIFF
--- a/doc/book/usage-examples.md
+++ b/doc/book/usage-examples.md
@@ -28,7 +28,7 @@ We assume also that:
 > You can use the built-in web server to run the examples. Run:
 >
 > ```bash
-> $ php -S 0:8080 -t public/
+> $ php -S 0.0.0.0:8080 -t public
 > ```
 >
 > from the application root to start up a web server running on port 8080, and


weierophinney:
Better option:

```sh
$ php -S 0.0.0.0:8080 -t public
```

(which is what we use in Apigility documentation). Windows cannot resolve `0:8080`, but *can* resolve `0.0.0.0:8080`; this solution is better than specifying `localhost`, as it then allows clients on other IPs to reach the host, using arbitrary hostnames (assuming that either is allowed by the host's firewall).

codeliner:
@weierophinney Updated the PR based on your comment :+1:

PR #44, opened 2015-08-08
Fix illegal type given when any method is allowed

codeliner:
Tried to add a route to the config similar to the one added in the ApplicationFactoryTest.
Running the modified test without the fix will give you the error:

```
There was 1 error:

1) ZendTest\Expressive\Container\ApplicationFactoryTest::testFactorySetsUpRoutesFromConfig
Argument 3 passed to Zend\Expressive\Application::route() must be of the type array, integer given
```

Not sure if my fix is the best solution. The other way would be to let the `route` method except the value of `Router\Route::HTTP_METHOD_ANY` instead of `null`. However, this was the easier fix :smile:.

### DIFF
--- a/src/Container/ApplicationFactory.php
+++ b/src/Container/ApplicationFactory.php
@@ -162,7 +162,7 @@ class ApplicationFactory
 
             $methods = (isset($spec['allowed_methods']) && is_array($spec['allowed_methods']))
                 ? $spec['allowed_methods']
-                : Route::HTTP_METHOD_ANY;
+                : null;
             $route = $app->route($spec['path'], $spec['middleware'], $methods);
 
             if (isset($spec['options']) && is_array($spec['options'])) {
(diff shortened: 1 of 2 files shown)


codeliner:
@weierophinney fixed the same bug #46. PR is no longer valid.

weierophinney:
Sorry about that! Came across it myself building an example app today, and
fixed it... without first checking to see if it had already been reported!

codeliner:
@weierophinney Same on my side (currently refactoring my cargo-sample to use zend-expressive, want to release the new version tomorrow), except that I was faster with pushing the PR :smile:. But no problem. It was only a small fix. Pulled in your changes and it works. That's the important thing.

Btw when will the first releaseof zend-expressive be available?

weierophinney:
Hopefully this coming week; we just have a few more features and changes
before we can tag the first version.

PR #52, opened 2015-08-09
Fixes zend-psr7bridge composer decription

samsonasik:


ojhaujjwal:
:-1:

samsonasik:
why :-1: ?

ojhaujjwal:
I think that if somebody wants to use `Zend\Expressive\Router\Zf2`, they should know that they need to install package `zendframework/zend-psr7bridge`. I think it is good now.

samsonasik:
ok ;)

PR #49, opened 2015-08-08
add unit tests for null param to __construct in Router

samsonasik:

PR #54, opened 2015-08-10
change $zf2Router to $router Zf2 router property for consistency

samsonasik:
with other routers


samsonasik:
if it get merged, #49 need to be updated.

weierophinney:
I named it intentionally, to draw attention to the fact that it belongs to the underlying implementation.

PR #55, opened 2015-08-10
Added the Template interface + plates and twig adapters

ezimuel:
This PR adds a Template interface and the adapters for [Plates](https://github.com/thephpleague/plates) and [Twig](https://github.com/twigphp/Twig) templating systems.

**Note:** we should think how to offer a template into a middleware callback. One solution can be to use a Factory class for middleware invokable components, like the one reported [here](https://github.com/ezimuel/zend-stratigility-skeleton/blob/master/app/src/Action/HomepageFactory.php).

### DIFF
--- a/src/Template/Plates.php
+++ b/src/Template/Plates.php
@@ -1,0 +1,74 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
+ * @copyright Copyright (c) 2015 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   https://github.com/zendframework/zend-expressive/blob/master/LICENSE.md New BSD License
+ */
+
+namespace Zend\Expressive\Template;
+
+use League\Plates\Engine;
+
+/**
+ * Template implementation bridging league/plates
+ */
+class Plates implements TemplateInterface
+{
+    /**
+     * @var Engine
+     */
+    protected $template;
+
+    public function __construct(Engine $template = null)
+    {
+        if (null === $template) {
+            $template = $this->createTemplate();
+        }
+        $this->template = $template;
+    }
+
+    /**
+     * Create a default Plates engine
+     *
+     * @params string $path
+     * @return Engine
+     */
+    private function createTemplate()
+    {
+        return new Engine();
+    }
+
+    /**
+     * Render
+     *
+     * @param string $name
+     * @param array $params
+     * @return string
+     */
+    public function render($name, array $params)
+    {
+        return $this->template->render($name, $params);
+    }
+
+    /**
+     * Set the template directory
+     *
+     * @param string $path
+     */
+    public function setPath($path)
+    {
+        $this->template->setDirectory($path);
+    }
+
+    /**
+     * Get the template directory
+     *
+     * @return string
+     */
+    public function getPath()
+    {
+        return $this->template->getDirectory();
+    }
+}
--- a/src/Template/Plates.php
+++ b/src/Template/Plates.php
@@ -53,22 +53,34 @@ class Plates implements TemplateInterface
     }
 
     /**
-     * Set the template directory
+     * Add a path for template
      *
      * @param string $path
+     * @param string $namespace
      */
-    public function setPath($path)
+    public function addPath($path, $namespace = null)
     {
-        $this->template->setDirectory($path);
+        if (!$namespace && !$this->template->getDirectory()) {
+            $this->template->setDirectory($path);
+            return;
+        }
+        $this->template->addFolder($namespace, $path, true);
     }
 
     /**
      * Get the template directory
      *
-     * @return string
+     * @return TemplatePath[]
      */
-    public function getPath()
+    public function getPaths()
     {
-        return $this->template->getDirectory();
+        $paths = [];
+        if ($this->template->getDirectory()) {
+            $paths[] = new TemplatePath($this->template->getDirectory());
+        }
+        foreach ($this->template->getFolders() as $folder) {
+            $paths[] = new TemplatePath($folder->getPath(), $folder->getName());
+        }
+        return $paths;
     }
 }
--- a/src/Template/TemplatePath.php
+++ b/src/Template/TemplatePath.php
@@ -1,0 +1,65 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
+ * @copyright Copyright (c) 2015 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   https://github.com/zendframework/zend-expressive/blob/master/LICENSE.md New BSD License
+ */
+
+namespace Zend\Expressive\Template;
+
+class TemplatePath
+{
+    /**
+     * @var string
+     */
+    protected $path;
+
+    /**
+     * @var string
+     */
+    protected $namespace;
+
+    /**
+     * Constructor
+     *
+     * @param string $path
+     * @param string $namespace
+     */
+    public function __construct($path, $namespace = null)
+    {
+        $this->path      = $path;
+        $this->namespace = $namespace;
+    }
+
+    /**
+     * __toString
+     *
+     * @return string
+     */
+    public function __toString()
+    {
+        return $this->path;
+    }
+
+    /**
+     * Get the namespace
+     *
+     * @return string
+     */
+    public function getNamespace()
+    {
+        return $this->namespace;
+    }
+
+    /**
+     * Get the path
+     *
+     * @return string
+     */
+    public function getPath()
+    {
+        return $this->path;
+    }
+}
--- a/src/Template/Twig.php
+++ b/src/Template/Twig.php
@@ -27,31 +27,41 @@ class Twig implements TemplateInterface
      */
     protected $template;
 
+    /**
+     *  Constructor
+     *
+     * @param TwigEnvironment $template
+     */
     public function __construct(TwigEnvironment $template = null)
     {
         if (null === $template) {
-            $template = $this->createTemplate();
-        } else {
-            if (!$template->getLoader()) {
-                $this->twigLoader = new TwigFilesystem();
-                $template->setLoader($this->twigLoader);
-            } else {
-                $this->twigLoader = $template->getLoader();
-            }
+            $template = $this->createTemplate($this->getDefaultLoader());
         }
-        $this->template = $template;
+        if (! $template->getLoader()) {
+            $template->setLoader($this->getDefaultLoader());
+        }
+        $this->template   = $template;
+        $this->twigLoader = $template->getLoader();
     }
 
     /**
      * Create a default Twig environment
      *
-     * @params string $path
      * @return TwigEnvironment
      */
-    private function createTemplate()
+    private function createTemplate(TwigFilesystem $loader)
     {
-        $this->twigLoader = new TwigFilesystem();
-        return new TwigEnvironment($this->twigLoader);
+        return new TwigEnvironment($loader);
+    }
+
+    /**
+     * Get the default loader for template
+     *
+     * @return TwigFilesystem
+     */
+    private function getDefaultLoader()
+    {
+        return new TwigFilesystem();
     }
 
     /**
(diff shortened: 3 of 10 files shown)


src/Template/Twig.php
```diff
@@ -0,0 +1,92 @@
...
+        if (null === $template) {
+            $template = $this->createTemplate();
+        } else {
+            if (!$template->getLoader()) {
+                $this->twigLoader = new TwigFilesystem();
+                $template->setLoader($this->twigLoader);
+            } else {
+                $this->twigLoader = $template->getLoader();
+            }
+        }
```
weierophinney: Let's rewrite the above to the following:

```php
if (null === $template) {
    $template = $this->createTemplate();
}

if (! $template->getLoader()) {
    $this->injectDefaultLoader($template);
}

$this->twigLoader = $template->getLoader();
```

`injectDefaultLoader()` would look like this:

```php
private function injectDefaultLoader(TwigEnvironment $template)
{
    $template->setLoader(new TwigFilesystem());
}
```

See notes below for changes to `createTemplate()`.

src/Template/Twig.php
```diff
@@ -0,0 +1,92 @@
...
+    /**
+     * Create a default Twig environment
+     *
+     * @params string $path
+     * @return TwigEnvironment
+     */
+    private function createTemplate()
+    {
+        $this->twigLoader = new TwigFilesystem();
+        return new TwigEnvironment($this->twigLoader);
```
weierophinney: Once the above is in place, this becomes:

```php
$template = new TwigEnvironment();
$this->injectDefaultLoader($template);
return $template;
```

src/Template/Twig.php
```diff
@@ -0,0 +1,92 @@
...
+     *
+     * @return string
+     */
+    public function getPath()
+    {
+        $path = $this->twigLoader->getPaths();
+        if (empty($path)) {
+            return null;
+        }
+        return is_array($path) ? $path[0] : $path;
```
weierophinney: Should we raise an exception if more than one path is found? Or maybe concatenate them?

weierophinney:
Overall, looks great! I'll make the changes suggested above regarding the Twig implementation's defaults on merge; I'll think on how we might want to address the "multiple paths" question, because it's not uncommon (each of Twig, League/Plates, and zend-view allow multiple paths). We may want to use `addPath()` and `getPaths()` instead.

test/Template/PlatesTest.php
```diff
@@ -0,0 +1,56 @@
...
+{
+    public function setUp()
+    {
+        $this->platesEngine = $this->prophesize('League\Plates\Engine');
+    }
+
+    public function testConstructorWithEngine()
+    {
+        $template = new PlatesTemplate($this->platesEngine->reveal());
+        $this->assertTrue($template instanceof PlatesTemplate);
```
samsonasik: assertInstanceOf

weierophinney:
> We may want to use addPath() and getPaths() instead.

I've looked at each of our initial templating targets, and this is what I'm seeing:

## league/plates

Defines provides three ways to specify paths:

- Via a constructor argument, which is equivalent to:
- `setDirectory()`.
- Alternately, you can use `addFolder()`.

The engine loops through all three, using the directory specified at instantiation or via `setDirectory()` as a fallback if it cannot find a corresponding template in its list of folders.

`addFolder()` takes two arguments, a `$name` and the `$directory`; the `$name` is used to namespace templates in that given directory. As an example:

```php
$plates->addFolder('admin', 'path/to/templates/admin');
```

will resolve the template `admin::welcome`. A third argument to `addFolder()` is used to determine whether or not a failed lookup should look for the template in the default directory.

You can retrieve the default directory via `getDirectory()`, and the list of folders via `getFolders()`. The latter returns a collection object, from which you can call `get($name)` to retrieve the path for a given namespace.

## twig

Defines three ways to set/add paths:

- `setPaths()`
- `addPath()`
- `prependPath()`

Each takes a directory (or array of directories, in the case of `setPaths()`), and optionally a namespace. Again, like Plates, the namespace feature allows segregating templates by namespace. If no namespace is provided, a "default" is assumed; lookups will fallback to the default.

You can retrieve the list of paths via `getPaths()` (which optionally will accept a namespace, in which case it only returns paths for that namespace; by default, it will only return those on the default namespace).

## zend-view

The `TemplatePathStack` resolver defines the following for adding paths:

- `addPaths()` (adds)
- `setPaths()` (overwrites)
- ` addPath()`

These only accept paths; we do not have a concept of template namespaces. `getPaths()` will return the full list.

## Recommendation

First, it's clear that all implementations support multiple paths; this *must* be a feature, as a result.

Second, one thing I learned while doing PSR-7: ambiguity sucks. `setPaths()`: does it overwrite, or merge? `addPaths()`: does it merge, or append? This is why we typically had separate methods for retrieving one item or a list of items, and a *single* method for altering state (e.g., `withAttribute()`; note the *singular* form).

As such, I recommend:

- `addPath($path, $namespace = null)`
- `getPaths(): TemplatePath[]`

Making namespace optional allows developers to opt-in to that feature, and implementations can either raise an exception based on that detail or silently ignore it (if they don't support it). Having `getPaths()` return an array of objects allows us to define an API around namespaces:

```php
class TemplatePath
{
    public function __toString(); // the string path; for use with echo'ing
    public function getNamespace(); // return the namespace for the path
    public function getPath(); // retrieve the path itself
}
```

In terms of Plates' "default path" functionality, in that implementation, I'd argue we should:

- Always pass a boolean `true` third argument to `addFolder()`.
- Treat lack of a namespace as setting the default path (i.e., proxy to `setDirectory()`).
- When calling `getPaths()`, the returned array should be the combination of the directory and folders, with the directory having an empty namespace.

@ezimuel Does the above make sense? Do you agree with it, or have reservations?

bakura10:
wheb using api only application I actually realise it's not really needed to have a view layer as we are just serialising Json.

Can't we make this an optional dependency then?

test/Template/PlatesTest.php
ezimuel: @samsonasik it's equivalent, I don't want to use assertInstanceOf for namespace reason.

ezimuel:
@weierophinney I updated the PR with the changes proposed.

@bakura10 the idea is to include only a default template system and propose the others adapters in require-dev. I know that in API use cases this is not needed, maybe we can think to move the template support in a different repository, like `zend-expressive-template` to be able to be required in composer as needed. @weierophinney waht do you think?

bakura10:
That sounds like a good idea Enrico. My concern is that we may lead to the same problem as Zend/mvc, with a library trying to solve too much problems.

And I remember matthew's talk about ZF3 where he told we will have multiple skeleton for different purposes, and I think this is where it will shine: we could have a Zend-skeleton-whatever that will come with Zend expressive and the template layer, while still having the small core for others.

Remember that a lot of people (this is stupid but this seems to be how it works in real life) will choose a library depending on the number of dependencies. If, with Eend expressive, we can keep this "marketing advantage" of having few, that would be nice :).

Envoyé de mon iPhone

weierophinney:
I'd like to keep the template interface and implementations in this repository, but not make them required. That way:

- Users who do not need template support are not adding more dependencies.
- Users will need to select which template support they want to use explicitly.

Template libraries are less of a commodity item than containers or routers, and developers tend to hold strong opinions about which template solution they prefer. This approach will let them use exactly what they want, and only require *one* extra dependency, instead of *two*.

I'll be merging shortly.

src/Template/Plates.php
```diff
@@ -0,0 +1,86 @@
...
+     * @param string $path
+     * @param string $namespace
+     */
+    public function addPath($path, $namespace = null)
+    {
+        if (!$namespace && !$this->template->getDirectory()) {
+            $this->template->setDirectory($path);
+            return;
+        }
+        $this->template->addFolder($namespace, $path, true);
```
weierophinney: Note to self when merging: test league/plates to see what happens when you call `addFolder()` with an empty namespace.

src/Template/Plates.php
```diff
@@ -0,0 +1,86 @@
...
+    {
+        $paths = [];
+        if ($this->template->getDirectory()) {
+            $paths[] = new TemplatePath($this->template->getDirectory());
+        }
+        foreach ($this->template->getFolders() as $folder) {
+            $paths[] = new TemplatePath($folder->getPath(), $folder->getName());
+        }
+        return $paths;
+    }
```
weierophinney: Note to self when merging: see if this can be rewritten as a generator.

src/Template/TemplatePath.php
```diff
@@ -0,0 +1,65 @@
...
+     */
+    public function __toString()
+    {
+        return $this->path;
+    }
+
+    /**
+     * Get the namespace
+     *
+     * @return string
```
weierophinney: Note to self on merging: change this to `null|string`.

src/Template/Twig.php
```diff
@@ -0,0 +1,110 @@
...
+
+    /**
+     *  Constructor
+     *
+     * @param TwigEnvironment $template
+     */
+    public function __construct(TwigEnvironment $template = null)
+    {
+        if (null === $template) {
+            $template = $this->createTemplate($this->getDefaultLoader());
```
weierophinney: Nice! I didn't like putting the `getDefaultLoader()` call in `createTemplate()`; this is much better!

src/Template/Twig.php
```diff
@@ -0,0 +1,110 @@
...
+     * @param string $path
+     * @param string $namespace
+     */
+    public function addPath($path, $namespace = null)
+    {
+        if (null === $namespace) {
+            $this->twigLoader->addPath($path);
+        } else {
+            $this->twigLoader->addPath($path, $namespace);
+        }
```
weierophinney: Note to self on merging: let's do `$namespace = $namespace ?: TwigFilesystem::MAIN_NAMESPACE;`, which will allow removing the if/else block.

src/Template/Twig.php
```diff
@@ -0,0 +1,110 @@
...
+    public function getPaths()
+    {
+        $paths = [];
+        foreach ($this->twigLoader->getNamespaces() as $namespace) {
+            $name = ($namespace !== TwigFilesystem::MAIN_NAMESPACE) ? $namespace : null;
+            foreach ($this->twigLoader->getPaths($namespace) as $path) {
+                $paths[] = new TemplatePath($path, $name);
+            }
+        }
+        return $paths;
```
weierophinney: Note to self on merge: let's see if this can be done as a generator.

test/Template/PlatesTest.php
```diff
@@ -0,0 +1,62 @@
...
+{
+    public function setUp()
+    {
+        $this->platesEngine = new Engine();
+    }
+
+    public function testConstructorWithEngine()
+    {
+        $template = new PlatesTemplate($this->platesEngine);
+        $this->assertTrue($template instanceof PlatesTemplate);
```
weierophinney: @ezimuel I saw your comment to @samsonasik about this, but there's another way to do it:

```php
$this->assertInstanceOf(PlatesTemplate::class, $template);
```

The `::class` constant will resolve even an aliased class to its canonical form, making it useful for this.

test/Template/PlatesTest.php
```diff
@@ -0,0 +1,62 @@
...
+    {
+        $template = new PlatesTemplate();
+        $template->addPath(__DIR__ . '/TestAsset');
+        $paths = $template->getPaths();
+        $this->assertTrue(is_array($paths));
+        $this->assertEquals(1, count($paths));
+        $this->assertTrue($paths[0] instanceof TemplatePath);
+        $this->assertEquals($paths[0]->getPath(), __DIR__ . '/TestAsset');
+        $this->assertEquals((string) $paths[0], __DIR__ . '/TestAsset');
+        $this->assertEmpty($paths[0]->getNamespace());
```
weierophinney: Note to self: create `assertTemplatePath()` and `assertTemplateNamespace()` to reduce duplication and simplify these assertions.

weierophinney:
@ezimuel Changes look great! I'm excited to give this a whirl!

src/Template/Plates.php
weierophinney: What I discovered is that if you the template name itself is not namespaced — e.g., `render('foo')` vs `render('bar::foo')` — if `setDirectory()` has not been called, you'll get an error when you call `render()` ("default directory not defined"). As such, calling without a namespace either needs to reset the default directory, or be a no-op and raise a warning, as it's going to have unexpected consequences.

src/Template/Plates.php
weierophinney: Likely doesn't make sense to use a generator, actually; we want to return something countable, not just iterable.

src/Template/Twig.php
weierophinney: Same note as above; we need something countable as well as iterable, so not a good fit for a generator.

reinink:
Hey all, pretty cool to see you're using Plates. Let me know if you have any questions or need anything tweaked with the library to make it fit better.

weierophinney:
@reinink I noticed two things that could be problematic for us, primarily due to the fact that we're doing an abstraction layer, however:

- Differences between `setDirectory()` and `addFolder()`. From what I can tell, if you omit a namespace when calling `addFolder()`, those templates are unreachable; as such, we're having lack of a namespace proxy to `setDirectory()`. However, since that can only be a single directory, and calling it multiple times overwrites the value, we're having to raise a warning when multiple calls to `addPath()` without a namespace are called, as doing so has potentially unexpected ramifications.
- `getFolders()` returns a `Folders` object, but there's no way to query that for paths *unless you already know what namespaces are available*, and I couldn't find a way to do that. As such, we're using reflection to get at the list and return it. :smile:

Otherwise, the main idea of this part of the library is to provide something that can be dropped into middleware easily, and consumed via a common API. As such, right now, it "just works," which is exactly what I generally want from software. :+1:

reinink:
@weierophinney Hmm, okay, let me see how best to handle these issues. The second one is an easy one I think, we just have to make some data available. The first one might be a little tricker, as I don't want to break any backwards compatibility. I'll get back to you on this.

weierophinney:
@reinink To be fair, I'm not sure there needs to be anything done for the first point. It's one of those cases where I suspect *caveat emptor* comes into play; the developer is responsible for knowing what implementation is being used, and the ways in which it may be safely consumed. It was just an interesting problem from an integration/abstraction standpoint for us. :)

ezimuel:
@reinink thanks for you help. As @weierophinney said we already integrate your template system without any issue. I really love the simplicity of Plates, great work!

PR #57, opened 2015-08-12
remove ! $template->getLoader() check in Twig::__construct

samsonasik:
as `getLoader()` on `TwigEnvironment` will call `\LogicException` when loader not defined


weierophinney:
This patch is contrary to what we're trying to do; we want to inject a loader if none is present. The more appropriate solution is to wrap the `getLoader()` statement in a try/catch block:

```php
try {
    $template->getLoader();
} catch (LogicException $e) {
    $template->setLoader($this->getDefaultLoader());
}
```

I'll make that change on merge.

PR #60, opened 2015-08-12
Add error handling via Whoops

weierophinney:
This patch does the following:

- Adds `filp/whoops` as a dependency.
- Creates a new `Zend\Expressive\ErrorHandler` class, which composes a whoops runtime, `PrettyPageHandler` instance, and optionally a template renderer.
  - When an exception is encountered, it invokes whoops to handle the exception.
  - For other errors or 404 scenarios, if a template is available, it uses the template to present an error page.
- Provides container factories for whoops, the whoops pretty page handler, and the error handler.

TODO
----

- [ ] Unit tests for the error handler
- [ ] Unit tests for each factory
- [ ] Documentation

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,8 @@
         "psr/http-message": "^1.0",
         "zendframework/zend-diactoros": "^1.1",
         "zendframework/zend-servicemanager": "^2.6",
-        "zendframework/zend-stratigility": "^1.1"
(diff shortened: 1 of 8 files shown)


weierophinney:
what the... I didn't intend to merge this to master! Will open a new PR once I have more commits.

PR #62, opened 2015-08-12
Error Handling feature

weierophinney:
This builds off of #60 (which was inadvertantly merged). It adds a `TemplatedErrorHandler`, and renames `ErrorHandler` to `WhoopsErrorHandler`, having it extend the `TemplatedErrorHandler` in order to provide common functionality between the two implementations.

TODO
----

- [X] Alternate, templated-only production error handler.
- [x] Unit tests for each error handler.
- [x] Unit tests for each factory.
- [x] Documentation.

### DIFF
--- a/doc/book/error-handling.md
+++ b/doc/book/error-handling.md
@@ -1,0 +1,345 @@
+# Error Handling
+
+zend-expressive provides error handling out of the box, via zend-stratigility's [FinalHandler
+implementation](https://github.com/zendframework/zend-stratigility/blob/master/doc/book/api.md#finalhandler).
+This pseudo-middleware is executed in the following conditions:
+
+- If the middleware stack is exhausted, and no middleware has returned a response.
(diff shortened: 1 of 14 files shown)


weierophinney:
@ezimuel This is now ready for review!

samsonasik:
as Plates in composer.json is on 'suggest', I think the composer require league/plates note need to be added

weierophinney:
Adding plates will be covered in the templating documentation; this is
intended for illustrating how to configure the error handler services, no
more, no less.

weierophinney:
> Should not this be zend_expressive?

We've typically used dash-separated for top-level keys related to modules in ZF2 (e.g., look at all Apigility modules), and I followed the same convention here.

> or a more general convention: ['vendor']['component']['option']?

Potentially. I hate pushing stuff two levels deep from the outset, though, and since we don't do this anywhere in any other ZF components (it's typically component.option in ZF config), we'd be introducing a new paradigm.

PR #63, opened 2015-08-12
[WIP] ModularApplicationFactory: support ZF2 modules

mtymek:
This proof-of-concept PR creates alternative factory for `Application` class, able to load ZF2 modules. 3rd party modules can provide configuration for `Zend\ServiceManager` in order to expose services with no configuration/wiring required. More complex `Expressive` applications can be split into modules as well, to improve maintainability.

Obviously, this powerful feature comes at risk: ZF2 modules often hook to various application events that won't work under `Expressive`. This could create confusion for module devs and consumers.

I can see three options to avoid this confusion:

1. Don't merge this feature at all (sad :) )
2. Create separate module ecosystem for `Expressive`
3. Allow people to tag/indicate if module is built for ZF2, Expressive, or both.

I like option 3, as sooner or later (with arrival of ZF3), we may have to add this feature to modules.zendframework.com anyway.

Let me know if you like this, and (if yes) how would you like me to structure it.

**Usage:**

```php
// index.php

$appFactory = new ModularApplicationFactory();
$app = $appFactory->create(
     // system configuration, similar to ZF2
    [
        'modules' => [
            'Application',
            'AnotherModule'
        ],
        'module_listener_options' => [
            'config_glob_paths' => [
                'config/autoload/{{,*.}global,{,*.}local}.php',
            ],
        ]
    ]
);
$app->run();
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,6 +26,8 @@
         "zendframework/zend-view": "^2.5",
         "phpunit/phpunit": "^4.7",
         "squizlabs/php_codesniffer": "^2.3",
+        "zendframework/zend-config": "^2.5",
+        "zendframework/zend-modulemanager": "^2.5",
         "zendframework/zend-mvc": "^2.5",
         "zendframework/zend-psr7bridge": "^0.1.0"
     },
@@ -44,6 +46,8 @@
         "nikic/fast-route": "^0.6.0 to use the FastRoute routing adapter",
         "zendframework/zend-mvc": "^2.5 to use the zend-mvc TreeRouteStack routing adapter",
         "zendframework/zend-psr7bridge": "^0.1.0 to use the zend-mvc TreeRouteStack routing adapter",
+        "zendframework/zend-config": "^2.5 to enable support for ZF2 modules",
+        "zendframework/zend-modulemanager": "^2.5 to enable support for ZF2 modules",
         "twig/twig": "^1.19 to use the Twig template adapter",
         "zendframework/zend-view": "^2.5 to use the Zend View template adapter"
     }
--- a/src/Container/ModularApplicationFactory.php
+++ b/src/Container/ModularApplicationFactory.php
@@ -1,0 +1,65 @@
+<?php
+
+namespace Zend\Expressive\Container;
+
+use Zend\Config\Factory as ConfigFactory;
+use Zend\Expressive\Application;
+use Zend\Expressive\Container\ApplicationFactory as ExpressiveApplicationFactory;
+use Zend\ModuleManager\Listener\ConfigListener;
+use Zend\ModuleManager\Listener\ModuleResolverListener;
+use Zend\ModuleManager\ModuleEvent;
+use Zend\ModuleManager\ModuleManager;
+use Zend\ServiceManager\Config;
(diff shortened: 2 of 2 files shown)


bakura10:
Hi,

As you said, the problem is that this won't really work once you introduce events in the loop. A lot of modules assume that some initializers like EventMznagerAwareInterface are set up, but those initializers are set in Zend/Mvc, and introducing a dependency to Mvc is a no go.

I've realised that a lot of modules can be rewritten (and actually completely abstracted from any framework) by using PSR-7 and the middleware logic. I'd therefore be for creating a new ecosystem or migrating existing module to a middleware compliant library. That sounds like a better solution than trying to make working two things that are fundamentally different in their workflow.

codeliner:
As much as I like the idea I agree with @bakura10 that integrating the zf2 module system will only cause problems.

I currently refactor a sample application to use zend-expressive instead of zend-mvc.
One of the first things I thought of was how can I set up Doctrine because `DoctrineORMModule` will not work. Same for `AssetManager` and a lot of other great zf2 modules.

Some more problems which make module integration nearly impossible:
- you don't have a `(shared) event manager` to attach to other module logic
- module `services configuration` will not work because zend-servicemanager is only a dafault and things like ControllerManager, ViewManager and others are not available at all
- you have no `MvcEvent` to hook into the dispatch process, so auth modules, AssetManager, etc. can not attach their listeners
- `Console requests` are not handled by zend-expressive
- ...

All modules which deal with the `request/response` lifecycle can provide `Psr-7 middleware` instead of MvcEvent listeners (or provide both to be usable in both systems)

`Infrastructure modules` like DoctrineORMModule need to provide `container factories` instead of `ServiceFactory`s. A container factory can look like this:

```php
namespace Vendor\ContainerFactory;

use Vendor\Component;

class ComponentFactory
{
    {
        $appConfig = $container->get('config');

        $dependency = $container->get('third_party_dependency');

        return new Component($appConfig['vendor']['component']['options'], $dependency);
    }
}
```
Problem is: a factory like shown above can only work if the `application config` is available as a service in the `container` and if the third party dependency is also registered in the container with the service id `third_party_container`. IMO these conventions would allow us to write reusable factory logic for components and use it in any ecosystem or framwork. If we would also have a Psr for an `event dispatcher` and a Psr for `Messaging` (something like [Zend\Stdlib\Message](http://framework.zend.com/apidoc/2.0/classes/Zend.Stdlib.Message.html)) I guess we would be able to create the next generation PHP ecosystem.

bakura10:
I've thought of a differnet way to do it, please see #72

weierophinney:
I'd prefer to see this as a separate project. I think it could definitely be interesting for ZF2 users wanting to migrate to middleware, but I also think it doesn't belong directly in zend-expressive; if we add this, we'll also have to add similar factories for Symfony, Laravel, etc. Keep it in a single-purpose library, and we can potentially bring it under the zendframework org later once the approach has proven itself.

mtymek:
Done - at least initial version :)

https://github.com/mtymek/modular-expressive
https://packagist.org/packages/mtymek/modular-expressive

PR #65, opened 2015-08-14
Resolve issue with the template adapter name in exception

mwillbanks:
Pretty much, Twig is not the only adapter that is using this :)


ezimuel:
@mwillbanks Thanks!

PR #69, opened 2015-08-15
using BadMethodCallException from Exception namespace

samsonasik:


samsonasik:
rebased

src/Application.php
```diff
@@ -9,7 +9,6 @@
 
 namespace Zend\Expressive;
 
-use BadMethodCallException;
 use DomainException;
```
weierophinney: If you're going to use the local `BadMethodCallException`, we should do so for `DomainException` as well. I'll take care of that on merge.

Issue #40, opened 2015-08-07
multiple routes with same path


rodsouto:
Hi, I tried to do this

$app->get('/users', 'Users::FetchAll');
$app->post('/users', 'Users::Create');

but the last route overwrites the first one because in https://github.com/zendframework/zend-expressive/blob/master/src/Router/Aura.php#L65 it's using the path as the route name

I'm currently using Aura.Router in other project and I give each route a different name to be able to call $router->generate($routeName, $params), I don't know if you would be interested in this, but at least if we use a combination of path+method as route name it will fix the problem

weierophinney:
I'm thinking we have a couple paths forward.

First, I will create both `setName()` and `getName()` methods on the `Route` class.

The implementation of `getName()` will use the path + *imploded*, delimited methods, in a specific order. Thus, if no name is set on the route, it will have one that can be discovered, but also, more importantly have a unique name.

Obviously, if `setName()` is called, that name will be used.

```php
$app->get('/users', 'Users::FetchAll')->setName('fetch-users');
```

Alternately, the name can be an additional argument. The only problem with that approach, I feel, is that it makes the implementation of `route()` far more difficult and less intuitive; if you want to allow all methods, you have to pass a value for the third argument.

Next, in the `Container\ApplicationFactory`, we have two options:

- Using the route key as the name:

  ```php
  'routes' => [
      'fetch-users' => [
          'path' => '/users',
          'middleware' => 'Users::FetchAll',
      ]
  ]
  ```

- Or via a "name" key in the route itself:

  ```php
  'routes' => [
      [
          'name' => 'fetch-users',
          'path' => '/users',
          'middleware' => 'Users::FetchAll',
      ]
  ]
  ```

I'm leaning towards the second, as it's more explicit, and then we don't have to worry about determining if an integer index is intended as the name, or if we should auto-generate a name.

I'll start working on this shortly.

ezimuel:
@weierophinney I was testing this issue and I found that also Zf2 router adapter has the same problem. I provided this test to add in RouteMiddlewareTest: https://gist.github.com/ezimuel/56ae3d169c2d3fc2b82d.

rodsouto:
:+1: to setName() 

:+1: to "name" key in the route config

They are both the more explicit solutions :)

Thanks @weierophinney!

weierophinney:
@ezimuel I didn't get to this today; if you want to take it while I'm on vacation, go for it; otherwise, I'll address it as soon as I return.

[...]

PR #73, opened 2015-08-18
Fixed issue #40 + minor fixes

ezimuel:
**NOTE:** this PR needs the merge of https://github.com/zendframework/zend-psr7bridge/pull/2.

This PR fixes issue #40. I added the name of the route in Router/Route. You can specify the name as the last optional parameter in the constructor. If the name is not specified I create a new one based on the path and the HTTP methods:
```php
$this->name = empty($name) ? $path . '^' . implode('_', (array) $this->methods) : $name;
```
I also added the naming support in Application::route(). You can add a name as last parameter, for instance adding the `'foo'` name:
```php
$app->route('/foo', $middleware, [ 'GET' ], 'foo');
\\ also using direct http methods
$app->get('/foo', $middleware, 'foo');
```
We cannot have a syntax like that: 
```php
$app->get('/foo', $middleware)->setName('foo');
```
Because the `route()` method, called by `get()`, add the route to the router adapter.

For the ZF2 route adapter I used the HTTP method route and a special `'fail'` route injected in each route to manage the 405 error (Method not allowed).

I also added some tests to cover more route scenarios with or without route name.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -132,17 +132,23 @@ class Application extends MiddlewarePipe
         if (! in_array(strtoupper($method), $this->httpRouteMethods, true)) {
             throw new BadMethodCallException('Unsupported method');
         }
-
-        if (count($args) !== 2) {
+        $numArgs = count($args);
+        if ($numArgs < 2) {
             throw new BadMethodCallException(sprintf(
-                '%s::%s requires exactly 2 arguments; received %d',
+                '%s::%s requires at least 2 arguments; received %d',
                 __CLASS__,
                 $method,
                 count($args)
             ));
         }
 
-        $args[] = [$method];
+        if ($numArgs === 3) {
+            $name    = $args[2];
+            $args[2] = [$method];
+            $args[]  = $name;
+        } else {
+            $args[] = [$method];
+        }
 
         // @TODO: we can use variadic parameters when dependency is raised to PHP 5.6
         return call_user_func_array([$this, 'route'], $args);
@@ -261,10 +267,11 @@ class Application extends MiddlewarePipe
      * @param string|Router\Route $path
      * @param callable|string $middleware Middleware (or middleware service name) to associate with route.
      * @param null|array $methods HTTP method to accept; null indicates any.
+     * @param null|string $name the name of the route
      * @return Router\Route
      * @throws Exception\InvalidArgumentException if $path is not a Router\Route AND middleware is null.
      */
-    public function route($path, $middleware = null, array $methods = null)
+    public function route($path, $middleware = null, array $methods = null, $name = null)
     {
         if (! $path instanceof Router\Route && null === $middleware) {
             throw new Exception\InvalidArgumentException(sprintf(
@@ -277,13 +284,14 @@ class Application extends MiddlewarePipe
             $route   = $path;
             $path    = $route->getPath();
             $methods = $route->getAllowedMethods();
+            $name    = $route->getName();
         }
 
         $this->checkForDuplicateRoute($path, $methods);
 
         if (! isset($route)) {
             $methods = (null === $methods) ? Router\Route::HTTP_METHOD_ANY : $methods;
-            $route   = new Router\Route($path, $middleware, $methods);
+            $route   = new Router\Route($path, $middleware, $methods, $name);
         }
 
         $this->routes[] = $route;
@@ -381,7 +389,7 @@ class Application extends MiddlewarePipe
     /**
      * Determine if the route is duplicated in the current list.
      *
-     * Checks if a route with the same path exists already in the list;
+     * Checks if a route with the same name or path exists already in the list;
      * if so, and it responds to any of the $methods indicated, raises
      * a DomainException indicating a duplicate route.
      *
@@ -411,7 +419,7 @@ class Application extends MiddlewarePipe
 
         if (count($matches) > 0) {
             throw new DomainException(
-                'Duplicate route detected; same path, and one or more HTTP methods intersect'
+                'Duplicate route detected; same name or path, and one or more HTTP methods intersect'
             );
         }
     }
--- a/src/Router/Route.php
+++ b/src/Router/Route.php
@@ -29,6 +29,7 @@ use Zend\Expressive\Exception;
 class Route
 {
     const HTTP_METHOD_ANY = 0xff;
+    const HTTP_METHOD_SEPARATOR = ':';
 
     /**
      * @var int|string[] HTTP methods allowed with this route.
@@ -84,7 +85,9 @@ class Route
         $this->path       = $path;
         $this->middleware = $middleware;
         $this->methods    = is_array($methods) ? $this->validateHttpMethods($methods) : $methods;
-        $this->name       = empty($name) ? $path . '^' . implode('_', (array) $this->methods) : $name;
+        $this->name       = empty($name) ?
+                            $path . '^' . implode(self::HTTP_METHOD_SEPARATOR, (array) $this->methods) :
+                            $name;
     }
 
     /**
--- a/src/Router/Route.php
+++ b/src/Router/Route.php
@@ -51,14 +51,20 @@ class Route
     private $path;
 
     /**
+     * @var string
+     */
+    private $name;
+
+    /**
      * @param string $path Path to match.
      * @param string|callable $middleware Middleware to use when this route is matched.
      * @param int|array Allowed HTTP methods; defaults to HTTP_METHOD_ANY.
+     * @param string|null $name the route name
      * @throws Exception\InvalidArgumentException for invalid path type.
      * @throws Exception\InvalidArgumentException for invalid middleware type.
      * @throws Exception\InvalidArgumentException for any invalid HTTP method names.
      */
-    public function __construct($path, $middleware, $methods = self::HTTP_METHOD_ANY)
+    public function __construct($path, $middleware, $methods = self::HTTP_METHOD_ANY, $name = null)
     {
         if (! is_string($path)) {
             throw new Exception\InvalidArgumentException('Invalid path; must be a string');
@@ -78,6 +84,7 @@ class Route
         $this->path       = $path;
         $this->middleware = $middleware;
         $this->methods    = is_array($methods) ? $this->validateHttpMethods($methods) : $methods;
+        $this->name       = empty($name) ? $path . '^' . implode('_', (array) $this->methods) : $name;
     }
 
     /**
@@ -86,6 +93,14 @@ class Route
     public function getPath()
     {
         return $this->path;
+    }
+
+    /**
+     * @return string
+     */
+    public function getName()
+    {
+        return $this->name;
     }
 
     /**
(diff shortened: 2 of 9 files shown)


test/RouteMiddlewareTest.php
```diff
@@ -257,4 +258,220 @@ public function testRoutingSuccessResultingInContainerExceptionReRaisesException
...
+        $adapters = [];
+        if (class_exists('Aura\Router\Router')) {
+            $adapters[] = [ 'Zend\Expressive\Router\Aura' ];
+        }
+        if (class_exists('FastRoute\RouteCollector')) {
+            $adapters[] = [ 'Zend\Expressive\Router\FastRoute' ];
+        }
+        if (class_exists('Zend\Mvc\Router\RouteMatch')) {
+            $adapters[] = [ 'Zend\Expressive\Router\Zf2' ];
+        }
```
samsonasik: I think `class_exists` check is not needed as the lib already in `require-dev`

ezimuel:
@weierophinney this PR fails on Travis because we need to merge this https://github.com/zendframework/zend-psr7bridge/pull/2 before.

test/RouteMiddlewareTest.php
ezimuel: You right, I fixed it. Thanks!

src/Application.php
```diff
@@ -132,17 +132,23 @@ public function __call($method, $args)
...
                 $method,
                 count($args)
             ));
         }
 
-        $args[] = [$method];
+        if ($numArgs === 3) {
+            $name    = $args[2];
+            $args[2] = [$method];
+            $args[]  = $name;
```
weierophinney: I don't think we should support passing the name as part of the argument list:

- We already default to using the route match string if no name is provided.
- Because `setName()` is available on the returned `Route` object, it can be set after-the-fact.
- If we support it, next, developers will want `$options` as a potential argument, and then we have issues with argument order to contend with.

Since name and options are both optional and can be set on the `Route` instance, let's keep them out of the signature.

src/Application.php
```diff
@@ -261,10 +267,11 @@ public function routeMiddleware(ServerRequestInterface $request, ResponseInterfa
      * @param string|Router\Route $path
      * @param callable|string $middleware Middleware (or middleware service name) to associate with route.
      * @param null|array $methods HTTP method to accept; null indicates any.
+     * @param null|string $name the name of the route
      * @return Router\Route
      * @throws Exception\InvalidArgumentException if $path is not a Router\Route AND middleware is null.
      */
-    public function route($path, $middleware = null, array $methods = null)
+    public function route($path, $middleware = null, array $methods = null, $name = null)
```
weierophinney: See my notes above.

src/Router/Route.php
```diff
@@ -78,6 +84,7 @@ public function __construct($path, $middleware, $methods = self::HTTP_METHOD_ANY
         $this->path       = $path;
         $this->middleware = $middleware;
         $this->methods    = is_array($methods) ? $this->validateHttpMethods($methods) : $methods;
+        $this->name       = empty($name) ? $path . '^' . implode('_', (array) $this->methods) : $name;
```
weierophinney: Underscore is an allowed character in an HTTP method, which could make this problematic. Let's go with `:` or `;` (as neither is allowed in HTTP method names).

src/Application.php
ezimuel: @weierophinney if you use setName() on the returned Route object, you are not changing the name of the route for the Application, because of [this](https://github.com/ezimuel/zend-expressive/blob/fix/40/src/Application.php#L297-L298).

src/Router/Zf2.php
```diff
@@ -19,17 +19,14 @@
  *
  * This router implementation consumes zend-mvc's TreeRouteStack, (the default
  * router implementation in a ZF2 application). The addRoute() method injects
- * segment routes into the TreeRouteStack, and manages an internal route stack
- * in order to do HTTP method negotiation after a successful match (as the ZF2
- * "Method" router implementation will return a result indistinguishable from a
- * 404 otherwise).
+ * segment routes into the TreeRouteStack, and create a fail route for HTTP
```
weierophinney: `s/create a fail route/creates a route failure/`

src/Router/Zf2.php
```diff
@@ -54,21 +57,69 @@ public function __construct(TreeRouteStack $router = null)
...
+        $options = $route->getOptions();
         $options = array_replace_recursive($options, [
-            'route'   => $route->getPath(),
-            'defaults' => [
-                'middleware' => $route->getMiddleware(),
-            ],
+            'route' => $path
         ]);
-
+        $childRouteName = implode('-', $route->getAllowedMethods());
```
weierophinney: This can be problematic if `getAllowedMethods()` returns `Route::HTTP_METHOD_ANY`, as that's not an array.

src/Router/Zf2.php
```diff
@@ -54,21 +57,69 @@ public function __construct(TreeRouteStack $router = null)
...
+     */
+    private function getFailRouteConfig()
+    {
+        return [
+            'type'     => 'segment',
+            'priority' => -1,
+            'options'  => [
+                'route'    => '[/]',
+                'defaults' => [
+                    'middleware' => null
```
weierophinney: Maybe instead of null, we should use a constant value?

src/Router/Zf2.php
```diff
@@ -109,63 +155,17 @@ private function createRouter()
...
-        $params = $match->getParams();
-        $middleware = isset($params['middleware'])
-            ? $params['middleware']
-            : $this->getMiddlewareFromRoute($match->getMatchedRouteName());
+        $params    = $match->getParams();
+        $routeName = $match->getMatchedRouteName();
+
+        if (null === $params['middleware']) {
+            return RouteResult::fromRouteFailure($this->routes[$routeName]);
+        }
```
weierophinney: Wouldn't this indicate a 405 situation? In which case, we should be passing the allowed methods…

test/RouteMiddlewareTest.php
```diff
@@ -257,4 +258,214 @@ public function testRoutingSuccessResultingInContainerExceptionReRaisesException
...
+     * Get the router adapters to test
+     */
+    public function getRouterAdapters()
+    {
+        return [
+          [ 'Zend\Expressive\Router\Aura' ],
+          [ 'Zend\Expressive\Router\FastRoute' ],
+          [ 'Zend\Expressive\Router\Zf2' ]
+        ];
+    }
```
weierophinney: Kudos on combining the tests for the three supported routers!!!!

src/Router/Zf2.php
weierophinney: We could even put the allowed methods in the defaults for the fail route, to simplify that situation.

src/Application.php
weierophinney: Yep, just noticed that. Still don't like it, but we'll have to live with it.

weierophinney:
> this PR fails on Travis because we need to merge this zendframework/zend-psr7bridge#2 before.

Merged; get the other changes in, and let's see if the status goes green. :)

samsonasik:
can travis built be restarted?

weierophinney:
> can travis built be restarted?

It can, but there are changes to make regardless; it can trigger with the next set of changes pushed.

src/Router/Zf2.php
weierophinney: Never mind; you *are* passing the HTTP methods here.

src/Router/Zf2.php
weierophinney: In fact, we don't even need the child routes if we can accept any methods.

src/Router/Zf2.php
```diff
@@ -19,17 +19,14 @@
...
+ * a result indistinguishable from a 404 otherwise).
  */
 class Zf2 implements RouterInterface
 {
-    /**
-     * @var Route[] Registered routes
-     */
-    private $routes = [];
+    // Name of the route fail
+    const ROUTE_FAIL = 'fail';
```
weierophinney: Let's use the name and value "HTTP_METHOD_NOT_ALLOWED" here, to make it explicit what it's for.

src/Router/Zf2.php
weierophinney: This is likely fine, actually.

src/Router/Route.php
ezimuel: Ok, I'll use a class constant named HTTP_METHOD_SEPARATOR = ':'

src/Router/Zf2.php
ezimuel: Changed.

src/Router/Zf2.php
ezimuel: If we have a route with `HTTP_METHOD_ANY` I just add to ZF2 route without child. I'm adding a unit test for that case.

src/Router/Zf2.php
ezimuel: I used "METHOD_NOT_ALLOWED_ROUTE" to be more explicit to the route name.

ezimuel:
@weierophinney I updated the PR with your feedbacks. Now, with the merge of https://github.com/zendframework/zend-psr7bridge/pull/2 everything is green!

samsonasik:
@ezimuel how about `404` ? I try non-registered path and still got `405`

weierophinney:
@samsonasik Can you provide an example, please? I can then use that to create a unit test.

src/Router/Route.php
```diff
@@ -133,6 +151,14 @@ public function setOptions(array $options)
     }
 
     /**
+     * @param string $name
+     */
+    public function setName($name)
```
weierophinney: I'm going to remove the setter on merge. Essentially, if we set it after-the-fact, the name is irrelevant for the primary use case of consuming it with a router for URI assembly. As such, making it mutable actually is counter-productive.

samsonasik:
@weierophinney for example, I have routes: 
```
    'routes' => [
        [
            'path' => '/',
        ],
   ],
```
and it means, that `/` is accessible. but when I access `/foo`, I got `405` method not allowed.

src/Router/Route.php
ezimuel: @weierophinney Are you sure about that? I can create a Route without a name in constructor, setName() after that and use it to add a route to a Route Adapter. This is perfectly fine. The name is auto-generated if you don't provide it in the constructor but you can change it with setName() if you want.

weierophinney: @ezimuel The point is that if the name is changed *after* the route has been added to the underlying router, the change will not have the expected outcome. As an example:

```php
$app->post('/foo', 'Foo::bar')->setName('foo-bar');
```

implies that I can reference the route by the name `foo-bar` when working with the underlying router implementation, but, in fact, I cannot.

Technically, this is true for every other aspect encapsulated by `Route` as well, including `options`, as those are set as soon as `addRoute()` is called on the routing adapter. The only way having mutable properties makes sense is if we delay passing the routes to the underlying router until we actually are ready to route.

weierophinney:
@samsonasik I'm assuming this is with the ZF2 router specifically? If so, please file a different issue for that, as it's unrelated to this one.

samsonasik:
@weierophinney I'm using default Application implementation, so should be Aura.

weierophinney:
@samsonasik So, create a new issue. The Aura router changes in here are completely unrelated, and are only for mapping names to routes.

samsonasik:
ok, created #74 for that

PR #53, opened 2015-08-18
Support route assembly

weierophinney:
Allow generating URIs based on a given route specification.

- [x] Add `generateUri($name, array $substitutes = [])` to `RouterInterface`
- [x] Implement `generateUri()` for Aura router
- [x] Implement `generateUri()` for ZF2 router
- [x] Implement `generateUri()` for FastRoute router. Since this routing implementation does not include URI generation by default, this will require writing such functionality.

### DIFF
--- a/src/Router/Aura.php
+++ b/src/Router/Aura.php
@@ -134,6 +134,14 @@ class Aura implements RouterInterface
     }
 
     /**
+     * {@inheritDoc}
+     */
+    public function generateUri($name, array $substitutions = [])
+    {
+        return $this->router->generate($name, $substitutions);
+    }
+
+    /**
      * Marshal a RouteResult representing a route failure.
      *
      * If the route failure is due to the HTTP method, passes the allowed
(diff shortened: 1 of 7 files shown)


weierophinney:
The ZF2 support is problematic. It required changes to how ZF2 registers the METHOD_NOT_ALLOWED route, as well as how it registers the route name (it now uses the provided name, not the path). This will likely require a little re-thinking, as technically it means we can treat additional methods as child routes of the original if desired (and likely this will be required, as the "failure" route will match early at this point). I plan to test this before continuing to the FastRoute implementation.

weierophinney:
And yes, tests are known to be failing; I have not added the `generateUri()` method to the FastRoute implementation yet.

ezimuel:
@weierophinney I just sent this PR https://github.com/zendframework/zend-expressive/pull/75 to use the name instead of the path for ZF2. Basically, I removed the METHOD_NOT_ALLOWED route and I manage the 405 error using an internal array $allowedMethods (ex $routes).

PR #75, opened 2015-08-19
Changed the route name instead of path for ZF2 routes

ezimuel:
This PR changes the ZF2 routes definition using the route name instead of path. This can simplify the implementation of https://github.com/zendframework/zend-expressive/pull/53.

### DIFF
--- a/test/Router/Zf2Test.php
+++ b/test/Router/Zf2Test.php
@@ -70,6 +70,16 @@ class Zf2Test extends TestCase
                             'middleware' => 'foo',
                         ]
                     ]
+                ],
+                Zf2Router::METHOD_NOT_ALLOWED_ROUTE => [
+                    'type'     => 'segment',
+                    'priority' => -1,
+                    'options'  => [
+                        'route'    => '[/]',
+                        'defaults' => [
+                            Zf2Router::METHOD_NOT_ALLOWED_ROUTE => '/foo',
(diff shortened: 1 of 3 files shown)


ezimuel:
@weierophinney I just realized that using the name of the route for ZF2 will work only for literal. If we define a route like `/foo[/:id]` and we have  `/foo/4`  as url request the match using  `$this->alloweMethods['/foo[/:id]']` will never work. I'm investigating in alternatives.

ezimuel:
@weierophinney I fixed the issue with params in the route. I re-insert the METHOD_NOT_ALLOWED_ROUTE route only in the first added route with the same path. I also changed the defaults value to return the path instead of a null middleware. This simplify the management of 405 Method not allowed case.

test/Router/Zf2Test.php
```diff
@@ -164,7 +168,7 @@ public function testMatch()
 
         $result = $zf2Router->match($request);
         $this->assertInstanceOf('Zend\Expressive\Router\RouteResult', $result);
-        $this->assertEquals('/foo/GET', $result->getMatchedRouteName());
+        $this->assertEquals('/foo^GET/GET', $result->getMatchedRouteName());
```
weierophinney: I'm thinking we may want to strip the child route from the returned matched route, as it's superfluous.

PR #74, opened 2015-08-19
404 doesn't work

weierophinney:
For example, I have routes:
```
    'routes' => [
        [
            'path' => '/',
        ],
   ],

```
and it means, that / is accessible. but when I access /foo, I got 405 method not allowed. I am using default `Application` implementation, that should be Aura route.

### DIFF
--- a/src/Router/Aura.php
+++ b/src/Router/Aura.php
@@ -127,7 +127,7 @@ class Aura implements RouterInterface
         $route  = $this->router->match($path, $params);
 
         if (false === $route) {
-            return $this->marshalFailedRoute();
+            return $this->marshalFailedRoute($request);
         }
 
         return $this->marshalMatchedRoute($route);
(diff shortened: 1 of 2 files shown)


weierophinney:
@samsonasik This appears to be working fine with current master; it's possible either #73 or #75 fixed the behavior you were seeing. Let me know if the test in this patch doesn't accurately convey what you're attempting.

samsonasik:
I updated latest master, still got 405.

weierophinney:
@samsonasik As I am unable to reproduce the issue, I'll need you to either provide updates to the test I've written or write your own unit test to provide me with a reproduce case.

weierophinney:
@samsonasik I've updated the assertions, and now see the failure on Aura (but not the other routers). This may be an issue upstream; I'll check.

weierophinney:
I've isolated the issue: it's due to the updates in #40, which were added to ensure 405s are correctly returned for routes that duplicate existing paths. The problem is that it uses the failed route's name to detect if the failure was due to method, but not the current request method.

samsonasik:
i updated, and now got 200 status instead of 404:
404_200_aura

samsonasik:
@weierophinney I've created PR #76 for failing tests for aura

PR #64, opened 2015-08-19
App does not route by default

weierophinney:
Hi,

I had trouble with figuring out how to set up a separate `/api` middleware pipe.
My system set up:

1. A `root` middleware pipe (Zend\Expressive\Application) with a `home` route `/` serving the layout + assets for a SPA.
2. A `API` middleware pipe (also Zend\Expressive\Application) with a programmatic created router following the [docs](https://github.com/zendframework/zend-expressive/blob/master/doc/book/router.aura.md#programmatic-creation)

This is how my `index.php` initializes the pipes:

```php
$services = new \Zend\ServiceManager\ServiceManager(new \Zend\ServiceManager\Config(require 'config/services.php'));

$app = $services->get('Zend\Expressive\Application');

$cargoBackend = $services->get('cargo.backend');

$app->pipe('/api', $cargoBackend);

$app->run();
```

And this is the factory responsible for the `cargo.backend` middleware pipe:

```php
    {
        $router = $container->get('cargo.backend.router');

        $app = AppFactory::create($container, $router);

        //This is the problem a want to address with the issue!!!
        $app->pipe("/", [$app, 'routeMiddleware']);

        return $app;
    }
```

I was not aware that I have to pipe the router manually when using the `AppFactory`. When passing a router to it I expect that the router gets activated automatically by the factory.
I would like to submit a PR that fixes the problem, but want to ask first if this would have side effects 
and if such a change is wanted/needed? Otherwise we should document it. The cookbook would be a good place.

I have a second problem but currently don't know what causes it:
Before I added the line `$app->pipe("/", [$app, 'routeMiddleware'])` to the factory my route was not invoked (`/api/locations`) but I got an empty response with a `200 ok status code` instead of a `404 not found`. Maybe this happens because the `api app` is set up with its own `emitter` and `final handler` by default, so the `final handler` of the `root app` is not invoked?

### DIFF
--- a/doc/book/router.aura.md
+++ b/doc/book/router.aura.md
@@ -61,6 +61,30 @@ $router = new AuraBridge($auraRouter);
 $app = AppFactory::create(null, $router);
 ```
 
+> ### Piping the route middleware
+>
+> If you programmatically configure the router and add routes without using
+> `Application::route()`, you may run into issues with the order in which piped
+> middleware (middleware added to the application via the `pipe()` method) is
+> executed.
+>
+> To ensure that everything executes in the correct order, you can call
+> `Application::pipeRouteMiddleware()` at any time to pipe it to the
+> application. As an example, after you have created your application
+> instance:
+>
+> ```php
+> $app->pipe($middlewareToExecuteFirst);
+> $app->pipeRouteMiddleware();
+> $app->pipe($errorMiddleware);
+> ```
+>
+> If you fail to add any routes via `Application::route()` or to call
+> `Application::pipeRouteMiddleware()`, the routing middleware will be called
+> when executing the application. **This means that it will be last in the
+> middleware pipeline,** which means that if you registered any error
+> middleware, it can never be invoked.
+
 ## Factory-Driven Creation
 
 We recommend using an Inversion of Control container for your applications;
--- a/doc/book/router.fast-route.md
+++ b/doc/book/router.fast-route.md
@@ -60,6 +60,30 @@ $router = new FastRouteBridge($fastRoute, $getDispatcher);
 $app = AppFactory::create(null, $router);
 ```
 
+> ### Piping the route middleware
+>
+> If you programmatically configure the router and add routes without using
+> `Application::route()`, you may run into issues with the order in which piped
+> middleware (middleware added to the application via the `pipe()` method) is
+> executed.
+>
+> To ensure that everything executes in the correct order, you can call
+> `Application::pipeRouteMiddleware()` at any time to pipe it to the
+> application. As an example, after you have created your application
+> instance:
+>
+> ```php
+> $app->pipe($middlewareToExecuteFirst);
+> $app->pipeRouteMiddleware();
+> $app->pipe($errorMiddleware);
+> ```
+>
+> If you fail to add any routes via `Application::route()` or to call
+> `Application::pipeRouteMiddleware()`, the routing middleware will be called
+> when executing the application. **This means that it will be last in the
+> middleware pipeline,** which means that if you registered any error
+> middleware, it can never be invoked.
+
 ## Factory-Driven Creation
 
 We recommend using an Inversion of Control container for your applications;
--- a/doc/book/router.zf2.md
+++ b/doc/book/router.zf2.md
@@ -73,6 +73,30 @@ $router = new Zf2Bridge($zf2Router);
 $app = AppFactory::create(null, $router);
 ```
 
+> ### Piping the route middleware
+>
+> If you programmatically configure the router and add routes without using
+> `Application::route()`, you may run into issues with the order in which piped
+> middleware (middleware added to the application via the `pipe()` method) is
+> executed.
+>
+> To ensure that everything executes in the correct order, you can call
+> `Application::pipeRouteMiddleware()` at any time to pipe it to the
+> application. As an example, after you have created your application
+> instance:
+>
+> ```php
+> $app->pipe($middlewareToExecuteFirst);
+> $app->pipeRouteMiddleware();
+> $app->pipe($errorMiddleware);
+> ```
+>
+> If you fail to add any routes via `Application::route()` or to call
+> `Application::pipeRouteMiddleware()`, the routing middleware will be called
+> when executing the application. **This means that it will be last in the
+> middleware pipeline,** which means that if you registered any error
+> middleware, it can never be invoked.
+
 ## Factory-Driven Creation
 
 We recommend using an Inversion of Control container for your applications;
--- a/src/Application.php
+++ b/src/Application.php
@@ -114,6 +114,7 @@ class Application extends MiddlewarePipe
      */
     public function __invoke(ServerRequestInterface $request, ResponseInterface $response, callable $out = null)
     {
+        $this->pipeRoutingMiddleware();
         $out = $out ?: $this->getFinalHandler($response);
         return parent::__invoke($request, $response, $out);
     }
@@ -182,6 +183,17 @@ class Application extends MiddlewarePipe
         }
 
         return $this;
+    }
+
+    /**
+     * Register the routing middleware in the middleware pipeline.
+     */
+    public function pipeRoutingMiddleware()
+    {
+        if ($this->routeMiddlewareIsRegistered) {
+            return;
+        }
+        $this->pipe([$this, 'routeMiddleware']);
     }
 
     /**
@@ -297,10 +309,7 @@ class Application extends MiddlewarePipe
 
         $this->routes[] = $route;
         $this->router->addRoute($route);
-
-        if (! $this->routeMiddlewareIsRegistered) {
-            $this->pipe([$this, 'routeMiddleware']);
-        }
+        $this->pipeRoutingMiddleware();
 
         return $route;
     }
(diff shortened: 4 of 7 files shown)


codeliner:
> Except that routing is the default and expected path; piping other middleware, while a feature, is not the primary feature of zend-expressive; routing is its value-add over zend-stratigility. If you don't need routing, use zend-stratigility by itself; if you do, zend-expressive is what you want.

This is a very important information and should be added to the README.md. However, in this case I would say the route middleware is ALWAYS piped by the factory and if you want to pipe another middleware before routing you have to call something like `$app->pipePreRoutingMiddleware()`.

codeliner:
The problem I have with `$app->route()` is that you have to know that the router is piped automatically when calling `$app-route()` the first time. So it is very important that you **first** pipe pre routing middleware **then** add routes with `$app->route()` and as a **last** step pipe post routing middleware. So two times you have to pipe middleware but in between you should add your routes.
Or do I have a wrong understanding of set up process?

weierophinney:
Addressing:

> However, in this case I would say the route middleware is ALWAYS piped by the factory and if you want to pipe another middleware before routing you have to call something like $app->pipePreRoutingMiddleware().

and:

> The problem I have with $app->route() is that you have to know that the router is piped automatically when calling $app-route() the first time.

You definitely understand the workflow, and it's deliberately done that way.

To re-iterate:

- If creating your Application instance programmatically, pipe any middleware you want to execute before routing before you `route()` anything and/or before calling `pipeRoutingMiddleware()`
- If using the `Container\ApplicationFactory`, pre/post middleware are added in order (see https://github.com/zendframework/zend-expressive/blob/master/src/Container/ApplicationFactory.php#L173-L175 — the changes in this PR are actually in the `injectRoutes()` method, so it doesn't affect that workflow).

I do not plan on adding pre/post pipe registration hooks, for one very big reason: Stratigility uses `SplQueue` internally, and its `enqueue`/`dequeue` methods. Since this is an internal detail used by both `MiddlewarePipe` and `Next`, it's not something we can easily modify — and, to be honest, knowing the difficulties developers have had understanding priority in ZF2's EventManager, using a priority queue is not something I want to introduce here. Since the same effect can be achieved programmatically in a trivial fashion, I'm comfortable leaving the workflow as-is.

The main point is to consider where the two different mechanisms originate:

- `pipe()` belongs to zend-stratigility, which is a middleware foundation.
- `route()` belongs to zend-expressive, which builds on zend-stratigility to specifically address:
  - using a container for lazy-loading middleware
  - dynamic routing of middleware (vs the static, literal-only routing in zend-stratigility)
  - error handling
  - templating

The fact that `pipe()` exists in zend-expressive is because we build on zend-stratigility. `pipe()`'s use-cases within zend-expressive should then typically be limited to:

- registering error handlers
- registering middleware to execute on each request
- segregating middleware by sub-path

For many, these may or may not be of interest; the basic features of routing and templating may be enough (and even templating may not be of much interest in some cases; consider APIs, for example). For those where they *are* of interest: we support it, but in a very specific workflow. Considering routing is the central value-add of zend-expressive, then making the default use case of routing an out-of-the-box experience is the primary goal. If we force users to register the route pipeline manually when they've already added routes, then we've missed that goal.

I think the proposed solution is a reasonable trade-off; you can do advanced routing configuration if desired, and still benefit from the integrated routing in zend-expressive.

weierophinney:
> To facilitate your use case, I'm thinking that we can pipe the routeMiddleare by default if we pass a valid $router object in the AppFactory

By the way, the above is not a viable option; it means there is zero way to have middleware that will execute prior to routing.

weierophinney:
@codeliner I think I found the source of your second report, and it's in your own code:

- https://gist.github.com/codeliner/ced591bedcca422c7e42#file-index-php-L26
- https://gist.github.com/codeliner/ced591bedcca422c7e42#file-index-php-L44

You're not returning!

In the first instance, you're writing to the response, but not returning it. This particular instance works because the stream in the body is mutable, and the FinalHandler detects that the body size has changed since execution began; as such, it will correctly return a 200 with the content.

However, in the second, because you do not call `return $next($request, $response);`, but instead `$next($request, $response);`, the original response passed to the middleware is what is passed to the final handler on completion of the pipeline, meaning you get a 200 with no content. Add the return statement, and it will work. (This is documented in Stratigility: https://github.com/zendframework/zend-stratigility/blob/master/doc/book/api.md#next)

codeliner:
@weierophinney Thank you for the detailed explanation. I still would prefer pre and post pipe hooks because IMO these are more explicit, but I can live with the trade-off ;)

> You're not returning!
Oh no, shame on me :tired_face: Sorry for that.

At the moment I have the feeling that understanding EventManager priorities was easier than all this middleware piping :smile: 
Maybe this will change when playing a bit more with the concept but currently I can not imagine how all this will work when adding more middleware to the pipe: oauth2, content-negotiation, handling CORS requests, serving assets, ... Sure, the easy answer is: Just pipe the middlewares in the correct order but I can imagine that this can become a complex task when I have to pipe > 10 middleware which all do something with the request and/or response. but we'll see.

weierophinney:
@codeliner We did a prototype of Apigility in node.js last year, using middleware. Interestingly, it was far easier to:

- understand the order in which operations occurred
- modify that order
- slip in alternate implementations

Keep playing! :smile:

codeliner:
@weierophinney :+1: I'll do that. Refactoring of my cargo sample from zend-mvc to zend-expressive is nearly finished. I guess people will ask me about middleware in the future so I want to be prepared :smile: 

One more propblem and one more question:
**The problem**: 
With Relay a second PSR-7 middleware dispatcher was added to the game by pmjones. So fare no problem. But now we have zend-stratigility which requires  an arity of 4 for error middlewares like sencha connect does and we have Relay which does not work with a special arity for error middlewares. If I now want to provide an error middleware (maybe something like AssetManager, listening on 404 responses and serving assets instead) I have a problem because I need to provide to types of middleware one for stratigiligty and one for relay. I think that is bad and misses the goal of interoperable middlewares. Would be nice if all PSR-7 middleware dispatcher use the same best practices.

**The question**
Do you think middleware can replace event-driven approaches in other areas (not request/response related areas) too? For example Tactician is a command bus which also uses middleware but of course with a totally different middleware signature. prooph/service-bus on the other hand is inspired by the event-driven zend-mvc dispatch. What I really like with the event-driven approach is that plugins can be `ListenerAggregate`s which know the correct position in the dispatch process for themselves. So a user don't need to worry about it. I'm a bit lost with this question because on the one hand middleware seems to be easier to understand but on the other hand if every plugin system is based on middleware then users need to pipe everything and have to deal with a lot of different middleware.

weierophinney:
@codeliner —

re:

> now we have zend-stratigility which requires an arity of 4 for error middleware like sencha connect does and we have Relay which does not work with a special arity for error middleware. If I now want to provide an error middleware I have a problem because I need to provide to types of middleware one for stratigiligty and one for relay.

This is one place where I wish people would do some homework before inventing something! The reason Stratigility, and by extension Expressive, uses the "arity of 4 === error middleware" paradigm is because it's already been invented and battle-tested elsewhere (i.e., in Connect and ExpressJS).

Interestingly, Relay *does not offer a concept of error middleware*. If you wanted to invoke a later middleware that can handle errors, you'd have to do something like adding the error to the request attributes, which becomes a *convention-based approach*, vs a *contract-based approach*.

Slim3 offers error handling, but it's separate from normal middleware; essentially, if an exception is raised within the application, Slim passes the request, response, and exception to the registered "error_handler" service; there is no stack.

As such, likely the safest, cross-project way to handle errors is to simply... not, and let the application handle them. In the case of Stratigility/Expressive, if an exception is raised, we pass it on to error middleware, if any, and then the final handler will process it. Thus, if you don't register any error middleware, but use our final handler capabilities as detailed [in the manual](http://zend-expressive.readthedocs.org/en/latest/error-handling/), you should be fine.

A group of us have been focused on identifying and categorizing the various middleware patterns, but, interestingly, error handling and error middleware has not been part of that discussion.

re:

> Do you think middleware can replace event-driven approaches in other areas (not request/response related areas) too?

Yes and no. I think events still have a place — for instance, anywhere you might use a command bus, or anywhere you might be tempted to use the observer pattern. Most of the time, these will be limited to your domain layer, however — not the request/response lifecycle. (That said, Guzzle uses middleware, too, for the client-lifecycle. This allows for things like response filtering and transformation.)

Essentially I've answered your question with, "it depends." I think middleware makes a ton of sense in creating a workflow engine, which is what you need for the request/response lifecycle. That's what we're addressing with this library.

codeliner:
@weierophinney 

> This is one place where I wish people would do some homework before inventing something!

:+1: 

> As such, likely the safest, cross-project way to handle errors is to simply... not, and let the application handle them.

yeah, seems to be the best idea.

> interestingly, error handling and error middleware has not been part of that discussion.

That is really interesting! Because good error handling should be part of every system.

> Essentially I've answered your question with, "it depends." I think middleware makes a ton of sense in creating a workflow engine, which is what you need for the request/response lifecycle.

Hhm, dispatching a message with a message bus has some similarity with dispatching a request and it is definitely some kind of workflow engine. However, the event-driven approach we use currently works great. But I will keep an eye on the middleware trend. You are absolutely right. "It depends".

Thank you for answering my off topic question.

PR #76, opened 2015-08-19
failing tests for 404

samsonasik:


samsonasik:
The failing tests seems not only for Aura: 
```
1) ZendTest\Expressive\RouteMiddlewareTest::testWithOnlyRootPathRouteDefinedRoutingToSubPathsShouldReturn404 with data set "aura" ('Zend\Expressive\Router\Aura')
Failed asserting that 200 matches expected 404.

/Users/samsonasik/www/zf-components-contribute/zend-expressive/test/RouteMiddlewareTest.php:531

2) ZendTest\Expressive\RouteMiddlewareTest::testWithOnlyRootPathRouteDefinedRoutingToSubPathsShouldReturn404 with data set "fast-route" ('Zend\Expressive\Router\FastRoute')
Failed asserting that 200 matches expected 404.

/Users/samsonasik/www/zf-components-contribute/zend-expressive/test/RouteMiddlewareTest.php:531

3) ZendTest\Expressive\RouteMiddlewareTest::testWithOnlyRootPathRouteDefinedRoutingToSubPathsShouldReturn404 with data set "zf2" ('Zend\Expressive\Router\Zf2')
Failed asserting that 200 matches expected 404.

/Users/samsonasik/www/zf-components-contribute/zend-expressive/test/RouteMiddlewareTest.php:531
```

weierophinney:
The test is flawed, because the route middleware never returns a 404. This is intentional; we want middleware piped later in the pipeline to be able to handle the request. The final handler returns a 404 if it detects that no middleware has returned a response and/or written to the response.

As such, if no middleware matches during routing, the route middleware returns the response untouched.

The code as it currently stands is correct.

PR #79, opened 2015-08-20
Inconsistent service name for config

weierophinney:
Today we tried to find out what is the most common name to register the application configuration in the container. In ZF2 you find three versions: `Config`, `config` and `configuration`.
The first two equal to the same service name because zend-servicemanager normalizes it. The latter is an alias for `Config`.
Looking at zend-expressive we recognized that it again uses the normalization "feature": 
https://github.com/zendframework/zend-expressive/blob/master/src/Container/ApplicationFactory.php#L188

Note that `Config` is used but the docs give an example with `config` as service name:
https://github.com/zendframework/zend-expressive/blob/master/doc/book/usage-examples.md#hello-world-using-a-configuration-driven-container

So docs + ApplicationFactory will only work as long as you use zend-servicemanager as container.
Otherwise, it will fail silently.
We should change that. But which version do you prefer?

### DIFF
--- a/doc/book/error-handling.md
+++ b/doc/book/error-handling.md
@@ -113,10 +113,10 @@ container implementations to simplify setup.
 ### WhoopsPageHandlerFactory
 
 > - Register this factory as `Zend\Expressive\WhoopsPageHandler`.
-> - This service optionally consumes the service `Config`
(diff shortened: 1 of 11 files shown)


codeliner:
:boom: :smile:

PR #81, opened 2015-08-21
added cookbook for set up custom 404 page

samsonasik:
@weierophinney this is what I can get from your site: https://github.com/weierophinney/mwop.net/blob/master/config/autoload/global.php . Still not sure about using 404 page with `WhoopsErrorHandler` or `TemplatedErrorHandler`. 

Note:
defining `'error' => true` in `middleware_pipeline` config make it not working. Please let me know if I missed something. Thanks.

### DIFF
--- a/doc/book/cookbook.md
+++ b/doc/book/cookbook.md
@@ -253,16 +253,8 @@ class NotFound
 }
 ```
 
-After it, we need to register the middleware into `invokables` type in our service config:
+> We can register the `Application\NotFound` instance as a service in [service container](https://github.com/zendframework/zend-expressive/blob/master/doc/book/container/intro.md).
 
-```php
-//... config/services.php
-    'invokables' => [
-        // ...
-        'Application\NotFound' => 'Application\NotFound',
-    ],
-//...
-```
 Now, we can configure the `middleware_pipeline` under `post_routing`:
 
 ```php
--- a/doc/book/cookbook.md
+++ b/doc/book/cookbook.md
@@ -48,6 +48,7 @@ The above works, because every `Application` instance is itself middleware, and,
 an instance of [Stratigility's `MiddlewarePipe`](https://github.com/zendframework/zend-stratigility/blob/master/doc/book/middleware.md),
 which provides the ability to compose middleware.
 
+<<<<<<< HEAD
 ## How can I specify a route-specific middleware pipeline?
 
 Sometimes you may want to use a middleware pipeline only if a particular route
@@ -234,8 +235,11 @@ return [
 ```
 
 ## How can I set 404 page?
+=======
+## How can I set custom 404 page handling?
+>>>>>>> wording current 404 for custom
 
-We can set 404 page with create new Middleware that set 404 status from response object.
+We can set custom 404 page handling with create new Middleware that set 404 status from response object, or do logging.
 
 Let's create a `NotFound` middleware:
 
@@ -248,6 +252,7 @@ class NotFound
 {
     public function __invoke($req, $res, $next)
     {
+        // other things can be done here, for eg: logging
         return $next($req, $res->withStatus(404), 'Page Not Found');
     }
 }
--- a/doc/book/cookbook.md
+++ b/doc/book/cookbook.md
@@ -48,7 +48,6 @@ The above works, because every `Application` instance is itself middleware, and,
 an instance of [Stratigility's `MiddlewarePipe`](https://github.com/zendframework/zend-stratigility/blob/master/doc/book/middleware.md),
 which provides the ability to compose middleware.
 
-<<<<<<< HEAD
 ## How can I specify a route-specific middleware pipeline?
 
 Sometimes you may want to use a middleware pipeline only if a particular route
@@ -234,10 +233,7 @@ return [
 ];
 ```
 
-## How can I set 404 page?
-=======
 ## How can I set custom 404 page handling?
->>>>>>> wording current 404 for custom
 
 We can set custom 404 page handling instead of [using the final handler](http://zend-expressive.readthedocs.org/en/latest/error-handling/) for logging or other purpose with create new middleware that we want to intercept if no other middleware has executed, and is indicate a "not found" situation.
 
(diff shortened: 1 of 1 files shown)


doc/book/cookbook.md
```diff
@@ -44,3 +44,58 @@ $app->run();
...
+class NotFound
+{
+    public function __invoke($req, $res, $next)
+    {
+        return $next($req, $res->withStatus(404), 'Page Not Found');
+    }
+}
+```
+
+After it, we need to register the middleware into `invokables` type in our service config:
```
weierophinney: This is specific to a single container implementation, zend-servicemanager. We should likely either (a) provide setup for multiple containers, or (b) leave container configuration as an exercise for the reader.

weierophinney:
> Note:
> defining 'error' => true in middleware_pipeline config make it not working. Please let me know if I missed something. Thanks.

You did. Error middleware in Stratigility and Expressive must have the following signature:

```php
function ($error, ServerRequestInterface $request, ResponseInterface $response, callable $next)
```

Note the first argument!

The way you have set this up is correct; it's not actually error middleware; it's middleware that you want to intercept if no other middleware has executed by that point, and is indicative of a "not found" situation.

That said, this is only one of several approaches.

The current recommend approach is to use the final handler. The final handler is invoked when the stack is exausted (i.e., no middleware returns a response), and that can then determine if we have a 404 situation. The [TemplatedErrorHandler](http://zend-expressive.readthedocs.org/en/latest/error-handling/#templated-errors) will use a template for rendering the 404 page. This allows us to have something baked into the library for the situation.

The approach you're suggesting here is more suited for custom 404 handling; e.g., you may want to log such requests, or provide suggestions based on the URI for other pages, etc. In those cases, I'd use middleware. For a generic 404 page, however, the final handler approach is easier, and already provided to the user.

So, my suggestions:

- Per inline comments, either remove the container configuration, or cover other containers as well as zend-servicemanager.
- Detail *why* you might want an alternative 404 handler, instead of using the final handler.

doc/book/cookbook.md
samsonasik: updated.

samsonasik:
I did [The TemplatedErrorHandler](http://zend-expressive.readthedocs.org/en/latest/error-handling/#templated-errors) example in skeleton, and it shows blank page, even I create a template file under `templates/error` named `404.php`. @weierophinney could you relieve it ? thanks ;)

weierophinney:
@samsonasik the 404 condition may be related to #93; looking into that today.

samsonasik:
@weierophinney it works now :+1: , I've updated the wording.

samsonasik:
@weierophinney merge-able now ?

weierophinney:
@samsonasik I'm still evaluating.

As noted above, I think the recommendation for 404 pages should be to use one of the final handler implementations we ship. However, I *do* think it would make sense to have some examples of different types of error handlers.

As an example, on my site, I have an "Unauthorized" error handler that looks to see if the response status has been set to 401 or 403 and/or certain crtieri; in such a case, it then renders a template with a login form and sets the appropriate status. That sort of approach would be a useful example for the cookbook.

samsonasik:
rebased.

doc/book/cookbook.md
```diff
@@ -228,7 +228,55 @@ return [
...
+        ],
+    ],
+],
+```
+
+The other thing that we can do is programmatically pipe in `public/index.php`:
+
+```php
+$app->pipe($services->get('Application\NotFound'));
+$app->run();
```
weierophinney: Just a quick note: the above only works if one of the following criteria are met:

- You have first called `$app->pipeRoutingMiddleware()`; or
-  if you pulled the `Application` instance from the container, and used the `ApplicationFactory` shipped with Expressive (which ensures the routing middleware has been added to the pipeline).

I'll add that note when merging.

doc/book/cookbook.md
```diff
@@ -228,7 +228,55 @@ return [
...
+{
+    public function __invoke($req, $res, $next)
+    {
+        // other things can be done here, for eg: logging
+        return $next($req, $res->withStatus(404), 'Page Not Found');
+    }
+}
+```
+
+> We can register the `Application\NotFound` instance as a service in [service container](https://github.com/zendframework/zend-expressive/blob/master/doc/book/container/intro.md).
```
weierophinney: Use a relative link here (`../container/intro.md`); I'll make that change on merge.

doc/book/cookbook.md
```diff
@@ -228,7 +228,55 @@ return [
...
+## How can I set custom 404 page handling?
+
+We can set custom 404 page handling instead of [using the final handler](http://zend-expressive.readthedocs.org/en/latest/error-handling/) for logging or other purpose with create new middleware that we want to intercept if no other middleware has executed, and is indicate a "not found" situation.
+
+Let's create a `NotFound` middleware:
+
+```php
+namespace Application;
+
+use Zend\Expressive\Template\TemplateInterface;
```
weierophinney: `s/TemplateInterface/TemplateRendererInterface/`; I'll make that change on merge.

doc/book/cookbook.md
```diff
@@ -228,7 +228,55 @@ return [
...
             ],
-        ], 
+        ],
     ],
 ];
 ```
+
+## How can I set custom 404 page handling?
+
+We can set custom 404 page handling instead of [using the final handler](http://zend-expressive.readthedocs.org/en/latest/error-handling/) for logging or other purpose with create new middleware that we want to intercept if no other middleware has executed, and is indicate a "not found" situation.
```
weierophinney: Two things:

- First, use a relative link here: `../error-handling/`; I  can fix that on merge.
- Second, the example you have below is not actually by-passing the final handler. What you're doing below is setting the response status, and, by calling `$next()` with an error condition, delegating to the next error middleware present and/or the final handler itself. As such, the verbiage above is misleading, and/or the example below is misleading.

I think we need to clarify which approach you want to illustrate, and make changes to ensure the narrative and example are consistent.

samsonasik: @weierophinney I'm not quite sure about final handler part, would you mind to tackle for it ? Thank you ^^

weierophinney: @samsonasik Sure.

PR #82, opened 2015-08-23
Allow providing route names via configuration

weierophinney:
New config key in a route definition: "name"; if present, it will be used to set the route name.


weierophinney:
Essentially, we forgot to update the Application factory when we added route name support! :)

PR #86, opened 2015-08-24
Update Aura.php

spiffyjr:
PhpStorm will actually highlight this as an error if you do not include the leading slash.

**EDIT** Went through a few more files and made generic docblock updates (and removed some unused imports).

### DIFF
--- a/src/Router/Aura.php
+++ b/src/Router/Aura.php
@@ -24,7 +24,7 @@ class Aura implements RouterInterface
     /**
      * Aura router
      *
-     * @var \Aura\Router\Router
+     * @var Router
      */
     private $router;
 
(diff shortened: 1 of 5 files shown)


src/Router/Aura.php
```diff
@@ -24,7 +24,7 @@ class Aura implements RouterInterface
     /**
      * Aura router
      *
-     * @var Aura\Router\Router
+     * @var \Aura\Router\Router
```
moderndeveloperllc: It would be better to use `AuraRoute` as this class has already been imported with `use Aura\Router\Route as AuraRoute;`

moderndeveloperllc: Just use `Router`. It's already been imported with `use Aura\Router\Router;`

PR #85, opened 2015-08-24
Aura\Router error with no get('/') registered

weierophinney:
```Fatal error: Call to a member function failedMethod() on null in vendor\zendframework\zend-expressive\src\Router\Aura.php on line 156```


spiffyjr:
:+1: commit looks good.

spiffyjr:
I did not know GitHub could convert issues to PRs.

weierophinney:
@spiffyjr You can do it via the API, which the `hub` tool uses. :)

PR #92, opened 2015-08-25
Implement middleware lazy-loading

weierophinney:
Prior to this patch, the ApplicationFactory implemented lazy loading of pipeline middleware. The feature is very useful, as it delays fetching of middleware until execution by wrapping the call to fetch and invoke it in a closure.

This patch pushes that feature into `Application` itself:

- `pipe()` now allows specifying middleware by service name.
- A new method, `pipeErrorHandler()`, proxies to `pipe()`, but wraps lazy-middleware in a wrapper that uses the error middleware signature.

`ApplicationFactory` has been updated to use these facilities, and removes its own. Documentation has been updated to indicate the lazy-middleware features.

### DIFF
--- a/doc/book/application.md
+++ b/doc/book/application.md
@@ -135,14 +135,37 @@ applications by subpath.
 The signature of `pipe()` is:
 
 ```php
-public function pipe($pathOrMiddleware, callable $middleware = null)
+public function pipe($pathOrMiddleware, $middleware = null)
 ```
 
 where:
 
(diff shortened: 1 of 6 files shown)


spiffyjr:
Very nice feature.

codeliner:
Wait! You use the container as a service locator ...
Oh sorry, this isn't the FIG mailing list. Forget my last sentence. I love this feature :+1: :smile:

PR #37, opened 2015-08-25
Move router and service to require-dev

weierophinney:
Hi, 

As the router and container can be moved what about moving those to require-dev and have a project/skelton repo which comes with service and router by default ?

Else the problem is we install the extra components and will not be using those.

Thank you.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,15 +11,15 @@ to provide a minimalist PSR-7 middleware framework for PHP, with the following
 features:
 
 - Routing. Choose your own router; we support:
-  - [Aura.Router](https://github.com/auraphp/Aura.Router)
-  - [FastRoute](https://github.com/nikic/FastRoute)
-  - [ZF2's MVC router](https://github.com/zendframework/zend-mvc)
+    - [Aura.Router](https://github.com/auraphp/Aura.Router)
+    - [FastRoute](https://github.com/nikic/FastRoute)
+    - [ZF2's MVC router](https://github.com/zendframework/zend-mvc)
 - DI Containers, via [container-interop](https://github.com/container-interop/container-interop).
   Middleware matched via routing is retrieved from the composed container.
 - Optionally, templating. We support:
-  - [Plates](http://platesphp.com/)
-  - [Twig](http://twig.sensiolabs.org/)
-  - [ZF2's PhpRenderer](https://github.com/zendframework/zend-view)
+    - [Plates](http://platesphp.com/)
+    - [Twig](http://twig.sensiolabs.org/)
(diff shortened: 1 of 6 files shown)


weierophinney:
I've created a pull request to do this, as we definitely want to support developers *choosing* which router, container, and/or template system they wish to use. Ideally, we'll be able to create composer scripts that will allow users to select one or another, likely as part of the skeleton application.

The attached patch pushes the default router and container to `require-dev` and `suggest`, and updates the quick start to add `aura/router` to the `composer require` line (as well as lines in the README and container pages detailing supported adapters and how to install them).

README.md
```diff
@@ -11,15 +11,15 @@ to provide a minimalist PSR-7 middleware framework for PHP, with the following
...
+    - [ZF2's MVC router](https://github.com/zendframework/zend-mvc)
 - DI Containers, via [container-interop](https://github.com/container-interop/container-interop).
   Middleware matched via routing is retrieved from the composed container.
 - Optionally, templating. We support:
-  - [Plates](http://platesphp.com/)
-  - [Twig](http://twig.sensiolabs.org/)
-  - [ZF2's PhpRenderer](https://github.com/zendframework/zend-view)
+    - [Plates](http://platesphp.com/)
+    - [Twig](http://twig.sensiolabs.org/)
+    - [ZF2's PhpRenderer](https://github.com/zendframework/zend-view)
```
weierophinney: I added indentation on the second-level items above in hopes that RTD will properly indent them.

weierophinney:
Ping @harikt and @codeliner - Please review. :)

composer.json
```diff
@@ -11,23 +11,23 @@
     ],
     "require": {
         "php": ">=5.5",
-        "aura/router": "^2.3",
         "container-interop/container-interop": "^1.1",
+        "filp/whoops": "^1.1",
```
harikt: I suggest moving this out for in a production environment we don't need this beautiful stuff ;) .

harikt:
Thank you @weierophinney for hearing the user opinions :) . I am also looking how to send a PR for Aura.View as a template engine. Will open another issue.

weierophinney:
@harikt done. :)

codeliner:
@weierophinney Great news :+1: 
The `require` section in composer.json looks very lean now.
And very good documentation (of the entire project). Can't wait for the first version :smile:

harikt:
@weierophinney cool. Thank you.

spiffyjr:
+1 to this change. If you're going to allow/promote developer choice then it should be done through the entire stack (including comopser deps).

PR #93, opened 2015-08-25
404 is never called from FinalHandler

weierophinney:
The Stratigility FinalHandler never calls `create404` because of [this line](https://github.com/zendframework/zend-stratigility/blob/716c647fb76303c76c0c254bb4c5bc628aefa7e1/src/FinalHandler.php#L83). In [getFinalHandler()](https://github.com/zendframework/zend-expressive/blob/23c48d8f770fd73a062b903682bbf6d4aff6918c/src/Application.php#L371) of Expressive the $response will be a Diactoros Response (or other PSR-7), which gets passed to the Final Handler, and then decorated by Stratigility.  The result is the FinalHandler response will **never** equal the invoked response so the $response is returned by the FinalHandler and the 404 is never shown.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -18,7 +18,6 @@ use Zend\Diactoros\Response;
 use Zend\Diactoros\Response\EmitterInterface;
 use Zend\Diactoros\Response\SapiEmitter;
 use Zend\Diactoros\ServerRequestFactory;
-use Zend\Stratigility\FinalHandler;
 use Zend\Stratigility\MiddlewarePipe;
 
 /**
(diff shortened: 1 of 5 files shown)


spiffyjr:
Confirmed fixed for my use-case (api, non-templated).

weierophinney:
I tried it with a templated error handler:

```php
use Zend\Expressive\Router\Aura;
use Zend\Expressive\Template\Plates;
use Zend\Expressive\TemplatedErrorHandler;

require 'vendor/autoload.php';

$templates = new Plates();
$templates->addPath('templates', 'error');
$finalHandler = new TemplatedErrorHandler(
    $templates,
    'error::404',
    'error::500'
);

$app = new Application(new Aura(), null, $finalHandler);

$app->get('/', function ($req, $res, $next) {
    $res->getBody()->write('Hello, world!');
    return $res;
});

$app->run();
```

and defined templates in `templates/404.php` and `templates/500.php` of the working directory.

Prior to the patch, I received empty 200 responses; with the changes, I receive 404 templated responses.

PR #96, opened 2015-08-25
Abstract layout capabilities in zend-view adapter

weierophinney:
This pull request better abstracts layout capabilities for the zend-view adapter. It does so by:

- Allowing passing a layout script name or ViewModel representing the layout as the second argument in the constructor.
- Allowing passing a "layout" parameter during rendering, which may be a layout script name or a ViewModel representing the layout.

In each case, rendering now munges the provided template and parameters into a view model, injecting it as a child of the layout view model if one is discovered (if a layout script name is passed, a layout view model is created to represent it).

This makes usage look more like the other adapters, and brings its capabilities to the same level of completion.

### DIFF
--- a/doc/book/template/zend-view.md
+++ b/doc/book/template/zend-view.md
@@ -59,24 +59,80 @@ $templates = new ZendView($zendView);
 ## Layouts
 
 Unlike the other supported template engines, zend-view does not support layouts
-out-of-the-box.
+out-of-the-box. Expressive abstracts this fact away, providing two facilities
+for doing so:
 
-Layouts are accomplished in one of two ways:
+- You may pass a layout template name or `Zend\View\Model\ModelInterface`
+  instance representing the layout as the second argument to the constructor.
+- You may pass a "layout" parameter during rendering, with a value of either a
+  layout template name or a `Zend\View\Model\ModelInterface`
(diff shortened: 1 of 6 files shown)


doc/book/template/zend-view.md
```diff
@@ -59,24 +59,80 @@ $templates = new ZendView($zendView);
...
+
+// Create the engine instance with a layout name:
+$zendView = new PhpRenderer(null, 'layout');
+```
+
+### Layout view model passed to constructor
+
+```php
+use Zend\Expressive\Template\ZendView;
+use Zend\View\Model\ViewModel
```
samsonasik: missing `;`

doc/book/template/zend-view.md
```diff
@@ -59,24 +59,80 @@ $templates = new ZendView($zendView);
...
+$content = $templates->render('blog/entry', [
+    'layout' => 'blog',
+    'entry'  => $entry,
+]);
+```
+
+### Provide a layout view model when rendering
+
+```php
+use Zend\View\Model\ViewModel
```
samsonasik: missing `;`

weierophinney:
@samsonasik Thanks -- fixed locally, and will include those fixes with merge.

PR #99, opened 2015-08-26
Fix example classname in documentation

moderndeveloperllc:
Fix for PSR-0/4 compliance. Important for those of us who like to copy/paste examples!


weierophinney:
Good catch; missed that I'd not kept the filename and classname in sync!

PR #100, opened 2015-08-26
Fix typo ZendView not PhpRenderer

harikt:


weierophinney:
This was fixed in 69eadac (part of a larger documentation polishing effort).

PR #105, opened 2015-08-26
Fixing code blocks and removing extra characters from docs.

kingjerod:
Only things removed/changed were removing the > characters and tabs. The extra tabs on the code blocks made each line into it's own code block.

### DIFF
--- a/doc/book/container/factories.md
+++ b/doc/book/container/factories.md
@@ -37,78 +37,78 @@ order to seed the `Application` instance:
 
 - `middleware_pipeline` can be used to seed pre- and/or post-routing middleware:
 
-  ```php
-  'middleware_pipeline' => [
-      // An array of middleware to register prior to registration of the
-      // routing middleware:
-      'pre_routing' => [
-      ],
-      // An array of middleware to register after registration of the
-      // routing middleware:
-      'post_routing' => [
-      ],
+```php
+'middleware_pipeline' => [
+  // An array of middleware to register prior to registration of the
+  // routing middleware:
+  'pre_routing' => [
   ],
-  ```
-
-  Each item of each array must be an array itself, with the following structure:
-
-  ```php
-  [
-      // required:
-      'middleware' => 'Name of middleware service, or a callable',
-      // optional:
-      'path'  => '/path/to/match',
-      'error' => true,
+  // An array of middleware to register after registration of the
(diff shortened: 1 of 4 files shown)


doc/book/container/factories.md
```diff
@@ -37,78 +37,78 @@ order to seed the `Application` instance:
...
-  ```php
-  'middleware_pipeline' => [
-      // An array of middleware to register prior to registration of the
-      // routing middleware:
-      'pre_routing' => [
-      ],
-      // An array of middleware to register after registration of the
-      // routing middleware:
-      'post_routing' => [
-      ],
```
weierophinney: I actually want to keep the original formatting here. While we're currently publishing to RTD, which will not render this correctly, our plan is to use [bookdown](http://bookdown.io) to publish docs on the ZF website. Bookdown uses CommonMark, which correctly interprets this as a code block, and, further, as a code block associated with the preceding bullet point, which is the intent.

doc/book/container/factories.md
```diff
@@ -37,78 +37,78 @@ order to seed the `Application` instance:
...
+```
 
-  The `middleware` key itself is the middleware to execute, and must be a
-  callable or the name of another defined service. If the `path` key is present,
-  that key will be used to segregate the middleware to a specific matched path
-  (in other words, it will not execute if the path is not matched). If the
-  `error` key is present and boolean `true`, then the middleware will be
-  registered as error middleware. (This is necessary due to the fact that the
-  factory defines a callable wrapper around middleware to enable lazy-loading of
-  middleware.)
```
weierophinney: Same thing here; this paragraph is part of the previous bullet point, and that was intentional.

doc/book/container/factories.md
```diff
@@ -37,78 +37,78 @@ order to seed the `Application` instance:
...
+  // An array of middleware to register after registration of the
+  // routing middleware:
+  'post_routing' => [
   ],
-  ```
+],
+```
 
-  Each item of each array must be an array itself, with the following structure:
+Each item of each array must be an array itself, with the following structure:
```
weierophinney: This is meant to be a nested bullet point.

weierophinney:
I'm closing this, per the comments made above.

kingjerod:
Gotcha, sounds good.

PR #108, opened 2015-08-27
make absolute link for doc tree and notes to make browseable in readthedocs.org

samsonasik:


ojhaujjwal:
:+1:

weierophinney:
No.

RTD is a temporary solution until we find a place for the docs on the ZF website; the only reason the README is rendered is so that we have an index page on RTD (without it, we have no RTD landing page, and developers would need to know a deep link to get anywhere into it).

If you render the documentation manually, the `bookdown.json` does not use the `README.md` at all, and thus those links go away.

Finally, making them absolute means they'll be incorrect for any forks or branches; there's a reason they're relative!

PR #116, opened 2015-08-29
Added the any() method.

spiffyjr:
This method is useful for defining a single middleware that will respond to any of the $httpRouteMethods. For example, a Restful controller that has logic to route to the appropriate method internally:

```
$app->any('/foo', Api\RestfulController::class);
```

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -163,7 +163,7 @@ class Application extends MiddlewarePipe
      */
     public function any($path, $middleware, $name = null)
     {
-        return $this->route($path, $middleware, $this->httpRouteMethods, $name);
+        return $this->route($path, $middleware, Router\Route::HTTP_METHOD_ANY, $name);
     }
 
     /**
--- a/src/Application.php
+++ b/src/Application.php
@@ -163,7 +163,7 @@ class Application extends MiddlewarePipe
      */
     public function any($path, $middleware, $name = null)
     {
-        return $this->route($path, $middleware, Router\Route::HTTP_METHOD_ANY, $name);
+        return $this->route($path, $middleware, null, $name);
     }
 
     /**
--- a/src/Application.php
+++ b/src/Application.php
@@ -156,6 +156,17 @@ class Application extends MiddlewarePipe
     }
 
     /**
+     * @param string|Router\Route $path
+     * @param callable|string     $middleware Middleware (or middleware service name) to associate with route.
+     * @param null|string         $name the name of the route
+     * @return Router\Route
+     */
+    public function any($path, $middleware, $name = null)
+    {
+        return $this->route($path, $middleware, $this->httpRouteMethods, $name);
+    }
+
+    /**
      * Overload pipe() operation.
      *
      * Middleware piped may be either callables or service names. Middleware
--- a/test/ApplicationTest.php
+++ b/test/ApplicationTest.php
@@ -68,6 +68,26 @@ class ApplicationTest extends TestCase
         $this->assertSame($this->noopMiddleware, $route->getMiddleware());
     }
 
+    public function testAnyRouteMethod()
+    {
+        $this->router->addRoute(Argument::type(Route::class))->shouldBeCalled();
+        $route = $this->getApp()->any('/foo', $this->noopMiddleware);
+        $this->assertInstanceOf(Route::class, $route);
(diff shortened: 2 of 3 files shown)


test/ApplicationTest.php
```diff
@@ -68,6 +68,26 @@ public function testRouteMethodReturnsRouteInstance()
...
 
+    public function testAnyRouteMethod()
+    {
+        $this->router->addRoute(Argument::type(Route::class))->shouldBeCalled();
+        $route = $this->getApp()->any('/foo', $this->noopMiddleware);
+        $this->assertInstanceOf(Route::class, $route);
+        $this->assertEquals('/foo', $route->getPath());
+        $this->assertSame($this->noopMiddleware, $route->getMiddleware());
+
+        $reflClass    = new \ReflectionClass($this->getApp());
```
samsonasik: move `ReflectionClass` to use statement

weierophinney:
That's what the route() method already does if no HTTP method are provided.
It an I missing something about your intent?

spiffyjr:
Oh, honestly, I just assumed that route() required methods. I should RTFM?

weierophinney:
@spiffyjr 

Relevant docs are:

- http://zend-expressive.readthedocs.org/en/latest/application/#adding-routable-middleware

Essentially, if you only pass the route and middleware to `route()`, it assumes any HTTP method; you can pass `null` or `Zend\Expressive\Router\Route::HTTP_METHOD_ANY` as the third argument to specify matching all methods, too.

That said, it's more verbose than your proposed `any()` method for situations when you want to specify a name, which might lend weight to adding this. Thoughts?

spiffyjr:
I'd agree. It's also similar to other frameworks where route is typically used for custom mapping, e.g., POST and PUT but not GET, etc.

weierophinney:
@spiffyjr Can you also update the `doc/book/application.md` file to indicate this new method, please?

spiffyjr:
I originally had an entire section on any() but removed in favor of this version. It's simple and the word "any" makes it obvious what it does.

src/Application.php
```diff
@@ -156,6 +156,17 @@ public function __call($method, $args)
...
 
     /**
+     * @param string|Router\Route $path
+     * @param callable|string     $middleware Middleware (or middleware service name) to associate with route.
+     * @param null|string         $name the name of the route
+     * @return Router\Route
+     */
+    public function any($path, $middleware, $name = null)
+    {
+        return $this->route($path, $middleware, $this->httpRouteMethods, $name);
```
weierophinney: Pass a `null` or `Router\Route::HTTP_METHOD_ANY` for the third argument. `$httpRouteMethods` only really exists to determine what methods are valid overloading methods.

test/ApplicationTest.php
weierophinney: @spiffyjr I agree with @samsonasik - but have a different suggestion:

- Import `ReflectionProperty` into the file.
- Use the following:
  ```php
  $app = $this->getApp();
  $r = new ReflectionProperty($app, 'httpRouteMethods');
  $r->setAccessible(true);
  $routeMethods = $r->getValue($app);
  ```

Since you're interested in a specific property, do reflection directly on it, instead of pulling it from class reflection. :smile:

spiffyjr:
I used Router\Route::METHOD_ANY for brevity.

weierophinney:
> for brevity

You mean "clarity," right? :) Looks good; will merge later today!

spiffyjr:
Dammit man, I'm a programmer not an English major. :D

weierophinney:
@spiffyjr Looks like you need to use `null`, and not the `Route::HTTP_METHOD_ANY` constant:

- https://travis-ci.org/zendframework/zend-expressive/jobs/78406676#L195

spiffyjr:
That's what I get for using the online editor.

weierophinney:
Still failing:

- https://travis-ci.org/zendframework/zend-expressive/jobs/78411695#L194

Essentially, you should be asserting against `Route::HTTP_METHOD_ANY` instead of the array of methods. (In other words, you can totally forget the `ReflectionProperty` stuff I mentioned in a previous comment.)

weierophinney:
Merged to develop for release with 0.2.0.

spiffyjr:
Having a kid at the moment so thanks for updating this.

PR #118, opened 2015-08-30
Route options not processed in ApplicationFactory

acelaya:
When a route instance is created by calling `$app->route(...)`, that instance is automatically added to the router, which tries to process the route options.
The problem is that the `ApplicationFactory` sets the options after that, which makes the router ignore any option defined for the route.

Creating the route as a new `Zend\Expressive\Router\Route` instance and passing it to the `$app->route(...)` method after processing the options fixes this problem.

### DIFF
--- a/src/Container/ApplicationFactory.php
+++ b/src/Container/ApplicationFactory.php
@@ -16,8 +16,8 @@ use Zend\Expressive\Application;
 use Zend\Expressive\Emitter\EmitterStack;
 use Zend\Expressive\Exception;
 use Zend\Expressive\Router\Aura as AuraRouter;
-use Zend\Expressive\Router\RouterInterface;
 use Zend\Expressive\Router\Route;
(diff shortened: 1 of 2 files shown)


acelaya:
Tests are not passing. I'll fix them and try to create some that cover this case.

src/Container/ApplicationFactory.php
```diff
@@ -17,6 +17,7 @@
 use Zend\Expressive\Exception;
 use Zend\Expressive\Router\Aura as AuraRouter;
 use Zend\Expressive\Router\RouterInterface;
+use Zend\Expressive\Router\Route;
```
samsonasik: sort the use statement, place `Zend\Expressive\Router\Route` before `Zend\Expressive\Router\RouterInterface`

acelaya: Done!

Issue #119, opened 2015-08-31
Renaming route to router?


bakura10:
Hi !

While reading some part of the source code, I've felt a bit confused by the naming of the various router adapters.

Actually, it seems that the various adapters are actually router (a list of route, and can be matched), but all the adapter are actually named "AuraRoutE", "FastRoutE", "Zf2RoutE". That's a bit confusing because in my head, a route is... well, a single route.

I'd suggest therefore renaming:

* AuraRoute => AuraRouter
* FastRoute => FastRouter
* Zf2 => Zf2Router (for consistency)

I know a version has already been tagged, but I think this would be a sane change to reduce confusness (and I really was ^^).

weierophinney:
The naming is a result of FastRoute, as that was the first we implemented,
and that's the package name.

I think a name change would be fine; the tag was in the 0.* series, after
all, so API changes are still allowed.

Want to create a PR?

If so, we should also standardize template implementation names.

bakura10:
Yep. I will do that tomorrow. What about FastRouteRouter ? This way we stay entirely consistent.

Envoyé de mon iPhone

weierophinney:
Fixed with #120.

PR #120, opened 2015-09-01
Rename router

bakura10:
As per #119 , this appends the "Router" word after all "router" class, to avoid ambiguousness between Route and Router.

### DIFF
--- a/src/Container/ApplicationFactory.php
+++ b/src/Container/ApplicationFactory.php
@@ -11,11 +11,9 @@ namespace Zend\Expressive\Container;
 
 use Interop\Container\ContainerInterface;
 use Zend\Diactoros\Response\EmitterInterface;
-use Zend\Diactoros\Response\SapiEmitter;
 use Zend\Expressive\Application;
-use Zend\Expressive\Emitter\EmitterStack;
 use Zend\Expressive\Exception;
-use Zend\Expressive\Router\AuraRouter as AuraRouter;
+use Zend\Expressive\Router\AuraRouter;
 use Zend\Expressive\Router\RouterInterface;
 
 /**
--- a/src/Container/ApplicationFactory.php
+++ b/src/Container/ApplicationFactory.php
@@ -15,7 +15,7 @@ use Zend\Diactoros\Response\SapiEmitter;
 use Zend\Expressive\Application;
 use Zend\Expressive\Emitter\EmitterStack;
 use Zend\Expressive\Exception;
-use Zend\Expressive\Router\Aura as AuraRouter;
+use Zend\Expressive\Router\AuraRouter as AuraRouter;
 use Zend\Expressive\Router\RouterInterface;
 
 /**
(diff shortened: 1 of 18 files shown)


test/RouteMiddlewareTest.php
```diff
@@ -265,9 +265,9 @@ public function testRoutingSuccessResultingInContainerExceptionReRaisesException
     public function routerAdapters()
     {
         return [
-          'aura'       => [ 'Zend\Expressive\Router\Aura' ],
-          'fast-route' => [ 'Zend\Expressive\Router\FastRoute' ],
-          'zf2'        => [ 'Zend\Expressive\Router\Zf2' ],
+          'aura'       => [ 'Zend\Expressive\Router\AuraRouter' ],
```
bakura10: Héhé, this is why ::class is so much better even in tests. Make refactoring much easier! ;)

src/Container/ApplicationFactory.php
```diff
@@ -15,7 +15,7 @@
 use Zend\Expressive\Application;
 use Zend\Expressive\Emitter\EmitterStack;
 use Zend\Expressive\Exception;
-use Zend\Expressive\Router\Aura as AuraRouter;
+use Zend\Expressive\Router\AuraRouter as AuraRouter;
```
samsonasik: no need alias as already named `AuraRouter`

test/Router/Zf2RouterTest.php
```diff
@@ -12,11 +12,11 @@
 use PHPUnit_Framework_TestCase as TestCase;
 use Prophecy\Argument;
 use ReflectionProperty;
-use Zend\Expressive\Router\Zf2 as Zf2Router;
+use Zend\Expressive\Router\Zf2Router as Zf2Router;
```
moderndeveloperllc: Another spot that aliasing can be removed.

test/Router/AuraRouterTest.php
```diff
@@ -10,10 +10,10 @@
 namespace ZendTest\Expressive\Router;
 
 use PHPUnit_Framework_TestCase as TestCase;
-use Zend\Expressive\Router\Aura as AuraRouter;
+use Zend\Expressive\Router\AuraRouter as AuraRouter;
```
moderndeveloperllc: Alias removal.

bakura10:
Thanks!

samsonasik:
@bakura10 it need rebase now ;)

test/RouteMiddlewareTest.php
weierophinney: Since this is the only place they're referenced inside this classfile, it's actually probably simpler this way; otherwise you would have had to change the import statements **and** this code. ;-)

weierophinney:
Merged to develop for release with 0.2.0.

weierophinney:
@samsonasik Already took care of it on merge. :smile:

PR #134, opened 2015-09-16
add getRouter() function in Routers

samsonasik:
I think this may be useful to add `getRouter()` so in real app we can use more function from it's router property when needed.

### DIFF
--- a/src/Router/AuraRouter.php
+++ b/src/Router/AuraRouter.php
@@ -67,6 +67,14 @@ class AuraRouter implements RouterInterface
     }
 
     /**
+     * @return Router
+     */
+    public function getRouter()
+    {
+        return $this->router;
+    }
+
+    /**
      * Add a route to the underlying router.
      *
      * Adds the route to the Aura.Router, using the path as the name, and a
--- a/src/Router/FastRouteRouter.php
+++ b/src/Router/FastRouteRouter.php
@@ -158,6 +158,14 @@ class FastRouteRouter implements RouterInterface
     }
 
     /**
+     * @return RouteCollector
+     */
+    public function getRouter()
+    {
+        return $this->router;
+    }
+
+    /**
      * Retrieve the dispatcher instance.
      *
      * Uses the callable factory in $dispatcherCallback, passing it $data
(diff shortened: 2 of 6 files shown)


bakura10:
Maybe renaming it to `getInnerRouter` ? $router->getRouter() is a bit misleading.

weierophinney:
Can you provide some use cases? DI should obviate the need to do that, and
usage of containers with delegator functionality will allow configuration
after normal creation...

samsonasik:
@bakura10 I can rename to that of course if it accepted
@weierophinney for example: I need to call `generateRaw()` as I uses `Aura\Router\Router`, We can already inject the middleware with `Zend\Expressive\Router\RouterInterface::class` as service, and from the router injected, I can call `$this->router->getRouter()->generateRaw('routename')`

weierophinney:
How is generateRaw() different from generateUrl()? This may be an
indication that we need an additional method in the router interface...

samsonasik:
generateRaw() generate the route without url encoding. Add addiitonal method in interface for what ? `getRouter()` or `generateRaw()` ? I agree for `getRouter`, but If `generateRaw()`, then we need to maintain for different adapters.

weierophinney:
I'm thinking we should assume that router implementations return unencoded (raw) URIs, and that users should encode manually, after retrieval, when necessary. I'd argue that this path would be best, as the original data is typically easy to get, and applying an escaping mechanism to it after-the-fact based on context would allow keeping the encoding functionality completely out of the `RouterInterface` functionality.

As such, I'd argue that we need a patch that does the following:

- Stipulates in the `generateUri()` docblock that the URI returned **MUST NOT** be encoded.
- Updates the `Aura.Router` implementation to use `generateRaw()` internally.

Does that seem reasonable @samsonasik ?

samsonasik:
my initial idea about it is not about specific `generateRaw()`, it was just in case we need to call another method from the internal router, we can call via `getRouter()` or `getInternalRouter()` without need to extends existing router adapter.

weierophinney:
In that case, you're depending on a specific, concrete adapter, and you
might as well inject your object with the underlying router if you need
that functionality.

samsonasik:
ok ;)

Issue #136, opened 2015-09-21
Renaming "Template" to "TemplateRenderer" or "TemplateEngine"?


danizord:
Maybe it's just me, but I think the current naming sounds a bit confusing.

The `TemplateInterface` currently exposes the public method `render($name, $params = [])`, but a template does not render stuff, it actually gets rendered by a renderer or template engine.

Should we rename `TemplateInterface` to `TemplateEngineInterface` or `TemplateRendererInterface`?

bakura10:
+1. I was a bit surprised too by the naming (same as it was with "Route" defining actually a "Router"). I like TemplateEngineInterface.

weierophinney:
Funny you both noticed that; I've been kicking myself for the naming recently myself. :)

I think `TemplateRendererInterface` makes most sense at this point, as it doesn't model an engine, but a renderer.

@danizord — care to create a pull request?

ojhaujjwal:
I was also a bit surprised by the fact that it is called Template and not TemplateRenderer

weierophinney:
Closed with #139.

PR #139, opened 2015-09-26
Renamed Template to TemplateRenderer

ojhaujjwal:
As discussed in #136, I have renamed all the occurance of `Template` to `TemplateRenderer` and `$template` to `$renderer`

### DIFF
--- a/doc/book/template/interface.md
+++ b/doc/book/template/interface.md
@@ -71,7 +71,7 @@ more paths to templates; these are then used when resolving a template name to
 the actual template. You may use the `addPath()` method to do so:
 
 ```php
-$templateRenderer->addPath('templates');
+$renderer->addPath('templates');
 ```
 
 Template engines adapted for zend-expressive are also required to allow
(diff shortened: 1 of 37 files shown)


src/Template/Plates.php
```diff
@@ -15,7 +15,7 @@
 /**
  * Template implementation bridging league/plates
  */
-class Plates implements TemplateInterface
+class Plates implements TemplateRendererInterface
```
bakura10: Similarily to Route, I think we should suffix by Renderer (PlatesRendere, ZendRenderer, TwigRenderer...)

doc/book/template/interface.md
```diff
@@ -71,7 +71,7 @@ more paths to templates; these are then used when resolving a template name to
 the actual template. You may use the `addPath()` method to do so:
 
 ```php
-$template->addPath('templates');
+$templateRenderer->addPath('templates');
 ```
```
weierophinney: Too verbose; please use `$renderer` instead.

src/Template/Plates.php
ojhaujjwal: Agree

doc/book/template/interface.md
```diff
@@ -1,13 +1,13 @@
 # The Template Interface
```
weierophinney: Note to self: update this title to "The Template Renderer Interface" on merge.

Issue #142, opened 2015-09-28
Adding parameters to templates before render-time


kynx:
Right now we've only got one chance to add parameters to templates, via `render($name, $params)`.

However it'd be really useful to add data in stages. So a piped middleware could be responsible for populating standard stuff like the current user's name, then the routed middleware adds the stuff it's responsible for.

Plates offers [addData()](http://platesphp.com/templates/data/) that does just this. From what I can tell it's possible in Zend View as well. Twig doesn't, but it could be faked easily enough by storing the data in the Twig bridge until render time.

Anyone else think this is good idea?

akrabat:
Twig has an `addGlobal()` method for this.

weierophinney:
@akrabat - I missed that when we were updating the twig implementation.
Could you maybe do a PR or open an issue against
zend-expressive-twigrenderer to use that instead of the current approach?

akrabat:
Looking at how zend-expressive-zendview is implemented, I'd say that the way the Twig one is done is correct as this way we have per-template default parameters which isn't supported by Twig's `addGlobal`.

However, I don't see a way to add a parameter to every template as per the example in this issue?

akrabat:
Ah -  there's a weird template name called "TEMPLATE_ALL"… I would *never* have guessed that!

weierophinney:
Evidently I need to make that more prominent in the docs... Any suggestions?

We could still have that proxy to twig's global, while providing the
abstraction for per-template. Thoughts?

weierophinney:
Closed with #143

PR #143, opened 2015-09-29
Adding parameters to templates before render-time

kynx:
My first run at #142. Only tried it with Plates so far. There's probably a better way to do this with ZendView, but I'm a little hazy on that. Let me know if I'm on the right track and I'll do the tests and docs.

### DIFF
--- a/src/Template/AddParametersTrait.php
+++ b/src/Template/AddParametersTrait.php
@@ -1,0 +1,51 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
+ * @copyright Copyright (c) 2015 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   https://github.com/zendframework/zend-expressive/blob/master/LICENSE.md New BSD License
+ */
+
+namespace Zend\Expressive\Template;
+
+
+trait AddParametersTrait
+{
+    /**
+     * @var array
+     */
+    protected $templateParams = [];
+
+    /**
+     * Add parameters to template
+     *
+     * If no template name is given, the parameters will be added to all templates rendered
+     *
+     * @param array|object $params
+     * @param string $name
+     */
+    public function addParameters($params, $name = '')
+    {
+        if (method_exists($this, 'normalizeParams')) {
+            $params = $this->normalizeParams($params);
+        }
+        $this->templateParams[$name] = array_merge($this->templateParams[$name] ?: [], $params, $name);
+    }
+
+    /**
+     * Returns merged global, template-specific and given params
+     *
+     * @param array $params
+     * @param string $name
+     * @return array
+     */
+    protected function mergeParams($params, $name)
+    {
+        return array_merge(
+            $this->templateParams[''] ?: [],
+            $this->templateParams[$name] ?: [],
+            $params
+        );
+    }
+}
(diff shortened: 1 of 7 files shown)


src/Template/AddParametersTrait.php
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+     * Add parameters to template
+     *
+     * If no template name is given, the parameters will be added to all templates rendered
+     *
+     * @param array|object $params
+     * @param string|null $name
+     */
+    public function addParameters($params, $name = null)
```
weierophinney: This method also needs to be added to the interface; otherwise, an implementation could omit it, leaving developers depending on concretions instead of abstractions.

src/Template/AddParametersTrait.php
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+namespace Zend\Expressive\Template;
+
+trait AddParametersTrait
+{
+    /**
+     * @var array
+     */
+    protected $templateParams = [];
```
weierophinney: Since these members (this and the `mergeParameters()` method) are in the trait, they can be `private` visibility.

weierophinney:
Some general notes:

- Terminology needs to be consistent. You have `$templateParams`, `addParameters()`, and `mergeParams()` — three different ways ways of referencing related things. `$name` is unclear as an argument; are you indicating a parameter name, or a template name? Additionally, these are *default* parameters to use, not the *only* parameters. As such, I'd go with:
  - `private $defaultParams = []`
  - `public function setDefaultParams(array $params, $templateName = null)` (more on this method later)
  - `private function mergeParams(array $params, $templateName = null)`
- I'd shy away from public methods that work with a collection, such as `addParameters()`. In my experience, these methods cause a fundamental confusion: does it merge or overwrite an existing set of parameters? One way of solving this is to do as we did in PSR-7 with attributes. We omitted `withAttributes()`, and instead provided only `withAttribute()` and `withoutAttribute()`; the first overwrites or creates the single name header, while the second removes it. I suggest doing the same here:
  - `public function addDefaultParam($templateName, $name, $value)`
  - `public function removeDefaultParam($templateName, $name)`
- Related to the previous point, I'd *require* the template name, and provide a constant in the template interface for specifying "all templates"; users who want to specify a global default template parameter would use something like `$renderer->addDefaultParam($renderer::ALL_TEMPLATES, 'url', $url)`.

Finally, you'll need to rebase, as #139 renamed `TemplateInterface` to `TemplateRendererInterface`, and appended the name `Renderer` to all impementations.

With regards to the ZF2 implementation, I'd handle it in `createModel()`, merging the default parameters with those passed in.

kynx:
Thanks for the feedback! I'll get on with that then.

weierophinney:
@kynx I've been working on the suggestions this morning as well; I can do a PR against your branch shortly, if you'd like.

kynx:
@weierophinney Please do! Looks like I messed up the rebase tho :(

weierophinney:
@kynx — I've created #151 so I could run Travis-CI checks; at this point, those are all passing. I've also added documentation of the new interface method.

We have two options at this point:

- I can merge #151, closing both that issue and this one.
- You can merge my branch to yours, push your branch, and I can merge this PR and close #151; this would be particularly useful if you have any additions or changes you want to include beyond what I've done.

Let me know!

weierophinney:
@kynx I've gone ahead and merged both branches at this time; thanks for the new feature!

PR #144, opened 2015-10-01
Fixed quick start documentation require statement

AndrewCarterUK:
With this directory structure:

composer.json
composer.lock
vendor/autoload.php
vendor/...
public/index.php

The require statement needs to step back a level before stepping into the vendor directory.

At least, I needed to make this change to make it work on my machine.


marciovinicius:
@AndrewCarterUK Do you have used chdir(__DIR__); or chdir(dirname(__DIR__)); ? 
The second option works fine for me.

AndrewCarterUK:
Good shout, apologies!

PR #145, opened 2015-10-05
Renamed 'Zend\Expressive\FinalHandler' to 'Zend\Stratigility\FinalHandler'

AndrewCarterUK:
I'm not actually in a position to run the tests at the moment. I'm creating this as a PR because it's a bit more helpful than creating an issue and I'm hoping a bot will tell me if the tests pass when I submit!

Essentially, 'Zend\Expressive\FinalHandler' doesn't exist, and as such violates the service naming convention suggested in the docs. The actual path is 'Zend\Stratigility\FinalHandler'.

### DIFF
--- a/doc/book/container/factories.md
+++ b/doc/book/container/factories.md
@@ -18,7 +18,7 @@ namespace, and define an `__invoke()` method that accepts an
     - `Zend\Expressive\Router\RouterInterface`. When provided, the service will
       be used to construct the `Application` instance; otherwise, an Aura router
       implementation will be used.
-    - `Zend\Expressive\FinalHandler`. This is a meta-service, as the only concrete
+    - `Zend\Stratigility\FinalHandler`. This is a meta-service, as the only concrete
       type required is a callable that can be used as a final middleware in the
       case that the stack is exhausted before execution ends. By default, an
       instance of `Zend\Stratigility\FinalHandler` will be used.
@@ -113,7 +113,7 @@ order to seed the `Application` instance:
 ## TemplatedErrorHandlerFactory
 
 - **Provides**: `Zend\Expressive\TemplatedErrorHandler`
-- **Suggested Name**: `Zend\Expressive\FinalHandler`
+- **Suggested Name**: `Zend\Stratigility\FinalHandler`
 - **Requires**: no additional services are required.
 - **Optional**:
     - `Zend\Expressive\Template\TemplateInterface`. If not provided, the error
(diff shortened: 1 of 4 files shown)


samsonasik:
it's not, already right as @weierophinney  said in #18

AndrewCarterUK:
That's very confusing!

If Zend Expressive will provide it's own final handler before the first stable release, then it makes sense to keep using it as a service name. Otherwise I struggle to see the logic.

In it's current form, it stands out as a glaring inconsistency. If the service name should be a class path that tells us something about the service, then it should reference a class path that actually exists. I don't think using callables as a final handler is a violation of this - the actual class path still tells us information about what the service does.

If it is the case that the final handler service requires different treatment than the other services, why use such a confusing service name? If the naming convention does not apply here, a service name should be used that makes it clear that we are not following the previous convention. In it's current form it makes us think that it is using the convention defined in the documentation, when actually it has no attachment to it at all.

weierophinney:
> If Zend Expressive will provide it's own final handler before the first stable release, then it makes sense to keep using it as a service name. 

We already do. We have two different versions:

- `TemplatedErrorHandler`, which is intended for production, and
- `WhoopsErrorHandler`, which is intended for development.

Regarding this statement:

> If the service name should be a class path that tells us something about the service

I'd argue that service names should reference *interfaces* whenever possible, to hint to the *capabilities* a service provides. That leaves us with a quandary, however: Stratigilty does not define an interface for "final handlers"; essentially, anything implementing the `$next()` contract of `function (ServerRequestInterface $request, ResponseInterface $response, $error = null)` can be used as a final handler. `Zend\Stratigility\FinalHandler` is simply a concrete implementation of it. (Stratigility does not define an interface for it, as it allows *any* callable to be used; the callable *should*, of course, accept that signature, but requiring an interface limits interoperability here.)

The two implementations I indicate above *also follow the signature*, and are specific to Expressive. I specifically did not use the name `Zend\Stratigility\FInalHandler` as I wanted to indicate that final handlers should provide better/more capabilities than that component provides.
`Zend\Stratigility\FinalHandler` *can* be used here, of course, but I *do not* want to specify an inheritance tree; neither of the implementations we provide in Expressive *extend* that class.

As such, the service name is correct as-is, and will not change at this time.

AndrewCarterUK:
Respectfully disagree.

I can see from your explanation why you are not using 'Zend\Stratigility\FinalHandler' as the service name - but I do not think that implies that switching 'Stratigility' for 'Expressive' makes the service name sensible.

I'd just re-quote my last point:

"If it is the case that the final handler service requires different treatment than the other services, why use such a confusing service name? If the naming convention does not apply here, a service name should be used that makes it clear that we are not following the previous convention. In it's current form it makes us think that it is using the convention defined in the documentation, when actually it has no attachment to it at all."

As middleware that does not implement an interface will be quite common - this is actually a really good opportunity to introduce a new naming convention for these types of services.

"@final_handler"
"@final-handler"
"final_handler"
"final-handler"

etc...

As long as it can't be confused as following the previous convention.

AndrewCarterUK:
I'd just like to add that I'm happy to submit a different PR for this which includes an explanation in the documentation if you agree?

PR #147, opened 2015-10-06
Fix docummentation example

malukenho:


weierophinney:
The example is wrong, but not in the way you think. :)

`Server` is the correct class here; the problem is it *also* needs a `ServerRequestFactory`. I'll submit an alternate PR momentarily.

weierophinney:
See #148.

Issue #131, opened 2015-09-14
This repo seems to be the next monolithic library.


Maks3w:
It's good to provide support for the most common/popular third party implementations but I think each one of them should be developed on individual packages.

danizord:
:+1:

codeliner:
:+1: 

A decision for this request will influence config approach discussed [here](https://github.com/zendframework/zend-expressive/pull/128#issuecomment-140098279)

weierophinney:
I need way more detail. What exactly do you feel is becoming monolithic? How do you propose changing the situation?

We have a number of implementations by default to (a) reduce dependencies, and (b) provide something that works out-of-the-box. We do not plan on adding more and more to these; they are mainly present to achieve those two objectives, as well as provide a blue-print for how others may develop conforming implementations. In point of fact, the proposed installer *already allows you to provide alternate packages for each dependency*.

I've even created a separate package myself: https://github.com/phly/phly-expressive-mustache

So, knowing that: what do you feel needs to change *at this time*?

Maks3w:
My feelings are the same as ZF2 discussion of components plus the discussion about zend-validator split.

Different components will improve the development release cycle (We are seeing the effect on various ZF components reaching 2.6 and more)

Prevent things like #129 (well, only prevent for to discuss it here)

Allow to quickly deprecate third party libraries and reduce the maintenance over the adapter.

At least I expect to split out those libraries before this project become stable.

harikt:
@weierophinney I don't know you like my idea. I did proposed one earlier. This is what I feel currently seeing the issue.

1. Move template / dependency injection container to septate bundles.
1.a This will bring something like `zend-expressive-twig`, `zend-expressive-plates` etc.
2. This brings another question of how do you handle the configuration
1.c What about keeping a class somewhat similar to how Aura.Di is doing? http://auraphp.com/blog/2014/04/07/two-stage-config/ . This will reduce the code people need to copy and paste.

Thoughts?

I know you do love to move somewhat similar but it is due to getting things harder to test and maintain.

Maks3w:
Also, develop sub components on external repo make us to feel the same sand as the "great unknown developer" which fight to create his own extension but with the limitations of not change nothing in the parent project.

codeliner:
:+1: for @Maks3w reasoning

:+1: for @harikt point 1, :-1: for point 2

We (at prooph) integrate third party packages exactly the way proposed here. 
[...]

PR #133, opened 2015-10-10
Use FastRoute by default

weierophinney:
Hi,

I don't know where this discussion (https://github.com/zendframework/zend-expressive/issues/131) will bring us, but if we keep one default router, I think it should be FastRoute.

I've tried FastRoute and Aura, and FastRoute is definitely easier to use. See the comparaison:

Aura:
```php
return [
    'routes' => [
        [
            'path' => '/users/:id',
            'options' => [
                'tokens' => [
                    'id' => '[0-9]+'
                ]
            ],
        ],
    ],
];
```

FastRoute:
```php
return [
    'routes' => [
        [
            'path' => '/users/{id:[0-9]+}',
        ],
    ],
];
```

Aura reminds me a bit the verbosity of ZF2. FastRoute allows very compact definitions.

I don't know though if FastRoute lacks some important features that Aura could have, though.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ All notable changes to this project will be documented in this file, in reverse 
 - [#132](https://github.com/zendframework/zend-expressive/pull/132) adds
   `Zend\Expressive\Router\ZendRouter`, replacing
   `Zend\Expressive\Router\Zf2Router`.
-- [#139](https://github.com/zendframework/zend-expressive/pull/134) adds:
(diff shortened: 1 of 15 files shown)


weierophinney:
Pinging @bakura10 — more than your original intent, but also addresses an underlying issue I've observed with route configuration. If you have a chance, please take a look! (Planning on merging shortly anyways, as I'd like to get a 0.4.0 release out sooner rather than later.)

PR #156, opened 2015-10-13
Added second argument to Container::has callas in order to make the ServiceManager to include abstract factories

acelaya:
Every time the `Container::has` method is internally called, the second argument is set to true, so that when using the Zend\ServiceManager, it includes abstract factories.
Without this, when using the ServiceManager v3, regular middleware, route middleware and any service that is internally fetched and should be created by abstract factories is instantiated instead of being pulled from the Container.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -201,10 +201,14 @@ class Application extends MiddlewarePipe
     {
         // Lazy-load middleware from the container when possible
         $container = $this->container;
-        if (null === $middleware && is_string($path) && $container && $container->has($path)) {
+        if (null === $middleware && is_string($path) && $container && $container->has($path, true)) {
             $middleware = $this->marshalLazyMiddlewareService($path, $container);
             $path       = '/';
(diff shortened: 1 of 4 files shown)


acelaya:
I have changed all the `Container::has` calls, not just those fetching middlewares, in case any other service (like the router or the final handler fetched in the `ApplicationFactory`) needs to be created by using an abstract factory.

weierophinney:
@acelaya I don't think we should do the `has()` checks in the factories, as those should all be considered "well-known names" (i.e., they should have dedicated factories). Middleware is a different story, as those are run-time resolutions, and very likely could benefit from a generalized construction pattern (such as an abstract factory).

One thing to note as well: Because of changes in the `FactoryInterface`, you can now write factories that can return different instances based on the requested name. As such, the relevance of abstract factories will likely decrease once we have the v3 release stable.

Can you please revert the changes to the various factories? Thanks in advance!

acelaya:
That makes sense.
I'll change it right away.

acelaya:
I think it's correct now

weierophinney:
@acelaya Perfect! Thanks!

PR #158, opened 2015-10-16
Updated documentation to cover skeleton application

weierophinney:
- Updates the README install instructions to prefer the skeleton app over manual installation.
- Adds a chapter to the docs on getting started with the skeleton application and installer. This chapter also covers some next steps: creating middleware, and using templates.
- Updated the title of the original quickstart to indicate it covers standalone usage.

### DIFF
--- a/doc/book/quick-start-skeleton.md
+++ b/doc/book/quick-start-skeleton.md
@@ -43,7 +43,7 @@ http://localhost:8080/ to see if your application responds correctly!
 ## Next Steps
 
 The skeleton makes the assumption that you will be writing your middleware as
-objects, and using configuration to map routes to middleware. It also provides a
+classes, and using configuration to map routes to middleware. It also provides a
 default structure for templates, if you choose to use them. Let's see how you
 can create first vanilla middleware, and then templated middleware.
 
@@ -62,9 +62,12 @@ Let's create a "Hello" action. Place the following in
 <?php
 namespace App\Action;
 
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+
 class HelloAction
 {
-    public function __invoke($request, $response, callable $next)
+    public function __invoke(ServerRequestInterface $request, ResponseInterface $response, callable $next)
     {
         $query  = $request->getQueryParams();
         $target = isset($query['target']) ? $query['target'] : 'World';
@@ -150,6 +153,8 @@ Replace your `src/Action/HelloAction.php` file with the following contents:
 <?php
 namespace App\Action;
 
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
 use Zend\Diactoros\Response\HtmlResponse;
 use Zend\Expressive\Template\TemplateRendererInterface;
 
@@ -162,7 +167,7 @@ class HelloAction
         $this->renderer = $renderer;
     }
 
-    public function __invoke($request, $response, callable $next)
+    public function __invoke(ServerRequestInterface $request, ResponseInterface $response, callable $next)
     {
         $query  = $request->getQueryParams();
         $target = isset($query['target']) ? $query['target'] : 'World';
--- a/doc/book/quick-start-skeleton.md
+++ b/doc/book/quick-start-skeleton.md
@@ -1,0 +1,259 @@
+# Quick Start: Using the Skeleton + Installer
+
+The easiest way to get started with Expressive is to use the [skeleton
+application and installer](https://github.com/zendframework/zend-expressive-skeleton).
+The skeleton provides a generic structure for creating your applications, and
+prompts you to choose a router, dependency injection container, template
+renderer, and error handler from the outset.
+
+## 1. Create a new project
+
+First, we'll create a new project, using Composer's `create-project` command:
+
+```bash
+$ composer create-project zendframework/zend-expressive-skeleton expressive
+```
+
+This will prompt you to choose:
+
+- A router. We recommend using the default, FastRoute.
+- A dependency injection container. We recommend using the default, Zend
+  ServiceManager.
+- A template renderer. You can ignore this when creating an API project, but if
+  you will be creating any HTML pages, we recommend installing one. We prefer
+  Plates.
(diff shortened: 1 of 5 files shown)


doc/book/quick-start-skeleton.md
```diff
@@ -0,0 +1,259 @@
...
+$ php -S 0.0.0.0:8080 -t public/
+```
+
+This starts up a web server on localhost port 8080; browse to
+http://localhost:8080/ to see if your application responds correctly!
+
+## Next Steps
+
+The skeleton makes the assumption that you will be writing your middleware as
+objects, and using configuration to map routes to middleware. It also provides a
```
danizord: /s/objects/classes

doc/book/quick-start-skeleton.md
```diff
@@ -0,0 +1,259 @@
...
+Let's create a "Hello" action. Place the following in
+`src/Action/HelloAction.php`:
+
+```php
+<?php
+namespace App\Action;
+
+class HelloAction
+{
+    public function __invoke($request, $response, callable $next)
```
danizord: As it's an example everyone will follow as good practice, I think we should typehint PSR interfaces here.

doc/book/quick-start-skeleton.md
```diff
@@ -0,0 +1,259 @@
...
+    {
+        $this->renderer = $renderer;
+    }
+
+    public function __invoke($request, $response, callable $next)
+    {
+        $query  = $request->getQueryParams();
+        $target = isset($query['target']) ? $query['target'] : 'World';
+
+        return new HtmlResponse(
```
danizord: I know creating a new `HtmlResponse` is very convenient, but I'm not sure if this is a good thing. Also, in previous example we was using the $response provided via argument.

doc/book/quick-start-skeleton.md
```diff
@@ -0,0 +1,259 @@
...
+- We no longer need to escape our target; the template takes care of that for us.
+- We're using a specific response type here, from
+  [Diactoros](https://github.com/zendframework/zend-diactoros), which is the
+  default PSR-7 implementation Expressive uses. This response type simplifies
+  our response creation.
+
+How does the template renderer get into the action, however? The answer is
+dependency injection.
+
+Let's create a factory. Create the file `src/Action/HelloActionFactory.php` with
```
danizord: Missing note that we are assuming the developer picked `Zend\ServiceManager`

doc/book/quick-start-skeleton.md
```diff
@@ -0,0 +1,259 @@
...
+    {
+        return new HelloAction(
+            $container->get(TemplateRendererInterface::class)
+        );
+    }
+}
+```
+
+With that in place, we'll now update our configuration. Open the file
+`config/autoload/dependencies.global.php`; it should have a structure similar to
```
danizord: `routes.global.php`

doc/book/quick-start-skeleton.md
```diff
@@ -0,0 +1,259 @@
...
+the following:
+
+```php
+return [
+    'dependencies' => [
+        'invokables' => [
+            /* ... */
+        ],
+        'factories' => [
```
danizord: The `factories` key will not be there.

doc/book/quick-start-skeleton.md
weierophinney: I'm showing it as an example of how to create and return a custom response. Additionally, I'd argue injection of CORS headers might be something to intercept before emitting. We can worry about that later; this is a common pattern, and answers the questions of "why is populating a response so hard?!?!!?" which is why the custom response types exist.

doc/book/quick-start-skeleton.md
weierophinney: Nope. `dependencies.global.php` is correct here.

doc/book/quick-start-skeleton.md
weierophinney: It is; I actually tried the code from this example. ;-)

When you use Zend ServiceManager, the `dependencies.global.php` file is populated as shown above, and actually has a factory for each of the `HomePageAction` and `Zend\Expressive\Application`. (It also defines an invokable pointing to the `PingAction` class.)

weierophinney:
@danizord incorporated and/or answered all of your feedback items; thanks for the review!

doc/book/quick-start-skeleton.md
danizord: But the middleware factories should not be declared in `routes.global.php`?

doc/book/quick-start-skeleton.md
danizord: Why there is `dependencies` key in `routes.global.php` if we will config the middleware factories in `dependencies.global.php`?

doc/book/quick-start-skeleton.md
danizord: Makes sense :+1:

doc/book/quick-start-skeleton.md
danizord: Ah nvm, now I realized that the `dependencies` key in `routes.global.php` is meant to declare the router factory.

doc/book/quick-start-skeleton.md
danizord: Nvm, I saw the `dependencies` key in `routes.global.php` and messed the whole thing :P

doc/book/quick-start-skeleton.md
weierophinney: No. They should be on the dependencies.global.php file.

doc/book/quick-start-skeleton.md
weierophinney: I wasn't aware it was there until I started writing this; I'll be removing
it this weekend, as it's redundant, and not used.

weierophinney: Okay, figured it all out. The `dependencies` key in the `routes.global.php` gets merged with the `dependencies` key in `dependencies.global.php`. It exists essentially just to keep the information together; you can tell at a glance in that file which router is used, and what factory is used to create it. While that information could be in `dependencies.global.php`, this is a nice solution as it keeps the information in one place.

PR #175, opened 2015-10-21
stringify UTF-8 to fix PHP warning

MECU:
Without the quotes, produces warning: 
Warning: htmlspecialchars(): charset `-8' not supported, assuming utf-8 in ... on line ...

And then fatal error:
Fatal error: Uncaught exception 'RuntimeException' with message 'Unable to emit response; headers already sent' ...


Ocramius:
Thanks! Merged also into `develop` at zendframework/zend-expressive@ac15b63ff2a12a0e29382bab831897bff922b17a

PR #189, opened 2015-11-17
Prototype for allowing immutable RouterInterface

DASPRiD:
This PR is aimed to make the `RouterInterface` completely immutable. The way to achieve this is by injecting a `RouterFactoryInterface` into the Application, instead of a `RouterInterface`. The factory allows the same `addRoute()` method as the prior mutable router did. When requested, a router is build using the supplied routes.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -64,6 +64,11 @@ class Application extends MiddlewarePipe
     private $routeMiddlewareIsRegistered = false;
 
     /**
+     * @var Router\RouterFactoryInterface
+     */
+    private $routerFactory;
+
+    /**
      * @var Router\RouterInterface
      */
     private $router;
@@ -81,7 +86,7 @@ class Application extends MiddlewarePipe
      * Calls on the parent constructor, and then uses the provided arguments
      * to set internal properties.
      *
-     * @param Router\RouterInterface $router
+     * @param Router\RouterFactoryInterface $routerFactory
      * @param null|ContainerInterface $container IoC container from which to pull services, if any.
      * @param null|callable $finalHandler Final handler to use when $out is not
      *     provided on invocation.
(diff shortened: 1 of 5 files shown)


src/Application.php
```diff
@@ -383,6 +392,10 @@ public function routeMiddleware(ServerRequestInterface $request, ResponseInterfa
      */
     public function route($path, $middleware = null, array $methods = null, $name = null)
     {
+        if (null !== $this->router) {
+            // @todo should we throw an DomainException here, as routes cannot be added after the router was created?
```
DASPRiD: See here.

danizord:
@DASPRiD LGTM, but I think it could be named `RouterBuilderInterface` in order to avoid confusion with callable container factories and because it does creation step-by-step.

weierophinney:
I'm still wondering what *benefit* this provides. Since we still end up with one mutable object (the "builder"), what does this extra indirection provide tangibly to the system or the end user?

DASPRiD:
@weierophinney Once the router is created from the builder, it is save to assume that it is immutable then. The app doesn't call the builder twice.

weierophinney:
@DASPRiD My point is: the liklihood of adding routes after bootstrapping is ridiculously low, and we can instead specify in the current `RouterInterface` that `addRoute()` **MUST** raise an exception once either `match()` or `generateUri()` have been called. That's the only real benefit I'm seeing to your changes, but I'm not sure that change outweighs the complexity, particularly when there could be a simpler approach.

DASPRiD:
@weierophinney Well, if we kinda allow the router to freeze, that'd probably be fine as well I suppose.

weierophinney:
@DASPRiD We already have verbiage stating that `addRoute()` should aggregate only, but not pass the route information to the underlying until one of those two methods are called. I think we can and should stipulate that it should raise an exception if called *after* one of those is called as well, which gives the benefits you're proposing in here.

Do you want to do a PR for that? Or alter this one? or leave it to me?

weierophinney:
Closing in favor of #202.

PR #191, opened 2015-11-19
added coveralls badge

samsonasik:


Maks3w:
I don't find useful this kind of status badges. Status are showed where really means (PRs) so this start to be only noise.

weierophinney:
I tend to agree with @Maks3w here. The main thing that's useful for somebody coming into the project is that the stable branches have passing tests. Coverage is useful for those submitting patches, to see if coverage improves or lessens, so they can get an idea of completeness.

PR #192, opened 2015-11-20
Added support for route-specific middleware pipe via configuration

### LINKED ISSUE #190: Support route-specific middleware pipeline via configuration?
When defining routes, we can specify the middleware to be executed in the configuration:

```php
return [
    'routes' => [
        [
            'name' => 'foo',
            'path' => '/foo/{id}',
            'middleware' => FooMiddleware::class,
            'allowed_methods' => ['GET'],
        ]
    ]
];
```

I have a use case where I'd like to service a route by chaining a few different middlewares. They are application-specific middlewares that are shareable among resources (think user input validation, entity retrieval, and things of that nature), but I can only specify a single middleware in this configuration. I considered using the `pre_routing` and `post_routing` configuration pipelines, but they execute too early or too late, respectively, and I only need them to execute for this specific route, not for every request.

I managed to implement what I wanted by having my `FooMiddleware` actually be a middleware pipe:

```php
class FooMiddlewareFactory
{
    {
        $middleware = new MiddlewarePipe();
        $middleware->pipe($container->get(SomeMiddleware::class));
        $middleware->pipe($container->get(SomeOtherMiddleware::class));
        $middleware->pipe($container->get(YetAnotherMiddleware::class));

        return $middleware;
    }
}
```

Since a `MiddlewarePipe` is itself middleware, Expressive will accept it as the target middleware for this route, and execute the middlewares in the proper sequence.

However, it'd be much more convenient to be able to create this pipeline in the configuration rather than needing to create a factory to do it programmatically:

```php
return [
    'routes' => [
        [
            'name' => 'foo',
            'path' => '/foo/{id}',
            'middleware' => [
                SomeMiddleware::class,
                SomeOtherMiddleware::class,
                YetAnotherMiddleware::class,
            ],
            'allowed_methods' => ['GET'],
        ]
    ]
];
```

This is currently not allowed because the value of `middleware` must be callable.

1. Is there a better way to do what I'm trying to do already via configuration?
2. Is the inability to create route-specific pipelines like above via configuration an oversight, or intentional?
3. If it's an oversight, would you accept a backwards-compatible patch to get this into 1.0.0?

michaelmoussa:
The ApplicationFactory will check to see if the specified middleware is
an array before injecting it into the Route. If it is, it will attempt
to turn the contents of the array (which must be callable or retrievable
from the DI container) into a MiddlewarePipe, which it will inject
instead.

This change should be backwards compatible unless an application's functionality currently depends on failing if an array is specified in `routes.[].middleware`.

If acceptable, this would resolve #190

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -343,7 +343,19 @@ class Application extends MiddlewarePipe
             return $middleware($request, $response, $next);
         }
 
-        if (! is_string($middleware)) {
+        if (is_array($middleware)) {
+            $middlewarePipe = new MiddlewarePipe();
+
+            foreach ($middleware as $middlewareItem) {
+                $middlewarePipe->pipe(
+                    is_callable($middlewareItem) ? $middlewareItem : $this->marshalMiddleware($middlewareItem)
+                );
+            }
+
+            return $middlewarePipe;
+        }
+
+        if (!is_string($middleware)) {
             throw new Exception\InvalidMiddlewareException(
                 'The middleware specified is not callable'
             );
@@ -377,7 +389,7 @@ class Application extends MiddlewarePipe
      * pipeline.
      *
      * @param string|Router\Route $path
-     * @param callable|string $middleware Middleware (or middleware service name) to associate with route.
+     * @param callable|string|array $middleware Middleware (or middleware service name) to associate with route.
      * @param null|array $methods HTTP method to accept; null indicates any.
      * @param null|string $name the name of the route
      * @return Router\Route
@@ -527,6 +539,36 @@ class Application extends MiddlewarePipe
         if (count($matches) > 0) {
             throw new Exception\DuplicateRouteException(
                 'Duplicate route detected; same name or path, and one or more HTTP methods intersect'
+            );
+        }
+    }
+
+    /**
+     * Attempts to retrieve middleware from the container, or instantiate it directly.
+     *
+     * @param string $middleware
+     *
+     * @return array
+     * @throws Exception\InvalidMiddlewareException If unable to obtain callable middleware
+     */
+    private function marshalMiddleware($middleware)
+    {
+        $callable = $this->marshalMiddlewareFromContainer($middleware);
+
+        if (is_callable($callable)) {
+            return $callable;
+        }
+
+        $callable = $this->marshalInvokableMiddleware($middleware);
+
+        if (is_callable($callable)) {
+            return $callable;
+        } else {
+            throw new Exception\InvalidMiddlewareException(
+                sprintf(
+                    'Unable to resolve middleware "%s" to a callable',
+                    $middleware
+                )
             );
         }
     }
--- a/src/Application.php
+++ b/src/Application.php
@@ -352,7 +352,7 @@ class Application extends MiddlewarePipe
                 );
             }
 
-            return $middlewarePipe;
+            return $middlewarePipe($request, $response, $next);
         }
 
         if (!is_string($middleware)) {
--- a/src/Application.php
+++ b/src/Application.php
@@ -563,14 +563,14 @@ class Application extends MiddlewarePipe
 
         if (is_callable($callable)) {
             return $callable;
-        } else {
-            throw new Exception\InvalidMiddlewareException(
-                sprintf(
-                    'Unable to resolve middleware "%s" to a callable',
-                    $middleware
-                )
-            );
         }
+
+        throw new Exception\InvalidMiddlewareException(
+            sprintf(
+                'Unable to resolve middleware "%s" to a callable',
+                $middleware
+            )
+        );
     }
 
     /**
--- a/src/Container/ApplicationFactory.php
+++ b/src/Container/ApplicationFactory.php
@@ -212,12 +212,17 @@ class ApplicationFactory
                 continue;
             }
 
-            $path       = isset($spec['path']) ? $spec['path'] : '/';
-            $middleware = $spec['middleware'];
-            $error      = array_key_exists('error', $spec) ? (bool) $spec['error'] : false;
-            $pipe       = $error ? 'pipeErrorHandler' : 'pipe';
+            $path  = isset($spec['path']) ? $spec['path'] : '/';
+            $error = array_key_exists('error', $spec) ? (bool) $spec['error'] : false;
+            $pipe  = $error ? 'pipeErrorHandler' : 'pipe';
 
-            $app->{$pipe}($path, $middleware);
+            if (is_array($spec['middleware'])) {
+                foreach ($spec['middleware'] as $middleware) {
+                    $app->{$pipe}($path, $middleware);
+                }
+            } else {
+                $app->{$pipe}($path, $spec['middleware']);
+            }
         }
     }
 
--- a/src/Container/ApplicationFactory.php
+++ b/src/Container/ApplicationFactory.php
@@ -17,6 +17,7 @@ use Zend\Expressive\Container\Exception\InvalidArgumentException as ContainerInv
 use Zend\Expressive\Router\FastRouteRouter;
 use Zend\Expressive\Router\Route;
 use Zend\Expressive\Router\RouterInterface;
+use Zend\Stratigility\MiddlewarePipe;
 
 /**
  * Factory to use with an IoC container in order to return an Application instance.
@@ -175,8 +176,20 @@ class ApplicationFactory
             } else {
                 $methods = Route::HTTP_METHOD_ANY;
             }
-            $name    = isset($spec['name']) ? $spec['name'] : null;
-            $route   = new Route($spec['path'], $spec['middleware'], $methods, $name);
+
+            $name = isset($spec['name']) ? $spec['name'] : null;
+
+            if (is_array($spec['middleware'])) {
+                $middlewarePipe = new MiddlewarePipe();
+
+                foreach ($spec['middleware'] as $middleware) {
+                    $middlewarePipe->pipe(is_callable($middleware) ? $middleware : $container->get($middleware));
+                }
+
+                $spec['middleware'] = $middlewarePipe;
+            }
+
+            $route = new Route($spec['path'], $spec['middleware'], $methods, $name);
 
             if (isset($spec['options'])) {
                 $options = $spec['options'];
--- a/src/Container/ApplicationFactory.php
+++ b/src/Container/ApplicationFactory.php
@@ -179,7 +179,7 @@ class ApplicationFactory
 
             $name = isset($spec['name']) ? $spec['name'] : null;
 
-            if (is_array($spec['middleware'])) {
+            if (!is_callable($spec['middleware']) && is_array($spec['middleware'])) {
                 $middlewarePipe = new MiddlewarePipe();
 
                 foreach ($spec['middleware'] as $middleware) {
--- a/src/Container/ApplicationFactory.php
+++ b/src/Container/ApplicationFactory.php
@@ -176,10 +176,8 @@ class ApplicationFactory
             } else {
                 $methods = Route::HTTP_METHOD_ANY;
             }
-
-            $name = isset($spec['name']) ? $spec['name'] : null;
-
-            $route = new Route($spec['path'], $spec['middleware'], $methods, $name);
+            $name    = isset($spec['name']) ? $spec['name'] : null;
+            $route   = new Route($spec['path'], $spec['middleware'], $methods, $name);
 
             if (isset($spec['options'])) {
                 $options = $spec['options'];
--- a/src/Container/ApplicationFactory.php
+++ b/src/Container/ApplicationFactory.php
@@ -179,16 +179,6 @@ class ApplicationFactory
 
             $name = isset($spec['name']) ? $spec['name'] : null;
 
-            if (!is_callable($spec['middleware']) && is_array($spec['middleware'])) {
-                $middlewarePipe = new MiddlewarePipe();
-
-                foreach ($spec['middleware'] as $middleware) {
-                    $middlewarePipe->pipe(is_callable($middleware) ? $middleware : $container->get($middleware));
-                }
-
-                $spec['middleware'] = $middlewarePipe;
-            }
-
             $route = new Route($spec['path'], $spec['middleware'], $methods, $name);
 
             if (isset($spec['options'])) {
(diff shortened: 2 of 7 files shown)


src/Container/ApplicationFactory.php
```diff
@@ -165,9 +166,20 @@ private function injectRoutes(Application $app, ContainerInterface $container)
             $methods = (isset($spec['allowed_methods']) && is_array($spec['allowed_methods']))
                 ? $spec['allowed_methods']
                 : null;
-            $name    = isset($spec['name']) ? $spec['name'] : null;
+            $name = isset($spec['name']) ? $spec['name'] : null;
             $methods = (null === $methods) ? Route::HTTP_METHOD_ANY : $methods;
-            $route   = new Route($spec['path'], $spec['middleware'], $methods, $name);
+
+            if (is_array($spec['middleware'])) {
```
danizord: That prevents people from specifiyng `[SomeClass::class, 'someMethod']` as callable middleware.

weierophinney: @michaelmoussa The point @danizord makes could be addressed by changing the condition to `! is_callable($spec['middleware']) && is_array($spec['middleware'])`

weierophinney:
@michaelmoussa build failures are due to CS checks.

src/Container/ApplicationFactory.php
michaelmoussa: Thanks. I'll make the change, and add a regression test for that condition.

src/Container/ApplicationFactory.php
```diff
@@ -165,9 +166,20 @@ private function injectRoutes(Application $app, ContainerInterface $container)
...
+
+            if (!is_callable($spec['middleware']) && is_array($spec['middleware'])) {
+                $middlewarePipe = new MiddlewarePipe();
+
+                foreach ($spec['middleware'] as $middleware) {
+                    $middlewarePipe->pipe(is_callable($middleware) ? $middleware : $container->get($middleware));
+                }
+
+                $spec['middleware'] = $middlewarePipe;
+            }
```
weierophinney: Reviewing this, I wonder if it might make sense to do a "container-aware middleware pipe" which would lazy-load the middleware only when calling it. Looking into the Stratigility codebase, this would require both an alternate `Next` as well as `Dispatch` implementation — both of which would be easier to accomplish if we make a few properties protected and/or allow injection. I'll try to look into this this week.

test/Container/ApplicationFactoryTest.php
```diff
@@ -110,7 +110,22 @@ public function testFactoryWillPullAllReplaceableDependenciesFromContainerWhenPr
...
-    public function testFactorySetsUpRoutesFromConfig()
+    public function callableMiddlewares()
+    {
+        return [
+           ['HelloWorld'],
+           [
+                function () {
+                }
+           ],
+           [[\ZendTest\Expressive\TestAsset\InvokableMiddleware::class, 'staticallyCallableMiddleware']],
```
weierophinney: Import this class into the test case to make this simpler.

test/Container/ApplicationFactoryTest.php
```diff
@@ -1071,4 +1086,85 @@ public function testCanSpecifyRouteOptionsViaConfiguration()
         $this->assertInstanceOf(Route::class, $route);
         $this->assertEquals($expected, $route->getOptions());
     }
+
+    public function testCanPipeRouteSpecificMiddlewareViaConfiguration()
+    {
+        $router = $this->prophesize('Zend\Expressive\Router\RouterInterface');
+        $emitter = $this->prophesize('Zend\Diactoros\Response\EmitterInterface');
```
weierophinney: Let's import both of these as well, so we can use the `::class` constant.

michaelmoussa:
@weierophinney I imported those few classes and updated my new test case, but the same could be done all throughout the test class. I didn't, though, because it seemed out of scope for this PR. Should I do so anyway?

test/Container/ApplicationFactoryTest.php
```diff
@@ -1071,4 +1090,85 @@ public function testCanSpecifyRouteOptionsViaConfiguration()
...
+                    'name' => 'home',
+                    'allowed_methods' => ['GET'],
+                    'options' => $expected
+                ],
+            ],
+        ];
+
+        $this->container
+            ->has(RouterInterface::class)
+            ->willReturn(true);
```
weierophinney: @michaelmoussa Rebase against latest master, and then use `$this->injectServiceInContainer($this->container, <service name>, <service/value/mock>)` to inject services in the container. (You'll see examples of this in the test class already). This will simplify your tests, and reduce the white noise, making the test setup easier to understand. If you need help, let me know.

michaelmoussa: All set. Nice helper! Really cleans up that test class.

src/Container/ApplicationFactory.php
weierophinney: @michaelmoussa In reviewing this for merge... I think the code above is in the wrong place! Everywhere else, we delay fetching from the container *unless we're dispatching the middleware*. As such, I think this needs to be moved to `routeMiddleware()`.

However, for that to happen, a few other changes are necessary:

- `RouteResult::getMatchedMiddleware()` (and its accompanying property, `$matchedMiddleware`, and the `$middleware` argument for `fromRouteMatch()`) would need to adjust the typehint to `string|array|callable`.
- `Route::getMiddleware()` (and its accompanying property `$middleware`, and the `$midleware` argument to the constructor) would need to adjust the typehint to `string|array|callable`.
- Similarly, we also need to update `Application::route()` (this needs updating anyways with your change!)

The above bit that creates a middleware pipe would then move into `routeMiddleware()`, between the `is_callable()` and `is_string()` checks. We could either inline it as you've done here, or create a method that returns the `MiddlewarePipe` instance, and call that.

If you go that route, you could likely move the bits from `is_string()` up to `return $callable()` into a method that could be used by both methods, for marshaling the individual middlewares.

So, now the question: do you want to tackle those changes, or do you want me to do a PR against your branch? This approach would mean we don't have to worry about changes to Stratigility (as I think it's fine to fetch all the middleware for a given `MiddlewarePipe` at once in this circumstance), and I think the feature has some cool benefits; I'm already thinking about how we could have the `ApplicationFactory` also do global routed-middleware, into which the matched middleware would be injected. :smile:

weierophinney:
@michaelmoussa You'll need to rebase again (sorry!!!), but also, please review this comment: https://github.com/zendframework/zend-expressive/pull/192#discussion_r45805830

Let me know if you want me to do a PR against your branch, or if you want to tackle the changes proposed.

weierophinney:
@michaelmoussa — Also, regarding this:

> I imported those few classes and updated my new test case, but the same could be done all throughout the test class. I didn't, though, because it seemed out of scope for this PR. Should I do so anyway?

Don't worry about it. Most of them are updated at this point, and we'll get the remaining ones in later PRs.

src/Container/ApplicationFactory.php
michaelmoussa: Makes sense. Why go through the effort of potentially expensive middleware retrievals from the container if you don't even know if those middlewares will be used in this request?

I _think_  3ecc821 takes care of this in the way you had in mind. If not, hopefully it's a good start. Let me know if I understood correctly and/or if this needs additional polish.

michaelmoussa: @weierophinney I noticed that the changes thus far would not (yet) allow this type of array configuration in the pre- and post-routing pipelines. Of course, one could just specify them separately, like this:

```php
return [
    'middleware_pipeline' => [
        'pre_routing' => [
            [
                'middleware' => 'Foo',
                'path' => '/foobar',
            ],
            [
                'middleware' => 'Bar',
                'path' => '/foobar',
            ],
            [
                'middleware' => 'Foobar',
                'path' => '/foobar',
            ],
        ],
    ],
];
```

but something like this would be much more concise and potentially easier to read:

```php
return [
    'middleware_pipeline' => [
        'pre_routing' => [
            [
                'middleware' => [
                    'Foo',
                    'Bar',
                    'FooBar'
                ],
                'path' => '/foobar',
            ],
        ],
    ],
];
```

Do you agree? Would you want to try to cram that into this PR as well or do it as a future PR hopefully still targeting 1.0.0?

weierophinney: Interestingly, this comment doesn't show on the website; I think because
its on code you changed in a later commit. Hopefully you'll see this
message!

Yes, I think we should add this into the current pull request, as it makes
the behavior and configuration syntax consistent.

On Tue, Nov 24, 2015 at 11:28 PM, Michael Moussa <notifications@github.com>
wrote:

> In src/Container/ApplicationFactory.php
> <https://github.com/zendframework/zend-expressive/pull/192#discussion_r45830835>
> :
>
> >              $methods = (null === $methods) ? Route::HTTP_METHOD_ANY : $methods;
> > -            $route   = new Route($spec['path'], $spec['middleware'], $methods, $name);
> > +
> > +            if (!is_callable($spec['middleware']) && is_array($spec['middleware'])) {
> > +                $middlewarePipe = new MiddlewarePipe();
> > +
> > +                foreach ($spec['middleware'] as $middleware) {
> > +                    $middlewarePipe->pipe(is_callable($middleware) ? $middleware : $container->get($middleware));
> > +                }
> > +
> > +                $spec['middleware'] = $middlewarePipe;
> > +            }
>
> @weierophinney <https://github.com/weierophinney> I noticed that the
> changes thus far would not (yet) allow this type of array configuration in
> the pre- and post-routing pipelines. Of course, one could just specify them
> separately, like this:
>
> return [    'middleware_pipeline' => [        'pre_routing' => [            [                'middleware' => 'Foo',                'path' => '/foobar',            ],            [                'middleware' => 'Bar',                'path' => '/foobar',            ],            [                'middleware' => 'Foobar',                'path' => '/foobar',            ],        ],    ],];
>
> but something like this would be much more concise and potentially easier
> to read:
>
> return [    'middleware_pipeline' => [        'pre_routing' => [            [                'middleware' => [                    'Foo',                    'Bar',                    'FooBar'                ],                'path' => '/foobar',            ],        ],    ],];
>
> Do you agree? Would you want to try to cram that into this PR as well or
> do it as a future PR hopefully still targeting 1.0.0?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/zendframework/zend-expressive/pull/192/files#r45830835>
> .
>



-- 
Matthew Weier O'Phinney
matthew@weierophinney.net
https://mwop.net/

weierophinney:
@michaelmoussa added this comment to a code line that was later changed, so I'm adding it to the discussion thread to ensure there's public awareness of the discussion:

> @weierophinney I noticed that the changes thus far would not (yet) allow this type of array configuration in the pre- and post-routing pipelines. Of course, one could just specify them separately, like this:
> 
> ```php
> return [
>     'middleware_pipeline' => [
>         'pre_routing' => [
>             [
>                 'middleware' => 'Foo',
>                 'path' => '/foobar',
>             ],
>             [
>                 'middleware' => 'Bar',
>                 'path' => '/foobar',
>             ],
>             [
>                 'middleware' => 'Foobar',
>                 'path' => '/foobar',
>             ],
>         ],
>     ],
> ];
> but something like this would be much more concise and potentially easier to read:
> 
> return [
>     'middleware_pipeline' => [
>         'pre_routing' => [
>             [
>                 'middleware' => [
>                     'Foo',
>                     'Bar',
>                     'FooBar'
>                 ],
>                 'path' => '/foobar',
>             ],
>         ],
>     ],
> ];
> ```
>
> Do you agree? Would you want to try to cram that into this PR as well or do it as a future PR hopefully still targeting 1.0.0?

My response was:

> Yes, I think we should add this into the current pull request, as it makes the behavior and configuration syntax consistent.

michaelmoussa:
@weierophinney _Shouldn't_ be too much extra work I hope so I'll try to tackle this soon. Does everything else look OK to you as-is?

weierophinney:
> Does everything else look OK to you as-is?

Absolutely; very pleased with this improvement/feature!

ojhaujjwal:
+1 on the new type of array configuration in the pre and post-routing pipelines

samsonasik:
@michaelmoussa documentation need to be updated too imo ;)

weierophinney:
I can take care of documentation once the code is ready.

michaelmoussa:
@weierophinney I think that should take care of it.

src/Application.php
```diff
@@ -343,7 +343,19 @@ public function routeMiddleware(ServerRequestInterface $request, ResponseInterfa
...
+        if (is_array($middleware)) {
+            $middlewarePipe = new MiddlewarePipe();
+
+            foreach ($middleware as $middlewareItem) {
+                $middlewarePipe->pipe(
+                    is_callable($middlewareItem) ? $middlewareItem : $this->marshalMiddleware($middlewareItem)
+                );
+            }
+
+            return $middlewarePipe;
```
danizord:
```php
return $middlewarePipe($request, $response, $next);
```

src/Application.php
```diff
@@ -532,6 +544,36 @@ private function checkForDuplicateRoute($path, $methods = null)
...
+
+        if (is_callable($callable)) {
+            return $callable;
+        }
+
+        $callable = $this->marshalInvokableMiddleware($middleware);
+
+        if (is_callable($callable)) {
+            return $callable;
+        } else {
```
danizord: No need for `else` keyword here

src/Container/ApplicationFactory.php
```diff
@@ -175,8 +176,10 @@ private function injectRoutes(Application $app, ContainerInterface $container)
             } else {
                 $methods = Route::HTTP_METHOD_ANY;
             }
-            $name    = isset($spec['name']) ? $spec['name'] : null;
-            $route   = new Route($spec['path'], $spec['middleware'], $methods, $name);
+
+            $name = isset($spec['name']) ? $spec['name'] : null;
+
+            $route = new Route($spec['path'], $spec['middleware'], $methods, $name);
```
danizord: This change can be reverted?

michaelmoussa:
Thanks @danizord - I've made the changes you indicated.

weierophinney:
I added documentation as part of merging.

Additionally, @michaelmoussa, see be13b0b — that commit does a slight refactor of the `routeMiddleware()` method to re-use the `marshalMiddleware()` method you created when preparing a single middleware to dispatch.

PR #193, opened 2015-11-21
Support for setting PHP settings through configuration

nikolaposa:
Per-environment PHP configuration is very common use-case and so far it has been usually implemented by placing inline `ini_set` calls within some `php_settings.php` configuration itself. It would be much nicer if we can have clean configurations, for example:
```php
return [
    'php_settings' => [
        'display_startup_errors'        => false,
        'display_errors'                => false,
        'date.timezone'                 => 'Europe/Belgrade'
    ]
];
```
In order to achieve that, PHP settings must be set at some point, and I think that the most suitable context is the `ApplicationFactory`.

### DIFF
--- a/src/Container/ApplicationFactory.php
+++ b/src/Container/ApplicationFactory.php
@@ -146,7 +146,7 @@ class ApplicationFactory
 
     /**
      * Sets provided PHP settings, if any.
-     * 
+     *
      * @param ContainerInterface $container
      */
     private function setPhpSettings(ContainerInterface $container)
--- a/src/Container/ApplicationFactory.php
+++ b/src/Container/ApplicationFactory.php
@@ -136,11 +136,30 @@ class ApplicationFactory
 
         $app = new Application($router, $container, $finalHandler, $emitter);
 
+        $this->setPhpSettings($container);
         $this->injectPreMiddleware($app, $container);
         $this->injectRoutes($app, $container);
         $this->injectPostMiddleware($app, $container);
 
         return $app;
+    }
+
+    /**
+     * Sets provided PHP settings, if any.
+     * 
+     * @param ContainerInterface $container
+     */
+    private function setPhpSettings(ContainerInterface $container)
+    {
+        $config = $container->has('config') ? $container->get('config') : [];
+
+        if (!isset($config['php_settings']) || !is_array($config['php_settings'])) {
+            return;
+        }
+
+        foreach ($config['php_settings'] as $name => $value) {
+            ini_set($name, $value);
+        }
     }
 
     /**
--- a/src/Container/ApplicationFactory.php
+++ b/src/Container/ApplicationFactory.php
@@ -159,7 +159,7 @@ class ApplicationFactory
         if (!isset($config['php_settings'])) {
             return;
         }
-        
+
         if (!is_array($config['php_settings'])) {
             throw new Exception\InvalidArgumentException('Invalid PHP settings configuration; must be an array');
         }
--- a/src/Container/ApplicationFactory.php
+++ b/src/Container/ApplicationFactory.php
@@ -13,6 +13,7 @@ use Interop\Container\ContainerInterface;
 use Zend\Diactoros\Response\EmitterInterface;
 use Zend\Expressive\Application;
 use Zend\Expressive\Exception;
+use Zend\Expressive\Container\Exception\PhpSettingsFailureException;
 use Zend\Expressive\Router\FastRouteRouter;
 use Zend\Expressive\Router\Route;
 use Zend\Expressive\Router\RouterInterface;
@@ -148,17 +149,23 @@ class ApplicationFactory
      * Sets provided PHP settings, if any.
      *
      * @param ContainerInterface $container
+     * @throws Exception\InvalidArgumentException for invalid PHP settings.
+     * @throws PhpSettingsFailureException if setting of PHP configuration fails.
      */
     private function setPhpSettings(ContainerInterface $container)
     {
         $config = $container->has('config') ? $container->get('config') : [];
 
-        if (!isset($config['php_settings']) || !is_array($config['php_settings'])) {
-            return;
-        }
+        if (isset($config['php_settings'])) {
+            if (!is_array($config['php_settings'])) {
+                throw new Exception\InvalidArgumentException('Invalid PHP settings configuration; must be an array');
+            }
 
-        foreach ($config['php_settings'] as $name => $value) {
-            ini_set($name, $value);
+            foreach ($config['php_settings'] as $name => $value) {
+                if (false === ini_set($name, $value)) {
+                    throw PhpSettingsFailureException::forOption($name, $value);
+                }
+            }
         }
     }
 
--- a/src/Container/ApplicationFactory.php
+++ b/src/Container/ApplicationFactory.php
@@ -156,15 +156,17 @@ class ApplicationFactory
     {
         $config = $container->has('config') ? $container->get('config') : [];
 
-        if (isset($config['php_settings'])) {
-            if (!is_array($config['php_settings'])) {
-                throw new Exception\InvalidArgumentException('Invalid PHP settings configuration; must be an array');
-            }
+        if (!isset($config['php_settings'])) {
+            return;
+        }
+        
+        if (!is_array($config['php_settings'])) {
+            throw new Exception\InvalidArgumentException('Invalid PHP settings configuration; must be an array');
+        }
 
-            foreach ($config['php_settings'] as $name => $value) {
-                if (false === ini_set($name, $value)) {
-                    throw PhpSettingsFailureException::forOption($name, $value);
-                }
+        foreach ($config['php_settings'] as $name => $value) {
+            if (false === ini_set($name, $value)) {
+                throw PhpSettingsFailureException::forOption($name, $value);
             }
         }
     }
--- a/test/Container/ApplicationFactoryTest.php
+++ b/test/Container/ApplicationFactoryTest.php
@@ -65,6 +65,60 @@ class ApplicationFactoryTest extends TestCase
         return $r->getValue($app);
     }
 
+    public function testFactorySetsPhpSettingsFromConfig()
+    {
+        $this->container
+            ->has('Zend\Expressive\Router\RouterInterface')
+            ->willReturn(false);
+        $this->container
+            ->get('Zend\Expressive\Router\RouterInterface')
+            ->shouldNotBeCalled();
+
+        $this->container
+            ->has('Zend\Diactoros\Response\EmitterInterface')
+            ->willReturn(false);
+        $this->container
+            ->get('Zend\Diactoros\Response\EmitterInterface')
+            ->shouldNotBeCalled();
+
+        $this->container
+            ->has('Zend\Expressive\FinalHandler')
+            ->willReturn(false);
+        $this->container
+            ->get('Zend\Expressive\FinalHandler')
+            ->shouldNotBeCalled();
+
+        $config = [
+            'php_settings' => [
+                'display_errors' => 1,
+                'date.timezone' => 'Europe/Belgrade',
+            ],
+        ];
+
+        $currentPhpSettings = [];
+        foreach ($config['php_settings'] as $name => $value) {
+            $currentPhpSettings[$name] = ini_get($name);
+        }
+
+        $this->container
+            ->has('config')
+            ->willReturn(true);
+
+        $this->container
+            ->get('config')
+            ->willReturn($config);
+
+        $this->factory->__invoke($this->container->reveal());
+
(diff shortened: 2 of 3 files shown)


test/Container/ApplicationFactoryTest.php
```diff
@@ -65,6 +65,60 @@ public function getRouterFromApplication(Application $app)
...
+            ->willReturn(true);
+
+        $this->container
+            ->get('config')
+            ->willReturn($config);
+
+        $this->factory->__invoke($this->container->reveal());
+
+        foreach ($config['php_settings'] as $name => $value) {
+            $this->assertEquals($value, ini_get($name));
```
Maks3w: this foreach should be wrapped in a try/finally where try is this loop and finally is the tear down loop

test/Container/ApplicationFactoryTest.php
```diff
@@ -65,6 +65,60 @@ public function getRouterFromApplication(Application $app)
         return $r->getValue($app);
     }
 
+    public function testFactorySetsPhpSettingsFromConfig()
+    {
+        $this->container
+            ->has('Zend\Expressive\Router\RouterInterface')
+            ->willReturn(false);
+        $this->container
```
Maks3w: This expectation is unrelated with the feature under test. Only adds noise.

test/Container/ApplicationFactoryTest.php
```diff
@@ -65,6 +65,60 @@ public function getRouterFromApplication(Application $app)
...
+            ->get('Zend\Expressive\Router\RouterInterface')
+            ->shouldNotBeCalled();
+
+        $this->container
+            ->has('Zend\Diactoros\Response\EmitterInterface')
+            ->willReturn(false);
+        $this->container
```
Maks3w: Remove this expectation.

test/Container/ApplicationFactoryTest.php
```diff
@@ -65,6 +65,60 @@ public function getRouterFromApplication(Application $app)
...
+            ->get('Zend\Diactoros\Response\EmitterInterface')
+            ->shouldNotBeCalled();
+
+        $this->container
+            ->has('Zend\Expressive\FinalHandler')
+            ->willReturn(false);
+        $this->container
```
Maks3w: Remove this expectation.

test/Container/ApplicationFactoryTest.php
```diff
@@ -65,6 +65,60 @@ public function getRouterFromApplication(Application $app)
         return $r->getValue($app);
     }
 
+    public function testFactorySetsPhpSettingsFromConfig()
+    {
+        $this->container
```
Maks3w: Is this really needed?

src/Container/ApplicationFactory.php
```diff
@@ -144,6 +145,24 @@ public function __invoke(ContainerInterface $container)
...
     /**
+     * Sets provided PHP settings, if any.
+     *
+     * @param ContainerInterface $container
+     */
+    private function setPhpSettings(ContainerInterface $container)
+    {
+        $config = $container->has('config') ? $container->get('config') : [];
+
+        if (!isset($config['php_settings']) || !is_array($config['php_settings'])) {
```
Maks3w: if the value is not an array I guess an exception should be raised instead silently ignore.

src/Container/ApplicationFactory.php
```diff
@@ -144,6 +145,24 @@ public function __invoke(ContainerInterface $container)
...
+            return;
+        }
+
+        foreach ($config['php_settings'] as $name => $value) {
+            ini_set($name, $value);
```
Maks3w: Should raise an exception if `ini_set` fails?

test/Container/ApplicationFactoryTest.php
```diff
@@ -65,6 +65,60 @@ public function getRouterFromApplication(Application $app)
...
+            ->shouldNotBeCalled();
+
+        $config = [
+            'php_settings' => [
+                'display_errors' => 1,
+                'date.timezone' => 'Europe/Belgrade',
+            ],
+        ];
+
+        $currentPhpSettings = [];
```
Maks3w: It's better use PHPUnit `$this->iniSet()` and set a different value than the expected so:

* You can verify the ini set really has changed
* You can  remove both loops (this one and the restore one) and simplify the code.

Maks3w:
TBH I think environment configuration is something must be do in the configuration scripts (vagrant, docker, ...)

In special since there are many settings can't be changed at runtime.

src/Container/ApplicationFactory.php
nikolaposa: I kinda imitated `injectPreMiddleware()` flow in this case, but I guess it exception should be thrown.

src/Container/ApplicationFactory.php
nikolaposa: Ok, I agree, exception can be raised there.

test/Container/ApplicationFactoryTest.php
nikolaposa: I agree that it adds noise, but for some reason I couldn't bypass those opening lines within the `ApplicationFactory::__invoke()`. I'll revisit that anyhow.

test/Container/ApplicationFactoryTest.php
nikolaposa: Thanks for the tip

test/Container/ApplicationFactoryTest.php
nikolaposa: @Maks3w Any idea how to bypass this error when I remove all those expectations:
```
Prophecy\Exception\Call\UnexpectedCallException: Method call:
  - has("Zend\Expressive\Router\RouterInterface")
on Double\ContainerInterface\P65 was not expected,
```
?
It seems like stub yields error when unexpected method call is made.

Maks3w: This is related with the previous statement. https://github.com/zendframework/zend-expressive/pull/193/files#r45551040

Ok, it must be as you wrote.

src/Container/ApplicationFactory.php
Maks3w: /cc @weierophinney I think all those type checks should throw Exception instead of silently ignore.

test/Container/ApplicationFactoryTest.php
```diff
@@ -27,6 +27,68 @@ public function setUp()
         $this->factory   = new ApplicationFactory();
     }
 
+    protected function getContainer($config = null, $router = null, $finalHandler = null, $emitter = null)
```
nikolaposa: @Maks3w What do think about this approach, of having a helper method for configuring container based on the test specificity?

test/Container/ApplicationFactoryTest.php
```diff
@@ -65,6 +127,47 @@ public function getRouterFromApplication(Application $app)
...
+                'date.timezone' => 'Europe/Belgrade',
+            ],
+        ];
+
+        $this->factory->__invoke($this->getContainer($config)->reveal());
+
+        foreach ($config['php_settings'] as $name => $value) {
+            $this->assertEquals($value, ini_get($name));
+            ini_restore($name);
+        }
```
nikolaposa: @Maks3w Is this in line with what you mentioned about having only one loop?

test/Container/ApplicationFactoryTest.php
Maks3w: While I saw too a helper or many helper methods are needed I disagree with this proposal because add too much logic.

The problem with add logic branches in a test case (if, for, ...) is difficult to trust is bugfree. What test method test getContainer works as expected?

test/Container/ApplicationFactoryTest.php
Maks3w: PHPUnit assertions throw exceptions when assertion fail. So ini_restore won't be restored when test fail so next tests in the run will be executed with a wrong environment setting.

After $config

```php
   $this->iniSet('display_errors', 0);
   $this->iniSet('date.timezone', 'UTC');
```

src/Container/ApplicationFactory.php
```diff
@@ -144,6 +146,30 @@ public function __invoke(ContainerInterface $container)
...
+     *
+     * @param ContainerInterface $container
+     * @throws Exception\InvalidArgumentException for invalid PHP settings.
+     * @throws PhpSettingsFailureException if setting of PHP configuration fails.
+     */
+    private function setPhpSettings(ContainerInterface $container)
+    {
+        $config = $container->has('config') ? $container->get('config') : [];
+
+        if (isset($config['php_settings'])) {
```
Maks3w: We prefer the early return approach. if not isset then return;

test/Container/ApplicationFactoryTest.php
nikolaposa: But this would allow us to have only one place for managing and maintaining container configuration and it also solves code duplication in all the test cases in this class. I agree though that it is somewhat unreliable due to its complexity, but after all, every test case which consumes will fail if it does not work as expected.

src/Container/ApplicationFactory.php
```diff
@@ -136,6 +137,7 @@ public function __invoke(ContainerInterface $container)
 
         $app = new Application($router, $container, $finalHandler, $emitter);
 
+        $this->setPhpSettings($container);
```
nikolaposa: @Maks3w What do you think of moving `$this->setPhpSettings($container);` at the very begging of a `__invoke()` method, because it doesn't need `$app` at all?

xtreamwayz:
Sure it would look nice, but needed? In stead you can easily add a ./config/autoload/local.php file with a few ini_set calls in it.

Or maybe add some logic in the config loader that detects and sets the php_settings. This way the settings are set even before the application is started. In case of ``display_startup_errors`` and ``display_errors`` this is what you probably want anyway. I mean if you want to set ``display_errors`` to ``false`` and there is an error before it is actually set to ``false``, the error is displayed.

nikolaposa:
> file with a few ini_set calls in it.
The very intent of this PR was to avoid just that, I explained it in the introductory comment.

> This way the settings are set even before the application is started
I think that there is no point to track errors before this line: https://github.com/zendframework/zend-expressive-skeleton/blob/master/public/index.php#L18. Errors in pre-run stage can happen within the framework itself, unless of course someone put some massive, dynamic logic into configurations.

Maks3w:
The alternative is use a different front controller (index_dev.php) with this kind of customizations.

For me the problem is this feature does not work with any setting, only settings which allow runtime changes.

nikolaposa:
Ok, as for me, feel free to close this PR if you think that there will be no more comments or evaluations.

nikolaposa:
... btw maybe at least we can consider fixing those inconsistencies regarding throwing Exception vs silent ignoring (https://github.com/zendframework/zend-expressive/pull/193#discussion_r45555722) and eventually that helper method in ApplicationFactory test case which solves code duplication (https://github.com/zendframework/zend-expressive/pull/193/files#diff-80ae022ff5121fde2617d4a28d0b4351R30).

Maks3w:
I've open the issue https://github.com/zendframework/zend-expressive/issues/196 for track the inconsistencies. 

@nikolaposa Do you want to propose a PR for fix it?

nikolaposa:
@Maks3w Why not. I'll send PR soon.

Maks3w:
@nikolaposa  One important thing @weierophinney point me exceptions thrown must implement `Interop\Container\Exception\ContainerException`

We could reuse https://github.com/nikolaposa/zend-expressive/blob/feature/php-settings/src/Container/Exception/InvalidServiceException.php or create an specific one.

nikolaposa:
@Maks3w Duly noted.

weierophinney:
I honestly don't think this should be done as part of the application factory.

INI settings are an environmental concern, and should be done typically in one of the following places, in descending order of preference:

- system-wide INI file
- virtual host
- `.htaccess`
- bootstrap script

The rationale? You want the environment completely setup before you execute any of your application code, to ensure consistency of execution. As others have noted on this thread, a number of INI settings **CANNOT** be done in runtime regardless, and those that can will affect all code following.

Considering the number of deployment options available to PHP developers today, such changes can be easily accommodated as part of deployment:

- ZPK (Zend Server packaging format) allows defining INI settings.
- Chef, Puppet, and Ansible all let you control this as part of your application container configuration.

In a worst-case scenario, I'd create a script for setting any application-specific INI settings you rely on that can be set at runtime, and including that from the top of your `public/index.php` script; however, with the plethora of other options, that really should be a worst-case scenario.

Regarding the specifics that led to this request (inability to track where bad configuration was leading to invalid application state): I believe these have now been addressed in #194 and #195, and #197 will address the final lingering issues.

nikolaposa:
@weierophinney Thanks for this detailed explanation, now I'm convinced that this proposal has no practical value.

PR #195, opened 2015-11-23
[test] Add helpers for deal with ContainerInterface mock

Maks3w:
This PR is on top of #194

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -201,13 +201,13 @@ class Application extends MiddlewarePipe
     {
         // Lazy-load middleware from the container when possible
         $container = $this->container;
-        if (null === $middleware && is_string($path) && $container && $container->has($path, true)) {
+        if (null === $middleware && is_string($path) && $container && $container->has($path)) {
             $middleware = $this->marshalLazyMiddlewareService($path, $container);
             $path       = '/';
         } elseif (is_string($middleware)
             && ! is_callable($middleware)
             && $container
-            && $container->has($middleware, true)
+            && $container->has($middleware)
         ) {
             $middleware = $this->marshalLazyMiddlewareService($middleware, $container);
         } elseif (null === $middleware && is_callable($path)) {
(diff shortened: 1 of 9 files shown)


test/ContainerTrait.php
```diff
@@ -0,0 +1,52 @@
...
+namespace ZendTest\Expressive;
+
+use Interop\Container\ContainerInterface;
+use Prophecy\Argument;
+use Prophecy\Prophecy\ObjectProphecy;
+
+/**
+ * Helper methods for mock Interop\Container\ContainerInterface.
+ */
+trait ContainerTrait
```
Maks3w: @nikolaposa ping

nikolaposa: @Maks3w :+1:

nikolaposa:
@Maks3w When will this be merged? :)

src/Application.php
```diff
@@ -201,13 +201,13 @@ public function pipe($path, $middleware = null)
     {
         // Lazy-load middleware from the container when possible
         $container = $this->container;
-        if (null === $middleware && is_string($path) && $container && $container->has($path, true)) {
+        if (null === $middleware && is_string($path) && $container && $container->has($path)) {
```
weierophinney: Noting for posterity: The boolean `true` argument existed because an earlier revision of zend-servicemanager for v3 defined it; when provided, it indicated whether or not to look in abstract factories when performing the check. We have since removed that flag in zend-servicemanager's develop branch as it violates the LSP, but also because it was leading to many subtle errors (it was far too easy to forget to add it, and have a false negative lookup).

Maks3w: This change has be do on #194. This PR is on top of that one for prevent merge conflicts.

test/ContainerTrait.php
```diff
@@ -0,0 +1,52 @@
...
+     *
+     * Adjust `has('service')` and `get('service')` returns.
+     *
+     * @param ObjectProphecy $container
+     * @param string $serviceName
+     * @param mixed $service
+     *
+     * @return void
+     */
+    protected function setServiceToContainer(ObjectProphecy $container, $serviceName, $service)
```
weierophinney: This method name does not make sense to this particular native English speaker. Can you please switch it to one of these possibilities?

- injectServiceInContainer
- setContainerService

Obviously, this will require changes in the test cases where the method is called.

Maks3w: I have renamed locally as `injectServiceInContainer`. I'll ammend and push when review ends.

src/Application.php
weierophinney: There's nothing to change, @Maks3w ; I made the comment in case somebody wonders what the purpose of that change was. The change is fine.

Maks3w: I know. :) Just the comment was better located on that PR :)

weierophinney:
@Maks3w Looks good; let me know when you've pushed the changes to the method name.

Maks3w:
@weierophinney Done. Waiting checks to complete. I've changed the method description too https://github.com/zendframework/zend-expressive/commit/0a82f3fb45c9534d6422ca2a45de5b9c4f92f998#diff-706505189b1c9065d474fd9df5b59f64R37

PR #197, opened 2015-11-23
Improved exceptions handling within the Container\ApplicationFactory

### LINKED ISSUE #196: Throw exception instead silently return for type checks
Split the following conditions and throw exception when type check is not valid.

https://github.com/zendframework/zend-expressive/blob/master/src/Container/ApplicationFactory.php#L165
https://github.com/zendframework/zend-expressive/blob/master/src/Container/ApplicationFactory.php#L172
https://github.com/zendframework/zend-expressive/blob/master/src/Container/ApplicationFactory.php#L217
https://github.com/zendframework/zend-expressive/blob/master/src/Container/ApplicationFactory.php#L217

Also refactor https://github.com/zendframework/zend-expressive/blob/master/src/Container/ApplicationFactory.php#L236 for return if container does not have config.

nikolaposa:
This PR addresses issue #196.

/cc @Maks3w 

Fix #196

### DIFF
--- a/src/Container/ApplicationFactory.php
+++ b/src/Container/ApplicationFactory.php
@@ -13,6 +13,7 @@ use Interop\Container\ContainerInterface;
 use Zend\Diactoros\Response\EmitterInterface;
 use Zend\Expressive\Application;
 use Zend\Expressive\Exception;
+use Zend\Expressive\Container\Exception\InvalidArgumentException as ContainerInvalidArgumentException;
 use Zend\Expressive\Router\FastRouteRouter;
 use Zend\Expressive\Router\Route;
 use Zend\Expressive\Router\RouterInterface;
@@ -152,6 +153,7 @@ class ApplicationFactory
     private function injectRoutes(Application $app, ContainerInterface $container)
     {
         $config = $container->has('config') ? $container->get('config') : [];
+
         if (! isset($config['routes'])) {
             $app->pipeRoutingMiddleware();
             return;
@@ -162,15 +164,30 @@ class ApplicationFactory
                 continue;
             }
 
-            $methods = (isset($spec['allowed_methods']) && is_array($spec['allowed_methods']))
-                ? $spec['allowed_methods']
-                : null;
+            if (isset($spec['allowed_methods'])) {
+                $methods = $spec['allowed_methods'];
+                if (! is_array($methods)) {
+                    throw new ContainerInvalidArgumentException(sprintf(
+                        'Allowed HTTP methods for a route must be in form of an array; received "%s"',
+                        gettype($methods)
+                    ));
+                }
+            } else {
+                $methods = Route::HTTP_METHOD_ANY;
+            }
             $name    = isset($spec['name']) ? $spec['name'] : null;
-            $methods = (null === $methods) ? Route::HTTP_METHOD_ANY : $methods;
             $route   = new Route($spec['path'], $spec['middleware'], $methods, $name);
 
-            if (isset($spec['options']) && is_array($spec['options'])) {
-                $route->setOptions($spec['options']);
+            if (isset($spec['options'])) {
+                $options = $spec['options'];
+                if (! is_array($options)) {
+                    throw new ContainerInvalidArgumentException(sprintf(
+                        'Route options must be an array; received "%s"',
+                        gettype($options)
+                    ));
+                }
+
+                $route->setOptions($options);
             }
 
             $app->route($route);
@@ -212,14 +229,26 @@ class ApplicationFactory
      */
     private function injectPreMiddleware(Application $app, ContainerInterface $container)
     {
-        $config = $container->has('config') ? $container->get('config') : [];
-        if (! isset($config['middleware_pipeline']['pre_routing']) ||
-            ! is_array($config['middleware_pipeline']['pre_routing'])
-        ) {
+        if (!$container->has('config')) {
             return;
         }
 
-        $this->injectMiddleware($config['middleware_pipeline']['pre_routing'], $app, $container);
+        $config = $container->get('config');
+
+        if (! isset($config['middleware_pipeline']['pre_routing'])) {
+            return;
+        }
+
+        $middlewareCollection = $config['middleware_pipeline']['pre_routing'];
+
+        if (! is_array($middlewareCollection)) {
+            throw new ContainerInvalidArgumentException(sprintf(
+                'Pre-routing middleware collection must be an array; received "%s"',
+                gettype($middlewareCollection)
+            ));
+        }
+
+        $this->injectMiddleware($middlewareCollection, $app, $container);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


Maks3w:
Oops check the CS rules

nikolaposa:
Have no idea why is that strange code formatting issue randomly happening, but after another file modification/commit, everything seems fine.

src/Container/ApplicationFactory.php
```diff
@@ -151,7 +152,12 @@ public function __invoke(ContainerInterface $container)
      */
     private function injectRoutes(Application $app, ContainerInterface $container)
     {
-        $config = $container->has('config') ? $container->get('config') : [];
```
Maks3w: This case is legitimate. Next isset provide a fallback. Revert.

src/Container/ApplicationFactory.php
```diff
@@ -162,15 +168,29 @@ private function injectRoutes(Application $app, ContainerInterface $container)
                 continue;
             }
 
-            $methods = (isset($spec['allowed_methods']) && is_array($spec['allowed_methods']))
-                ? $spec['allowed_methods']
-                : null;
+            $methods = Route::HTTP_METHOD_ANY;
```
Maks3w: I would move this as `else` of the next if

Maks3w: Good catch about the double default value assigned

src/Container/ApplicationFactory.php
nikolaposa: Oops, you're right, I'll revert it after you finish a review.

src/Container/ApplicationFactory.php
nikolaposa: I was in doubt between the two, I eventually decided to use no-else style. But ok, I'll put `else` there.

src/Container/ApplicationFactory.php
```diff
@@ -13,6 +13,7 @@
 use Zend\Diactoros\Response\EmitterInterface;
 use Zend\Expressive\Application;
 use Zend\Expressive\Exception;
+use Zend\Expressive\Container\Exception\InvalidArgumentException as CntInvalidArgumentException;
```
Maks3w: I prefer `ContainerInvalidArgumentException`. I didn't know what Cnt means at the begin.

nikolaposa: Agree, it can be confused with "count".

test/Container/ApplicationFactoryTest.php
```diff
@@ -1071,4 +1071,226 @@ public function testCanSpecifyRouteOptionsViaConfiguration()
         $this->assertInstanceOf(Route::class, $route);
         $this->assertEquals($expected, $route->getOptions());
     }
+
+    /**
+     * @expectedException \Zend\Expressive\Container\Exception\InvalidArgumentException
+     */
+    public function testExceptionIsRaisedInCaseOfInvalidRouteMethodsConfiguration()
+    {
+        $router = $this->prophesize('Zend\Expressive\Router\RouterInterface');
```
Maks3w: Oops. After #195 is merged this should be refactored.

You can rebase your changes against my fork.

test/Container/ApplicationFactoryTest.php
```diff
@@ -1071,4 +1071,226 @@ public function testCanSpecifyRouteOptionsViaConfiguration()
         $this->assertInstanceOf(Route::class, $route);
         $this->assertEquals($expected, $route->getOptions());
     }
+
+    /**
+     * @expectedException \Zend\Expressive\Container\Exception\InvalidArgumentException
```
Maks3w: We prefer the "old" `setExpectedException` because:

* We place it just above of the statement which will throw the exception so there is less risk of catch an exception thrown by any other unexpected site.

* Make easy for many tools track the uses of the exception when PHP 5.5 ::class is used too.

test/Container/ApplicationFactoryTest.php
nikolaposa: Yeah, I remember that.

src/Container/ApplicationFactory.php
Maks3w: It's a microoptimization, save one memory write.

test/Container/ApplicationFactoryTest.php
nikolaposa: Makes sense.

Maks3w:
I'll review this again after to be rebased on top of #195.

nikolaposa:
Agreed.

test/Container/ApplicationFactoryTest.php
Maks3w: Also fill the second argument, expect message.

nikolaposa: What's your preference regarding second argument? Whole message, partial segment?

nikolaposa: ... I'm thinking of for example: `'must be in form of an array; received "string"'`.

nikolaposa: ... or simply `'string'`, in order to test-cover sprintf placeholder for type.

Maks3w: My rule is:

1. It must be unique for don't be confused with any other exception of the same type
2. Make the less efort (usually copy the whole message), if too much then only the essential part for acomplish the first rule.

Maks3w: > 'must be in form of an array; received "string"'

It's too generic and that substring can be reused for a different exception. (Even is recommended reuse same style of phrases for to reduce the mental effort)

`Allowed HTTP methods for a route must be in form of an array`

It's a better candidate. There is no risk of see any exception reusing the same string :)

Following my rule 2 of the less effort, just use the exact phrase because is enough short `Allowed HTTP methods for a route must be in form of an array; received "string"`

nikolaposa: Ok, I'll deal these exception messages more detailed after I sync with #195. Thanks for the tip.

test/Container/ApplicationFactoryTest.php
```diff
@@ -1071,4 +1071,230 @@ public function testCanSpecifyRouteOptionsViaConfiguration()
...
+        ];
+
+        $this->container
+            ->has('Zend\Expressive\Router\RouterInterface')
+            ->willReturn(true);
+        $this->container
+            ->get('Zend\Expressive\Router\RouterInterface')
+            ->will(function () use ($router) {
+                return $router->reveal();
+            });
```
weierophinney: @nikolaposa Can you please update these to use `ZendTest\Expressive\ContainerTrait`, now that #194/#195 have been merged? Doing so will cut down on a lot of noise, making it easier to see what's being tested.

Thanks!

nikolaposa: @weierophinney Sure, I've already discussed with @Maks3w how we should handle this.

test/Container/ApplicationFactoryTest.php
```diff
@@ -581,4 +581,80 @@ public function testCanSpecifyRouteOptionsViaConfiguration()
...
+                    'middleware' => 'HelloWorld',
+                    'allowed_methods' => 'invalid',
+                ],
+            ],
+        ];
+
+        $this->injectServiceInContainer($this->container, 'config', $config);
+
+        $this->setExpectedException(
+            'Zend\Expressive\Container\Exception\InvalidArgumentException',
```
Maks3w: Import and use class constant

Maks3w:
Please apply the last comment, and squash the commits. 


Thanks for your contribution

nikolaposa:
Done. I guess this is it.

Maks3w:
@weierophinney This is good for me.

PR #200, opened 2015-11-30
Implement subject/observer for route result notification

weierophinney:
This patch implements a subject/observer for notifying subscribers of a `RouteResult`. The primary use case is for URI generators, which may want access to the calculated `RouteResult`, without needing to resort to injection of middleware.

`update()` was chosen as the observer method, for affinity with [SplObserver](http://php.net/SplObserver).


akrabat:
Looks good to me.

PR #204, opened 2015-12-02
Removed Router and Template subcomponents

weierophinney:
These are now shipped with the following packages:

- zendframework/zend-expressive-router
- zendframework/zend-expressive-template

This package now *depends* on those, and removes the code common between them.


akrabat:
👍

PR #219, opened 2015-12-08
GLOB_BRACE fails on Win 7 using Xampp stack

weierophinney:
Using code like

return ConfigFactory::fromFiles(
        glob('config/autoload/{global,local}.php', GLOB_BRACE)
);

as described in Expressive example for config/config.php does not work on a stack described in the header (see http://zend-expressive.readthedocs.org/en/stable/usage-examples/#hello-world-using-a-configuration-driven-container).

http://www.php.net/function.glob notes: The GLOB_BRACE flag is not available on some non GNU systems, like Solaris. On my system it produces "Catchable fatal error: Argument 2 passed to Zend\Stdlib\ArrayUtils::merge() must be of the type array, integer given..."

Is there a requirement to change examples, or change Zend Config Factory::fromFiles, or change Zend Stdlib ArrayUtils::merge?

### DIFF
--- a/doc/book/usage-examples.md
+++ b/doc/book/usage-examples.md
@@ -256,10 +256,10 @@ repository, and also provides a mechanism for slip-streaming in
 configuration based on our environment (you might use different settings in
 development than in production, after all!).
 
-First, install zend-config:
+First, install zend-config and zend-stdlib:
 
 ```bash
-$ composer require zendframework/zend-config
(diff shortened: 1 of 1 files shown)


weierophinney:
@mike-a — I've converted this issue to a pull request that now contains the updated documentation.

mike-a:
Thanks for fast response!

weierophinney:
Reopening, as I've not merged the changes yet! :smile:

PR #217, opened 2015-12-08
How to setup form view helpers

weierophinney:
I have setup my `Zend\Expressive` application and really like the lightweight approach. But currently I am struggling to set up form view helpers to output a form in a view template. One issue I tracked down is due to the fact, that `Zend\Form` is not fully refactored to use the latest `Zend\ServiceManager` changes. But I am not sure if that is all to fix this.

I found some posts. The first is about integrating with Aura.DI:
http://harikt.com/blog/2015/11/13/integrating-zend-form-in-zend-expressive-and-view/

And the second with `Zend\ServiceManager` (which I could not get to run due to the current incompatibilities mentioned above):
http://www.masterzendframework.com/zend-expressive-enable-form-view-helpers/

It would be nice if the usage of form view helpers could be supported to help beginners adapt `Zend\Expressive`. I don't care if this can be done with an additional package, or within the Cookbook. At least the problem should be addressed to help the beginners.

What needs to be done to use the good old form view helpers within my `Zend\Expressive` project?

### DIFF
--- a/doc/book/cookbook.md
+++ b/doc/book/cookbook.md
@@ -360,3 +360,231 @@ This must be done *after*:
 
 This is to ensure that the `NotFound` middleware executes *after* any routed
 middleware, as you only want it to execute if no routed middleware was selected.
+
+## How can I use zend-form view helpers?
+
+If you've selected zend-view as your preferred template renderer, you'll likely
+want to use the various view helpers available in other components, such as:
+
+- zend-form
+- zend-i18n
+- zend-navigation
+
+By default, only the view helpers directly available in zend-view are available;
+how can you add the others?
+
+If you installed Expressive via the skeleton, the service
+`Zend\View\HelperPluginManager` is registered for you, and represents the helper
+plugin manager injected into the `PhpRenderer` instance. As such, you only need
+to configure this. The question is: where?
+
+You have a two options:
+
+- Add a delegator factory to or extend the `HelperPluginManager` service to
+  inject the additional helper configuration; or
+- Add pre_routing pipeline middleware that composes the `HelperPluginManager`
+  and configures it.
+
+### Delegator factories/service extension
+
+[Delegator factories](http://framework.zend.com/manual/current/en/modules/zend.service-manager.delegator-factories.html)
+and [service extension](https://github.com/silexphp/Pimple/tree/1.1#modifying-services-after-creation)
+operate on the same principle: they intercept after the original factory was
+called, and then operate on the generated instance, either modifying or
+replacing it. We'll demonstrate this for zend-servicemanager and Pimple; at the
+time of writing, we're unaware of a mechanism for doing so in Aura.Di.
+
+#### zend-servicemanager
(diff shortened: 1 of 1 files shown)


weierophinney:
@RalfEggert — I've converted this issue into a pull request containing updated (cookbook) documentation on how to use the helper configuration files from components to configure the `HelperPluginManager`, as well as how to add and configure custom helpers.

The documentation is based on zendframework/zend-expressive-skeleton#34, which updates the base zend-view configuration to ensure the `HelperPluginManager` factory is registered by default, and also to add the `view_helpers` configuration key to the templates configuration.

harikt:
If you have any doubts using with Aura.Di, I am happy to help.

Thank you.

weierophinney:
@harikt Is there a way to extend/decorate a service in Aura.Di? If so, can you provide a gist? I can then update this PR to incorporate that information.

harikt:
Hm. that is a tough question that I don't know or I am not getting the question. May be we should ping @pmjones .

weierophinney:
I've also added a section detailing a simpler method: replacing the `HelperPluginManager` factory with an application-specific one.

harikt:
One thing probably I can tell you,

If you have registered a service 

```php
$di->set('Zend\View\HelperPluginManager', $di->lazyNew('Zend\View\HelperPluginManager'));

// middleware instantiation

$di->params['FormHelpersMiddleware']['helpers'] = $di->lazyGet('Zend\View\HelperPluginManager');
```

In your middleware you can inject the `Zend\View\HelperPluginManager` service .

```php
use Zend\Form\View\HelperConfig;

class FormHelpersMiddleware
{
    private $helpers;

    public function __construct(HelperPluginManager $helpers)
    {
        $this->helpers = $helpers;
    }

    public function __invoke($request, $response, callable $next)
    {
        $config = new HelperConfig();
        $config->configureServiceManager($this->helpers);
        return $next($request, $response);
    }
}
```

Is this something you were looking for ?

harikt:
I noticed your commit little late ie why the first reply. Seems you already found some other way.

weierophinney:
@harikt I actually demonstrate the middleware-based solution already, and added another solution today that demonstrates simply replacing the `HelperPluginManager` factory. I think we're probably good at this point.

harikt:
yes, I copied the middleware code from your commit. The only thing I added was the aura/di config.

> I think we're probably good at this point.

great! .

weierophinney:
> The only thing I added was the aura/di config.

It's unneeded; the `config/container.php` we use for Aura.Di can use both the `invokables` and `factories` configuration used by the other containers, meaning no changes are required to make it work. :wink:

RalfEggert:
@weierophinney I will test it while working on the Album tutorial code and give some feedback here when finished...

RalfEggert:
@weierophinney tested the first solution with "Replacing the HelperPluginManager factory" and it worked like a charm 

:+1: nice

PR #220, opened 2015-12-09
Fix code in the example.

mtymek:


weierophinney:
Nice catch, @mtymek — thanks!

Issue #226, opened 2015-12-12
[Cookbook] How to setup localization


RalfEggert:
A suggestion for the cookbook: How to setup the localization?

Let's say in the routing is a `lang` parameter which defines the current language. This should be used to set the locale for the application. How can this be archieved? Are their multiple ways to handle this? Which way is the most recommended way?

weierophinney:
I presented a potential solution in #199

RalfEggert:
Ok, thanks for the link. I must have missed it. How could that solution be combined with the routing? Should it be placed in the `post_routing` pipeline? 

When I played around with a Middleware it only worked for `pre_routing` but wasn't called for `post_routing`. Here is my test middleware which is just setting the locale without an access to the routing for testing:

```
namespace Application\I18n;

use Locale;
use Psr\Http\Message\ResponseInterface;
use Psr\Http\Message\ServerRequestInterface;

class LocalizationMiddleware
{
    public function __invoke(
        ServerRequestInterface $request,
        ResponseInterface $response,
        callable $next
    ) {
        Locale::setDefault('de_DE');

        return $next($request, $response);
    }
}
```

When I add it here, it works:

```
return [
    'dependencies' => [
        'invokables' => [
            Application\I18n\LocalizationMiddleware::class =>
                Application\I18n\LocalizationMiddleware::class,
        ],
    ],

    'middleware_pipeline' => [
        'pre_routing' => [
            [
                'middleware' => [
                    Application\I18n\LocalizationMiddleware::class,
                ],
            ],
        ],

        'post_routing' => [
        ],
    ],
];
```

When I add it here, it doesn't:

```
return [
    'dependencies' => [
        'invokables' => [
            Application\I18n\LocalizationMiddleware::class =>
                Application\I18n\LocalizationMiddleware::class,
        ],
    ],

    'middleware_pipeline' => [
        'pre_routing' => [
        ],

        'post_routing' => [
            [
                'middleware' => [
                    Application\I18n\LocalizationMiddleware::class,
                ],
            ],
        ],
    ],
];
```

Is the configuration for a `post_routing` middleware different than for a `pre_routing` middleware?

harikt:
@RalfEggert are you calling the `$next()` from inside the action ? If not the `post_routing` will not work. Related https://github.com/zendframework/zend-expressive/issues/198

weierophinney:
post_routing middleware is piped to the application *after* the routing
middleware. This means it will only execute:

- if routing fails, or
[...]

PR #227, opened 2015-12-15
Added cookbook recipe for localization depending on routing

RalfEggert:
Added a recipe depending on the discussion and solutions in issue #226

### DIFF
--- a/doc/book/cookbook/setting-locale-depending-routing-parameter.md
+++ b/doc/book/cookbook/setting-locale-depending-routing-parameter.md
@@ -1,0 +1,303 @@
+# How can I setup the locale depending on a routing parameter?
+
+It is a common task to have an localized web application where the setting of
+the locale (and therefor the language) depends on the routing. There are other 
+ways of setting a language like using the session or a specialized sub-domain.
+In this recipe we will concentrate on using the a routing parameter.
+
+## Setting up the route ##
+
+If you want to set the locale depending on an routing parameter, you first have
+to add a language parameter to your routes. In this example we use the `lang` 
+parameter which should consist of two small letters,
+
+```php
+return [
+    'dependencies' => [
+        'invokables' => [
+            Zend\Expressive\Router\RouterInterface::class =>
+                Zend\Expressive\Router\ZendRouter::class,
+        ],
+        'factories' => [
+            Application\Action\HomePageAction::class =>
+                Application\Action\HomePageFactory::class,
+            Application\Action\ContactPageAction::class =>
+                Application\Action\ContactPageFactory::class,
+        ],
+    ],
+    'routes' => [
+        [
+            'name' => 'home',
+            'path' => '/:lang',
+            'middleware' => Application\Action\HomePageAction::class,
+            'allowed_methods' => ['GET'],
+            'options'         => [
+                'constraints' => [
+                    'lang' => '[a-z]{2}',
+                ],
+            ],
+        ],
+        [
+            'name' => 'contact',
+            'path' => '/:lang/contact',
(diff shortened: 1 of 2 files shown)


doc/book/cookbook/setting-locale-depending-routing-parameter.md
```diff
@@ -0,0 +1,303 @@
...
+
+It is a common task to have an localized web application where the setting of
+the locale (and therefor the language) depends on the routing. There are other 
+ways of setting a language like using the session or a specialized sub-domain.
+In this recipe we will concentrate on using the a routing parameter.
+
+## Setting up the route ##
+
+If you want to set the locale depending on an routing parameter, you first have
+to add a language parameter to your routes. In this example we use the `lang`
```
weierophinney: @RalfEggert — there's another way to do this that doesn't require touching each and every route: pre_routing pipeline middleware.

As an example, you would define such middleware like this:

```php
$uri = $request->getUri();
$path = $uri->getPath();
if (! preg_match('#^/(?P<lang>[a-z]{2})/#', $path, $matches) {
    Locale::setDefault('de_DE');
    return $next($request, $response);
}

$lang = $matches['lang'];
Locale::setDefault($lang);

$path = substr($path, 3);
$request = $request
    ->withUri($uri->withPath($path))
    ->withAttribute('lang', $lang);
return $next($request, $response);
```

This approach has a couple of benefits:

- No need for a route result observer.
- No need to prefix each and every route with `/:lang`.

The primary problem with it will be URI generation, as the router will be generating URIs without the language prefix, which will require a custom URI helper. However, this could even be something as simple as:

```php
namespace My\Application;

use Locale;
use Zend\Expressive\ZendView\UrlHelper;

class LocalizedUrlHelper extends UrlHelper
{
    public function __invoke($route = null, $params = [])
    {
        return sprintf('/%s%s', Locale::getDefault(), parent::__invoke($route, $params));
    }
}
```

I'm personally of the mind that we should demonstrate both approaches in the cookbook. Thoughts?

RalfEggert: But how would you handle redirects within a middleware? Currently I do this:

```
$routeParams = [
    'id'   => $id,
    'lang' => $request->getAttribute('lang'),
];

return new RedirectResponse(
    $this->router->generateUri('pizza-show', $routeParams)
);
```

That could be changed to 

```
$routeParams = [
    'id'   => $id,
    'lang' => Locale::getDefault(),
];

return new RedirectResponse(
    $this->router->generateUri('pizza-show', $routeParams)
);
```

But then again, in most projects I don't want to have `de_DE` or `en_US` in my URLs, I would preper `de` and `en`. So I could use `Locale::getPrimaryLanguage(Locale::getDefault())` to achieve that.

Sounds reasonable to me. Though I need to think about restructuring the cook book for both ideas. Personally I like the idea of the route result observer.

RalfEggert: @weierophinney 

I added a new PR #244 with an additional recipe. To have both variants in one recipe would bloat it. I hope this is ok.

weierophinney:
I was inclined to close this in favor of #244, but I think the write-up in here could be useful in creating portable middleware, as well as detailing how you might manage a mixture of routes both with *and without* the localization string (the solution in #244 basically assumes that *all* routes can and should work with the language prefix, which may not be the intention). As such, I'm going to merge both, but add some detail as to when you might use one over the other.

RalfEggert:
Yep, some cross links between the two recipes might be helpful. Thanks!

Issue #129, opened 2015-09-13
Don't recommend pimple-interop


akrabat:
`mouf/pimple-interop` is based on Pimple 1. The Pimple project is currently on version 3.0.2.  We probably don't want to be recommending Pimple 1.

weierophinney:
It doesn't look like v3 has container-interop support; is there a project that provides that?

(Also, completely missed that pimple-interop was using v1!!!!)

harikt:
Why not send a PR to make it compatible then ?

xtreamwayz:
First of all, I actually had a class in the [expressive installer](https://github.com/xtreamwayz/expressive-composer-installer/commit/5d6c8199cabab6ec7aea73915d7e6e1b1ce41efa) that made it work on a very basic level. I don't know the current status of the PSR, all I know is it needs a get and has function. This is easily done: https://xtreamwayz.github.io/blog/pimple-3-container-interop/. But then again, after that post I've seen stuff about a delegator look up and that wouldn't work with my concept. That's why I removed it a few days ago.

Secondly, after some research I found out they are not interested until a real PSR is out: https://github.com/silexphp/Pimple/issues/162.

Sam-Burns:
Hi guys, sorry to butt in if this isn't relevant, but would a project like https://github.com/Sam-Burns/pimple3-containerinterop be of any interest to you? \cc @mnapoli

weierophinney:
Yes, yes it would, @Sam-Burns .

I *think* it can even be a drop-in replacement for what we have in place already. Would you be willing to do a PR against zendframework/zend-expressive-skeleton to change the dependency used for Pimple? I can then do some integration testing to validate. If it looks good, I'll also update the docs in this repo to reflect the change in dependencies.

Sam-Burns:
@weierophinney Certainly willing to, yes.  Will make sure you're on \cc when the PR is ready.

weierophinney:
@Sam-Burns Thanks!

Sam-Burns:
Ref. https://github.com/zendframework/zend-expressive/pull/228

xtreamwayz:
Correct me if I'm wrong, but it looks like the pimple3-containerinterop wraps the container in a custom container class. This way it looses the flexibility of pimple, especially when building the container.

[...]

PR #228, opened 2015-12-15
Using a Pimple 3 wrapper as an example of a Container-Interop container

Sam-Burns:
Using samburns/pimple3-containerinterop as example of DI container

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,8 +72,8 @@ can recommend the following implementations:
 
 - [zend-servicemanager](https://github.com/zendframework/zend-servicemanager):
   `composer require zendframework/zend-servicemanager`
-- [pimple-interop](https://github.com/moufmouf/pimple-interop):
-  `composer require mouf/pimple-interop`
+- [pimple3-containerinterop](https://github.com/Sam-Burns/pimple3-containerinterop):
+  `composer require samburns/pimple3-containerinterop`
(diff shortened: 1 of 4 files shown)


Sam-Burns:
Ref. https://github.com/zendframework/zend-expressive/issues/129.  \cc @weierophinney

weierophinney:
@Sam-Burns Awesome! Can you also do a PR against https://github.com/zendframework/zend-expressive-skeleton ? Relevant files are:

- https://github.com/zendframework/zend-expressive-skeleton/blob/master/src/ExpressiveInstaller/config.php#L7
- https://github.com/zendframework/zend-expressive-skeleton/blob/master/src/ExpressiveInstaller/Resources/config/container-pimple-interop.php

As I noted on #129, I can then do some integration testing based on your patch.

Thanks!

xtreamwayz:
@weierophinney As explained in #129 I'm wondering if we should use another package that is extending Pimple in stead of placing it in another class. That way you won't loose all other functionality from Pimple and it's most likely easier to convert to a new Pimple release which supports container-interop.

weierophinney:
Closed in favor of #237. No hard feelings, @Sam-Burns — I just liked the approach of extending the pimple container better.

Sam-Burns:
@weierophinney it was an honour just to be nominated ;-)

PR #241, opened 2015-12-22
Use absolute urls to fix links in the documentation

xtreamwayz:
README.md is used as the index page for the documentation. However the links for docs and the notes are relative links. They should be absolute urls to work on https://zend-expressive.readthedocs.org/en/latest/.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ documentation.
 
 ## Documentation
 
-Documentation is [in the doc tree](doc/), and can be compiled using [bookdown](http://bookdown.io):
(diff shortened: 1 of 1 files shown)


mtymek:
Isn't it better to link directly to https://zend-expressive.readthedocs.org/en/latest/ instead of github.com?

xtreamwayz:
The files are not included in the documentation. They are referring to locations in the source.

weierophinney:
@xtreamwayz We won't *always* be using rtfd.org; it's a shim until we can do a website rewrite on framework.zend.com to add the Expressive documentation. As such, the README should have relative links; otherwise, they'll be incorrect for tags and branches.

PR #244, opened 2015-12-23
Added cookbook recipe for lang setting without additional route param

RalfEggert:
In regard to PR #227 and issue #226 I added another cookbook recipe to set up the language of an application without the need of using an observer or changing the routes. The recipes from this PR and PR #227 should probably cross-linked when merged.

### DIFF
--- a/doc/book/cookbook/setting-locale-without-routing-parameter.md
+++ b/doc/book/cookbook/setting-locale-without-routing-parameter.md
@@ -1,0 +1,162 @@
+# How can I setup the locale without an additional parameter?
+
+It is a common task to have an localized web application where the setting of
+the locale (and therefor the language) depends on the routing. In this recipe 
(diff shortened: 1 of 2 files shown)


doc/book/cookbook/setting-locale-without-routing-parameter.md
```diff
@@ -0,0 +1,162 @@
+# How can I setup the locale without an additional parameter?
+
+It is a common task to have an localized web application where the setting of
```
samsonasik: s/an localized/a localized

doc/book/cookbook/setting-locale-without-routing-parameter.md
```diff
@@ -0,0 +1,162 @@
...
+the locale (and therefor the language) depends on the routing. In this recipe
```
samsonasik: s/therefor/therefore

Issue #245, opened 2015-12-23
Use Foo::class instead of 'Foo'


yourwebmaker:
I just took a look into the project and it seems very promissor. :+1: 

https://github.com/zendframework/zend-expressive/blob/master/test/AppFactoryTest.php#L32

No more explanations. The issue title talks by itself. Can we do that?

Thanks guys.

bakura10:
Sure, that would be good, especially for refactoring a class name, at least with ::class those are automatically updated. Can you do a PR? :)

Maks3w:
I would prefer if this kind of checks could be automated and fixed by some tool.

weierophinney:
Closed with #246.

PR #246, opened 2015-12-24
Replace class names with ::class references

nightlinus:
PR for issue https://github.com/zendframework/zend-expressive/issues/245

### DIFF
--- a/test/AppFactoryTest.php
+++ b/test/AppFactoryTest.php
@@ -29,36 +29,36 @@ class AppFactoryTest extends TestCase
     public function testFactoryReturnsApplicationInstance()
     {
         $app = AppFactory::create();
-        $this->assertInstanceOf('Zend\Expressive\Application', $app);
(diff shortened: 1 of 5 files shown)


test/AppFactoryTest.php
```diff
@@ -29,36 +29,36 @@ public function getRouterFromApplication(Application $app)
     public function testFactoryReturnsApplicationInstance()
     {
         $app = AppFactory::create();
-        $this->assertInstanceOf('Zend\Expressive\Application', $app);
+        $this->assertInstanceOf(\Zend\Expressive\Application::class, $app);
```
samsonasik: `Application::class` is enough

nightlinus:
Do you mean to simlify all FQN's?

weierophinney:
Yes. It they're imported, there's no reason to fully qualify them.

weierophinney:
My only note is that any class you're referencing via `::class` notation should likely be imported. I can take care of that during merge, however.

Thanks, @nightlinus !

PR #254, opened 2016-01-01
Happy new year 2016

samsonasik:


bakura10:
You broke the tests :D.

samsonasik:
that should be fixed by change by @weierophinney at https://github.com/zendframework/zend-expressive/pull/255 at bookdown config

samsonasik:
I meant, mkdoc config

snapshotpl:
Every year this pr make me laugh. It doesn't start any functionality and real value. I propose to change this (and any other zf repo) to something else. e.g.:
```
Copyright (c) since 2015, Zend Technologies USA, Inc.
```

samsonasik:
then all 2005-2015 must be changed to 2005 in all components?

snapshotpl:
Yeah, why not? And so we need to change to current year :)

weierophinney:
@snapshotpl I did some research on this last month, and, unfortunately, the notation required by the US copyright office is `(c) <year>` or `(c) <first year>-<last year>`; `(c) since <year>` is not valid. While I admit to hating the annual patches, they're an unfortunate legal necessity.

Maks3w:
@weierophinney Is required to put a copyright holder in every file? Could be the LICENSE file enough?

weierophinney:
@Maks3w That's a hard question to answer.

Technically, the package itself is considered the redistributable work, which would mean that the LICENSE.md file needs to travel with the distribution, and would likely be sufficient. However, in cases where a developer might merge the source of the package with their own code and re-distribute, having the copyright and license in each class file ensures that their re-distribution complies with the license.

If you look historically at BSD license usage, the vast majority of projects have opted to include a reference to the copyright owner and a link to the license (if not the full license text itself) in each code file. My interpretation is that this ensures that any individual piece of code can be redistributed without requiring redistribution of the entire package in which it first appeared. I like this approach as it grants the greatest amount of freedom and least amount of effort on the part of users who want to redistribute.

As such, I feel we're stuck with these yearly updates, for better or for worse.

Maks3w:
PHP Community needs a tool for this :) Volunteers?

weierophinney:
We already have @samsonasik :)

Bigger problem is getting around to merging all of them!!!!

Maks3w:
I could sponsor that batch of merges if  @samsonasik is present in the IRC for coordination.

samsonasik:
I'm on irc now as samsonasik ;)

samsonasik:
@Maks3w i sign off now from irc as you didn't reply my greeting :p

Maks3w:
> (20:51:02) samsonasik [70d74245@gateway/web/freenode/ip.112.215.66.69] ha entrado en la sala.
(20:51:55) SLY1986 ha salido de la sala (quit: ).
(20:53:45) samsonasik: Maks3w, any something I can do?
(20:55:55) Maks3w: samsonasik: Ideally we should host a tool in https://github.com/zendframework/maintainers for automatize the change of year
(20:57:12) weierophinney: Maks3w, samsonasik : ++ to that idea!
(20:57:57) Maks3w: I think the copyright pattern is similar for any project @copyright Copyright (c)  start-end
(20:58:01) samsonasik ha salido de la sala (quit: Ping timeout: 252 seconds).

samsonasik:
ops, it seems irc timeout sorry...

samsonasik:
will see if I have a chance to create tool like that...

Warm regards,

Abdul Malik Ikhsan

Pada 5 Jan 2016, pukul 03.56, Maks3w <notifications@github.com> menulis:

PR #258, opened 2016-01-02
Final Handler factory

snapshotpl:


.gitignore
```diff
@@ -3,3 +3,4 @@ vendor/
 clover.xml
 composer.lock
 coveralls-upload.json
+/nbproject/private/
```
samsonasik: remove this line addition as you can setup globally for your specific ide

xtreamwayz:
I'm just wondering how I would use this. Some documentation would be helpful.

Looking at the stratigility docs: 

> Zend\Stratigility\FinalHandler is a default implementation of middleware to execute when the stack exhausts itself. It expects three arguments when invoked: a request instance, a response instance, and an error condition (or null for no error). It returns a response.

From the expressive docs:

> Error Handling
Expressive provides error handling out of the box, via zend-stratigility's FinalHandler implementation. This pseudo-middleware is executed in the following conditions:
If the middleware stack is exhausted, and no middleware has returned a response.
If an error has been passed via $next(), but not handled by any error middleware.

To me it tries to do exactly what expressive is already doing?

Oh... and it's 2016 these days :)

snapshotpl:
My plan it's to make a PR to skeleton with configured service in container and default configuration.

In other hand I don't like how I can setup display errors. I think that we shouldn't depend on env. But this is problem with stratigility.

xtreamwayz:
There are several other ways without depending on env:

You can also have a display_errors config setting, pass it to the template renderer and with some logic display the error or not in the error templates.

Or don't display the error traces in the default error templates and use the whoops error handler during development. This is how the skeleton is doing it.

weierophinney:
We chose deliberately *not* to provide a factory for `Zend\Stratigility\FinalHandler` in Expressive for a number of reasons:

- If you instantiate `Application` manually without a final handler (which is the default behavior when using `AppFactory::create()`, it will create a `Zend\Stratigility\FinalHandler` in production mode by default. This is reasonable default behavior.
- If you use `Zend\Expressive\Container\ApplicationFactory`:
  - If no `Zend\Expressive\FinalHandler` service is configured, it will, again, use Stratigility's `FinalHandler` in production mode.
  - Otherwise, it uses the configured service.
  - As such, you get something with reasonable default behavior without requiring any additional setup.
- We felt that the most common use cases for the final handler would be:
  - in production, to provide a *templated* page for detailing that an error/404 occurred; hence, the `TemplatedErrorHandler`.
  - in development, to provide a useful display with the exception stack trace; hence, the `WhoopsErrorHandler`.

These latter two provide a far better end-user experience in each case, which is why they were written, and why they are the defaults used by the skeleton as well.

As such, I'm going to close this feature request. If you want/need to use the Stratigility final handler implementation, you can write your own factory and/or use it as an invokable. However, the additional handlers provided by Expressive are far better suited for most needs, and, as such, will remain the only supported handlers for now (though we may provide an additional handler for APIs in the future).

Issue #242, opened 2015-12-22
Extend the conceptual interaction and usage of expressive based apps


ganeshkbhat:
Taking the first image (expressive components) in http://framework.zend.com/expressive I request if there can be an extension of this image. I am looking at a high level usage extension; basically that addresses the following:

* The flow of the request and response, and method of interaction of expressive components.
* How the request and response is handled with created middleware, routes, and pipes (http://zend-expressive.readthedocs.org/en/latest/application/ , http://zend-expressive.readthedocs.org/en/latest/router/piping/)

The docs are too crisp and I stumbled at some places. A little more documentation on routes, pipes, middleware would be nice. I saw a post by Ralf Eggert on addition of sample working codes. I recommend thatt. May be samples with Zend components.

weierophinney:
@ganeshkbhat Can you take a look at #261 (and/or the [rendered version of it](https://github.com/weierophinney/zend-expressive/blob/hotfix/diagram/doc/book/features.md#flow-overview) for me, and let me know if it helps towards this goal?

ganeshkbhat:
@weierophinney - `yes. this is it. `. thank you.

may be simple complete git project should help - may be i was and am expecting more. one has to hack in route match and middleware association here
http://zend-expressive.readthedocs.org/en/latest/router/uri-generation/ 
http://zend-expressive.readthedocs.org/en/latest/helpers/body-parse/

demo codes for below core of expressive would help:
`$app->pipe($middlewareToExecuteFirst);
$app->pipeRouteMiddleware();
$app->pipe($errorMiddleware);` in http://zend-expressive.readthedocs.org/en/latest/router/piping/ it was not clear else where.

weierophinney:
> may be simple complete git project should help

@RalfEggert is working on a full tutorial, which will likely aid in that regard.

ganeshkbhat:
@weierophinney  thank you very much for following up. just pre-during-post router and basic functionality with associating core mods to get used to the flavor.

PR #261, opened 2016-01-07
Added architectural diagram and narrative

weierophinney:
This patch includes an ASCII flow diagram, demonstrating typical workflows of middleware applications, as requested in #242. It also includes a narrative to describe the diagram, and thus the architecture of Expressive.


sandrokeil:
I like the diagram but with [js-sequence-diagrams](https://bramp.github.io/js-sequence-diagrams/) it's more readable. It's easy to use and programmatically.

Issue #266, opened 2016-01-11
Rename post_routing pipeline config param?


RalfEggert:
When I started working with `Zend\Expressive` I stumbled upon the `post_routing` parameter in the middleware pipeline configuration. Coming from `Zend\Mvc` I thought that these middleware will be executed directly after the routing. But through investigating it I noticed that these middleware are executed after the middleware for the current route. From that point of view the name sounds correct, but is leading to misunderstandings like here: https://github.com/zendframework/zend-expressive/issues/259

My suggestion would be to rename `post_routing` to `post_dispatching` to reduce future misinterpretations from new users.

weierophinney:
There's a number of confusions that I'm observing.

First "pipeline". Many are pointing out that a pipe is typically unidirectional. However, with middleware, we actually have two directions, as the middleware that first returns a response causes execution to return back out the path it traveled. This means that so-called "pre_routing" middleware might operate on the response returned by *routed* middleware, making it more like a "post" operation! As an example, consider the following middleware that adds a header to the returned response before returning it itself:

```php
function ($request, $response, $next)
{
    $response = $next($request, $response);
    return $response->withHeader('X-Clacks-Overhead', 'GNU Terry Pratchett');
}
```

The above would be registered as **pre_routing** middleware if you want it to execute for any response returned!

The next issue is then, of course, with the naming of `pre_routing` and `post_routing` as configuration keys. The "pre" and "post" here refer to *the order in which they are piped to the application*. This is not necessarily the *order in which operations are executed*, however, for the reasons outlined above. Additionally, if you consider error middleware, these are often registered towards the end of the pipeline; if a middleware calls `$next()` with an error argument, this will result in skipping over all other middleware until the first error middleware is reached, and any further calls to `$next()` then start from that offset.

The term "routing" was chosen for these configuration variables for affinity with the already established "routing middleware". Since we don't make any reference to "dispatching" in the code (other than `Zend\Stratigility\Dispatch`, which is an internal class), choosing that terminology seemed like it would require a fair amount of explanation, and might confuse readers.

So, the questions I have are:

- Do you have any better suggestions for the term "pipeline" that would disambiguate how the execution flow actually works?
- Knowing that "pre" and "post" refer to when the middleware is registered to the pipeline in relation to the *routing* middleware, do the terms require more disambiguation, or would documentation to that effect solve the issue?

RalfEggert:
@weierophinney Thanks for the explanations.

[...]

Issue #259, opened 2016-01-05
Separate routing from dispatching


danizord:
I have a middleware that needs a matched route param to perform some pre-dispatching logic, therefore this middleware should run **after** routing and **before** the "action" middleware.

Currently this can be achieved by injecting the router instance into a pre-routing middleware and calling `$router->match($request)` just to get the matched route params, but I think this is a very common use case and could be achieved more elegantly with a middleware registered between routing and dispatching.

It's possible to break the [Application#routeMiddleware()](https://github.com/zendframework/zend-expressive/blob/684e790e7584ba120ca3ce861330c85c5d72b539/src/Application.php#L357) into 2 middlewares?

RalfEggert:
Have you looked at the RouteResultObservers?

http://zend-expressive.readthedocs.org/en/latest/router/result-observers/

RalfEggert:
Here is a recipe that uses it:

http://zend-expressive.readthedocs.org/en/latest/cookbook/setting-locale-depending-routing-parameter/

danizord:
@RalfEggert `RouteResultObserver` will not work for me because I need access to the `ServerRequestInterface` instance. =\

Also, `RouteResultObserver` does not allow us to stop middleware pipeline by returning the `Response` instance without calling `$next`.

harikt:
I think it would be nice to have events built in the core. May be that will help ?

weierophinney:
@harikt — The only place events are needed are between routing and dispatching, which is something we've attempted to solve with the route result observers. I have no desire to add any events at this time, as you can typically solve the same problems by layering middleware.

@danizord — Are you aware that you can add a _list_ of middleware to dispatch when defining a route? As an example:

```php
[
    'name' => 'guarded',
    'route' => '/api/resource[/{id:[a-f0-9]{8}}]',
    'allowed_methods' => ['GET', 'POST'],
    'middleware' => [
        ValidateIdentifierMiddleware::class,
        ResourceMiddleware::class,
    ],
]
```

In the above example, if the route is matched, Expressive will create a `MiddlewarePipe` with the various middlewares listed. As a result, your `ValidateIdentifierMiddleware` could verify the `id` provided in the route result to determine if it's valid, use the server request, etc., and return a response if something is wrong; if all is well, it would call `$next()`, and delegation would pass on to the `ResourceMiddleware`.

If that won't work, perhaps you can explain the workflow you want to achieve in more detail?

nesl247:
[...]

PR #270, opened 2016-01-14
Simplify the middleware pipeline configuration

weierophinney:
This patch simplifies a number of features of Expressive:

- It simplifies the middleware pipeline configuration, by removing the `pre_routing` and `post_routing` keys; all middleware is piped at the same level of configuration, including routing.
- It splits the routing middleware into distinct routing and dispatch middleware. This allows replacing either process. This necessitates removal of autoregistration of the routing middleware; I consider this a *feature*, however, as it also ensures that you can form the middleware pipeline independent of when routes are added to the application.
- Because of the above, this patch also deprecates the route result observer system, as it's redundant. You can create "route result observers" as middleware that introspects the route result (which is injected as a request attribute following routing), and inject it in the pipeline immediately following the routing middleware. This simplifies route result observer registration as well, by providing exactly one way to accomplish it, and doing it in a portable fashion (middleware!).

I've noticed one potential issue with the change: configuration merging with indexed arrays means that we cannot guarantee order. This will potentially lead to issues if the `middleware_pipeline` is defined in multiple configuration files. To address this problem, this patch introduces a new middleware specification key, `priority`, which uses `SplPriorityQueue` semantics, and which can be used to guarantee the order in which middleware are piped to the application.

## TODO

- [x] deprecate pre/post routing keys
- [x] remove auto-registration of routing middleware
- [x] split routing middleware into routing + dispatch middleware
- [x] deprecate route result observers
- [x] update flow diagram
- [x] solve middleware_pipeline ordering issues post-merge (implement priority/order key)

## TODO post-merge:

- [ ] tag RC6 of zend-expressive.
- [x] Update zend-expressive-helpers to modify the `UrlHelper` to remove the `RouteResultObserverInterface` implementation, and update the `UrlHelperMiddleware` to push the route result from the request, if present, into the `UrlHelper`. Once complete, a new version should be tagged.
- [x] Update zend-expressive-router to mark the `RouteResult(Observer|Subject)Interface`s as deprecated.
- [ ] Update zend-expressive-skeleton:
  - [ ] update the default middleware pipeline to follow that demonstrated in this patch.
  - [ ] update the zend-expressive-helpers version to pull in the new tag.
  - [ ] update the zend-expressive version to rc6.


weierophinney:
Possible fix for #259 and #266; as noted; still a WIP.

danizord:
>the first time route() is called, the routing middleware is piped to the application. I'm considering removing that, and requiring developers explicitly do so

Yeah :+1:, that would bring us a lot of flexibility. Btw it would be great if we have to register a ` Zend\Expressive\Container\ApplicationFactory::ROUTED_MIDDLEWARE,` or `Zend\Expressive\Container\ApplicationFactory::DISPATCHING_MIDDLEWARE` as well, if we're going to split routing from dispatchig (#259).

>Additionally, I noticed a potential issue with the change: configuration merging with indexed arrays means that we cannot guarantee order. This will potentially lead to issues if the middleware_pipeline is defined in multiple configuration files.

I'd suggest us to document that 3rd-party modules should not register middlewares in the pipeline themselves, but let the developer (consumer) register them in proper order.

weierophinney:
> if we're going to split routing from dispatchig

I'm not planning on doing that. That's something you can do in custom routing solutions.

bakura10:
> I'd suggest us to document that 3rd-party modules should not register middlewares in the pipeline themselves, but let the developer (consumer) register them in proper order.

Yeah, I think this will cause some hard to debug issues when merging config. Maybe a possible solution would be to use a specific key that would say: anything coming after is pre_routing until next token. For instance:

```php
return [
   ApplicationFactory::PRE_ROUTING,
   // every middleware coming after is PRE_ROUTING until next token
   ApplicationFactory::ROUTING,
   // every middleware comign after is ROUTING until next token
   ApplicationFactory::POST_ROUTING
   // every middleware comign after is POST_ROUTING until next token
   ApplicationFactory::PRE_ROUTING
   // If an other module's config is merged, this oculd happen... but those will still be on Pre-routing
]
```

bakura10:
Hhmm nevermind, I think I misunderstood the purpose of this PR, it's actually to remove this distinction I suppose...

danizord:
>I'm not planning on doing that. That's something you can do in custom routing solutions.

Why not? This would simplify a lot things, completely remove the need of "route result observers" and bring a lot of flexibility to framework users.

weierophinney:
> I'd suggest us to document that 3rd-party modules should not register middlewares in the pipeline themselves, but let the developer (consumer) register them in proper order.

I both agree and disagree.

On the one hand, requiring that users manually register the middleware gives greater control in your applications.

On the flip side, it's not just 3rd party modules, but your own configuration. As an example, you may want middleware registered only locally, for debug purposes: debug toolbars, error handlers, etc. If you put these in a `*.local.php` file, they'll be merged, but not necessarily in an order that will work correctly (e.g., debug toolbars may need to be in the pipeline prior to routing, while error handlers need to appear after). As such, requiring manual registration *still doesn't solve the problem.*

The solutions we're looking at currently:

- For `config/autoload/` configurations, you can use the fact that globbing follows sort order, and have configurations that need to register middleware prior to the routing middleware occur higher in the sort list, and vice versa for error middleware.
- When using the expressive-config-manager, you can use the order of modules *to a degree*; however, this only works when a module defines middleware *only* for either pre or post routing purposes.
- Alternately, a "priority" flag in the middleware specification could be used by the application factory to sort them prior to attachment. This approach would address all strategies, but require more configuration by users. In most cases, it would be a single integer for each of pre, routing, and post, with more granularity *if needed*.

I'll get some commits in before long.

weierophinney:
> > I'm not planning on doing that. That's something you can do in custom routing solutions.
> Why not? This would simplify a lot things, completely remove the need of "route result observers" and bring a lot of flexibility to framework users.

Because the current implementation uses the results of routing to retrieve the middleware. If we separate them into distinct middleware, the changes required include:

- determining a way to pass the routing results to the dispatching middleware
- ensuring the dispatching middleware has access to the container
- ensuring the dispatching middleware has access to the same mechanisms for marshaling lazy middleware as the application

Not only does this result in complexity, it also requires documentation, and adds to the cognitive overhead of users trying to learn the system.

Keeping all of this internal to the `Application` instance is far simpler, and addresses the 80% use case; if you need the distinction, *this patch will allow you to do that yourself.*

danizord:
@weierophinney I'm really sorry for being very insistent :( , but I think this is actually easier for framework users and would be easy to implement as well (I can even do a PR if you want).

>* determining a way to pass the routing results to the dispatching middleware

```php
$request->getAttribute(RouteResult::class);
```
>* ensuring the dispatching middleware has access to the container
>* ensuring the dispatching middleware has access to the same mechanisms for marshaling lazy middleware as the application

Just inject needed dependencies
```php
$pipeline->pipe($cointainer->get(MiddlewareDispatcher::class));
```

>Not only does this result in complexity, it also requires documentation, and adds to the cognitive overhead of users trying to learn the system.

Actually I'd say that it makes the workflow even more easy to understand as it becomes explicit. Also this is much more clean that route result observers IMHO.

>Keeping all of this internal to the Application instance is far simpler, and addresses the 80% use case; if you need the distinction, this patch will allow you to do that yourself.

Yeah, actually it's **code** is simpler, but violates some SOLID principles and thus less maintainable, less replaceable, less flexible to changes...

Anyway sorry, this discussion is getting out of scope for this issue, let's go back to #259 :D

RalfEggert:
I really like this approach. It really reduces the possible misunderstandings. What I like the most is that it will come within a RC6 and not with 1.1.0. Great work!

weierophinney:
@danizord dcbcd09 accomplishes the split of the routing/dispatch responsibilities. I'm now working on the last patch for this PR, which will remove the route result observer system.

weierophinney:
oops! forgot to update the routing middleware test expectations. /me heads off to fix those

weierophinney:
@bakura10 We'll be doing an RC6 with these changes.

weierophinney:
`RouteMiddlewareTest` assumptions and setups updated to reflect the code change to split the routing and dispatch into separate middleware.

samsonasik:
I thinks features.md file need to be updated as well ( https://github.com/zendframework/zend-expressive/blob/master/doc/book/features.md )

samsonasik:
-> Flow Overview diagram at features.md

weierophinney:
@samsonasik Thanks for the heads-up; this is still a WIP, but I'd definitely missed that, and will include it in the patch.

weierophinney:
@danizord latest patch (3cbfd19) deprecates the route result observers, too. In updating the documentation, I discovered that around 75% of the docs for the `LocalizationObserver` example @RalfEggert wrote could be removed, as now the only possibility is to register middleware following routing. I call that a win!

weierophinney:
Created a checklist to detail the various pieces of this patch, and the progress.

danizord:
@weierophinney thanks for the awesome work!

>I discovered that around 75% of the docs for the LocalizationObserver example @RalfEggert wrote could be removed, as now the only possibility is to register middleware following routing. I call that a win!

Nice :)

doc/book/application.md
```diff
@@ -186,21 +186,24 @@ function ($error, ServerRequestInterface $request, ResponseInterface $response,
 
 Read the section on [piping vs routing](router/piping.md) for more information.
 
-### Registering routing middleware
+### Registering routing and dispatch middleware
 
-Routing is accomplished via dedicated a dedicated middleware method, 
-`Application::routeMiddleware()`. It is an instance method, and can be
-piped/registered with other middleware platforms if desired.
+Routing is accomplished via dedicated a dedicated middleware method,
```
danizord: `deticated` was wrote twice.

doc/book/application.md
```diff
@@ -186,21 +186,24 @@ function ($error, ServerRequestInterface $request, ResponseInterface $response,
...
+`Application::routeMiddleware()`; similarly, dispatching of routed middleware
+has a corresponding instance middleware method, `Application::dispatchMiddleware()`.
+Each can be piped/registered with other middleware platforms if desired.
```
danizord: Hmm, not sure how this can be possible given that they rely on `Application` state.

doc/book/container/factories.md
```diff
@@ -39,18 +39,15 @@ order to seed the `Application` instance:
...
-      // routing middleware:
-      'pre_routing' => [
-      ],
-      // An array of middleware to register after registration of the
-      // routing middleware:
-      'post_routing' => [
-      ],
+      // An array of middleware to register.
+      [ /* ... */ ],
+      Zend\Expressive\Container\ApplicationFactory::ROUTING_MIDDLEWARE,
```
danizord: MIssing `Zend\Expressive\Container\ApplicationFactory::DISPATCH_MIDDLEWARE`?

doc/book/container/factories.md
```diff
@@ -39,18 +39,15 @@ order to seed the `Application` instance:
...
+      [ /* ... */ ],
   ],
   ```
 
-  Each item of each array must be an array itself, with the following structure:
+  Each item of the array, other than the entry for routing middleware, must be
```
danizord: other than the entry for routing **and dispatch** middleware

doc/book/cookbook/custom-404-page-handling.md
```diff
@@ -86,22 +86,17 @@ From there, you still need to register the middleware. This middleware is not
 routed, and thus needs to be piped to the application instance. You can do this
 via either configuration, or manually.
 
-To do this via configuration, add an entry under the `post_routing` key of the
-`middleware_pipeline` configuration:
+To do this via configuration, add an entry under the `middleware_pipeline`
+configuration, after the routing middleware:
```
danizord: after the dispatch middleware

doc/book/cookbook/custom-404-page-handling.md
```diff
@@ -86,22 +86,17 @@ From there, you still need to register the middleware. This middleware is not
...
-            //...
-        ],
-
-    ],
-    'post_routing' => [
-        [
-            'middleware' => 'Application\NotFound',
-        ],
+    /* ... */
+    Zend\Expressive\Container\ApplicationFactory::ROUTING_MIDDLEWARE,
```
danizord: Missing `Zend\Expressive\Container\ApplicationFactory::DISPATCH_MIDDLEWARE` here

doc/book/application.md
weierophinney: If you define routes specific to the Application instance, you can pipe each of those individually to another stack. I'd likely just pipe the Application instance, however.

weierophinney:
@danizord Incorporated your feedback; thanks!

doc/book/container/factories.md
weierophinney: I changed this to only reference the `DISPATCH_MIDDLEWARE`.

doc/book/cookbook/debug-toolbars.md
```diff
@@ -74,11 +74,9 @@ return [
...
     'middleware_pipeline' => [
-        'pre_routing' => [
-            [
-                'middleware' => [
-                    PhpDebugBarMiddleware::class,
-                ],
+        [
+            'middleware' => [
+                PhpDebugBarMiddleware::class,
             ],
```
danizord: Should have routing and dispatch middlewares here to illustrate that `PhpDebugBarMiddleware::class` must be piped before them?

danizord: Ha I see this is`middleware-pipeline.local.php`

doc/book/cookbook/setting-locale-depending-routing-parameter.md
```diff
@@ -114,232 +112,44 @@ namespace Application\I18n;
...
 
-class LocalizationObserver implements RouteResultObserverInterface
+class LocalizationObserver
 {
-    public function update(RouteResult $result)
+    public function __invoke($request, $response, $next)
     {
-        if ($result->isFailure()) {
-            return;
+        $result = $request->getAttribute(RouteResult::class, false);
```
danizord: I think fetching `lang` from `$request->getAttribute('lang')` would be easier.

doc/book/cookbook/setting-locale-without-routing-parameter.md
```diff
@@ -102,19 +102,16 @@ return [
...
+                Application\I18n\SetLanguageMiddleware::class,
                 /* ... */
             ],
         ],
 
-        'post_routing' => [
-            /* ... */
-        ],
+        Zend\Expressive\Container\ApplicationFactory::ROUTING_MIDDLEWARE,
+
```
danizord: Missing dispatch middleware

doc/book/cookbook/using-a-base-path.md
```diff
@@ -106,10 +106,9 @@ return [
         /* ... */
     ],
     'middleware_pipeline' => [
-        'pre_routing' => [
-            [ 'middleware' => [ Blast\BaseUrl\BaseUrlMiddleware::class ] ],
-            /* ... */
-        ],
+        [ 'middleware' => [ Blast\BaseUrl\BaseUrlMiddleware::class ] ],
+        /* ... */
+        Zend\Expressive\Container\ApplicationFactory::ROUTING_MIDDLEWARE,
```
danizord: Missing dispatch middleware

doc/book/helpers/body-parse.md
```diff
@@ -43,13 +43,10 @@ return [
...
     ],
     'middleware_pipeline' => [
-        'pre_routing' => [
-            [ 'middleware' => Helper\BodyParams\BodyParamsMiddleware::class ],
-            /* ... */
-        ],
-        'post_routing' => [
-            /* ... */
-        ],
+        [ 'middleware' => Helper\BodyParams\BodyParamsMiddleware::class ],
```
danizord: Should probably run after routing and before dispatching

doc/book/helpers/server-url-helper.md
```diff
@@ -78,20 +79,23 @@ $container->set(
...
+To register the `ServerUrlMiddleware` as pipeline middleware following the
+routing middleware:
 
 ```php
 use Zend\Expressive\Helper\ServerUrlMiddleware;
 
-// Do this early, before piping other middleware or routes:
+// Programmatically:
+$app->pipeRoutingMiddleware();
 $app->pipe(ServerUrlMiddleware::class);
```
danizord: I think it should be piped **before** routing, so that if some error occurs during the routing phase, the error middlewares can still take advantage of it, for example a 404 error template.

doc/book/helpers/url-helper.md
```diff
@@ -112,20 +113,23 @@ return ['dependencies' => [
...
 // Or use configuration:
 // [
 //     'middleware_pipeline' => [
-//         'pre_routing' => [
-//             ['middleware' => UrlHelperMiddleware::class],
-//         ],
+//         /* ... */
+//         Zend\Expressive\Container\ApplicationFactory::ROUTING_MIDDLEWARE,
+//         ['middleware' => UrlHelperMiddleware::class],
+//         /* ... */
```
danizord: Missing dispatch middleware

doc/book/helpers/url-helper.md
```diff
@@ -112,20 +113,23 @@ return ['dependencies' => [
...
+To register the `UrlHelperMiddleware` as pipeline middleware following the
+routing middleware:
 
 ```php
 use Zend\Expressive\Helper\UrlHelperMiddleware;
 
-// Do this early, before piping other middleware or routes:
+// Programmatically:
+$app->pipeRoutingMiddleware();
 $app->pipe(UrlHelperMiddleware::class);
```
danizord: Pipe dispatch middleware here

doc/book/migration/rc-to-v1.md
```diff
@@ -0,0 +1,359 @@
...
+            //    'error' => true,
+            //],
+        ],
+    ],
+]
+```
+
+### Impact
+
+While the configuration from RC5 and earlier will continue to work, it will
```
danizord: This compatibility will be removed before stable 1.0?

doc/book/migration/rc-to-v1.md
```diff
@@ -0,0 +1,359 @@
...
+```php
+return [
+    'middleware_pipeline' => [
+        // An array of middleware to pipe to the application.
+        // Each item is of the following structure:
+        // [
+        //     // Required:
+        //     'middleware' => 'Name or array of names of middleware services and/or callables',
+        //     // Optional:
+        //     'path'  => '/path/to/match',
```
danizord: Do we allow specifiyng path for pipeline middlewares?

doc/book/migration/rc-to-v1.md
```diff
@@ -0,0 +1,359 @@
...
+        [
+            'middleware' => [
+                Helper\ServerUrlMiddleware::class,
+                Helper\UrlHelperMiddleware::class,
+            ],
+        ],
+
+        // The following is an entry for the routing middleware:
+        Zend\Expressive\Container\ApplicationFactory::ROUTING_MIDDLEWARE,
+
```
danizord: Could add an example here to illustrate that you can pipe a middleware between them?

doc/book/migration/rc-to-v1.md
```diff
@@ -0,0 +1,359 @@
...
+Once you have made the above changes, you should no longer receive deprecation
+notices when running your application.
+
+## Route result observer deprecation
+
+As of RC6, the following changes have occurred with regards to route result
+observers:
+
+- They are deprecated for usage with `Zend\Expressive\Application`, and that
+  class will not be a route result subject starting in 1.1. You will need to
```
danizord: >that class will not be a route result subject starting in 1.1

Is not it a BC break?

RalfEggert:
@weierophinney 

> I discovered that around 75% of the docs for the LocalizationObserver 
> example @RalfEggert wrote could be removed, as now the only possibility 
> is to register middleware following routing. I call that a win!

What? Deleting my docs is a win? WTF!

;-) Just kidding. 

I guess renaming the `LocalizationObserver` class to `LocalizationObserver` would be another idea for a little improvement. Otherwise people will wonder why its called observer.

Again, great stuff!

doc/book/cookbook/setting-locale-without-routing-parameter.md
```diff
@@ -102,19 +102,16 @@ return [
     ]
 
     'middleware_pipeline' => [
-        'pre_routing' => [
-            [
-                'middleware' => [
-                    Application\I18n\SetLanguageMiddleware::class,
```
ZergMark: To keep consistent, I would change all references to "lang" or "language" to "locale".

doc/book/cookbook/setting-locale-depending-routing-parameter.md
```diff
@@ -114,232 +112,44 @@ namespace Application\I18n;
...
+        if (! $result) {
+            return $next($request, $response);
         }
 
         $matchedParams = $result->getMatchedParams();
 
         $lang = isset($matchedParams['lang']) ? $matchedParams['lang'] : 'de_DE';
```
ZergMark: More "lang" to "locale" changes?

renanbr:
I see this issue walking for a good way, bur just to add another suggestion that affects user experience of middleware users and developers. I wonder that users could be able to add middleware to their applications without take care about position stuff, once the position is a part of implementation (in the cases I saw).

Fo example, if I want to add some `vendor/toolbar` or `vendor/template-engine` it would be wonderful do this in a simple way:

```php
return [
    'middlewares' => [
        Vendor\Toolbar::class,
        Vendor\TemplateEngine::class,
    ]
];
```

We should provide a simple way to install middlewares.

A possible solution for that:
- Middlewares that implements some interface could provide his correct position information, once it's related to  its behavior (programmatically). Something like `getMiddlewareDependencies(): string[]`, that returns list of middleware that must run __before__ this one.


Cons:
- Pipe engine will be more complex, especially for lazy services.
- More complex middleware development, even it doesn't affect the use of a middleware in other frameworks, usually the middlewares don't have specific implementations for some specific framework. People may see it as a "dirty-but-necessary" piece of code just for Expressive users...

Pros:
- Provide a easy way to install a middleware
- Avoid middleware installation mistakes

doc/book/cookbook/debug-toolbars.md
weierophinney: The latest patch also adds a `priority` key, and has the above example updated to specify priority 1000 (meaning prior to routing).

doc/book/cookbook/setting-locale-depending-routing-parameter.md
weierophinney: Agreed; rewriting this example now.

doc/book/helpers/server-url-helper.md
weierophinney: Yep; updated in latest patches.

doc/book/migration/rc-to-v1.md
weierophinney: No. If we did that, we'd need another RC between RC6 and stable that removes the functionality. Additionally, I know a number of folks are already pushing Expressive applications to production, so giving them until 1.1 ensures they have adequate time to upgrade.

doc/book/migration/rc-to-v1.md
weierophinney: Yes, but only *literal* paths. This is a feature from Stratigility and is used to allow segregating applications by path. You could, for instance, have an API application that triggers on the `/api` path, and which could be itself a fully configured Expressive instance. Routing done in such situations then becomes relative to the subpath specified, which allows you to re-use applications easily without modification.

doc/book/migration/rc-to-v1.md
weierophinney: Correct, but see my notes above. Technically, we're saying that while the feature will be available in the stable release, *we will not support it*, and have messaged when it will be removed. If we'd already had a stable release, we'd be issuing a new major version; because we haven't, we can mark the feature as deprecated and unsupported.

doc/book/cookbook/setting-locale-without-routing-parameter.md
weierophinney: Good point; making those changes now, and will push shortly.

weierophinney:
@renanbr One interesting thing that can be done due to changes a couple months ago is that you can specify an *array* of middleware inside a specification. As such, you can still have groupings for "pre" or "post" routing middleware by having a single middleware specification that has an *array* of middleware.

```php
// A pre-routing specification:
[
    'middleware' => [
        ServerUrlHelperMiddleware::class,
        BodyParamsMiddleware::class,
    ],
    'priority' => 1000,
]
```

I'm actually updating the documentation right now to recommend this as the migration path.

Doing it this way has several benefits:

- You can find the correct specification, and just add your middleware to the list in the position you want.
- It remains explicit to your application; no "magic" you need to understand to determine where it will execute.
- Merge order is preserved per-file, and if multiple files have specifications with the same priority, these will result in multiple entries piped to the application. As such, you can define priority per configuration file in order to shape the entire application pipeline.

There are several reasons we cannot use the configuration you presented above:

- We need to be able to distinguish between normal and error middleware for lazy services. `Zend\Stratigility\MiddlewarePipe` looks at the arity of the provided callable to determine what kind of middleware was provided — which we cannot do for lazy services, as we do not have the callable *yet*. Hence, configuration makes this explicit for lazy services.
- To allow path-based middleware segregation, we need to be able to provide the path as a key.
- To allow granular configuration by segregating by files, we have enabled configuration merging. Unfortunately, with lists, the default behavior of such merging is to append entries, which breaks the order. Hence, the `priority` key as proposed in this patch.
- Having the middleware specify its own position removes flexibility on the part of end-users, as they cannot then override that information to create their own workflow.

As such, I think the proposed changes in this patch help reduce some complexity (removal of the confusing `pre_routing`/`post_routing` terminology), while simultaneously providing far more granularity in how users can define the order in which middleware should execute.

weierophinney:
All code changes are now complete; will work on the diagram in the next couple days.

doc/book/migration/rc-to-v1.md
basz: Do I understand this correctly? Can expressive apps embed expressive apps? If so wow

weierophinney:
Added a todo list for post-merge to the issue description.

weierophinney:
@RalfEggert 8a9a4a3 introduces a `priority` key for middleware specifications in order to guarantee order in which they are attached to the pipeline, and address the configuration merging issues; can you review and provide feedback?

doc/book/migration/rc-to-v1.md
weierophinney: @basz Yes, indeed. :smile:

RalfEggert:
The addition of the `priority` key is a really good step in the direction of simplicity and flexibility. It is easy to understand and easy to use and I really like it. I think this makes `Zend\Expressive` even more useful than it already was before.

:+1: 

Really looking forward for the RC6 release.

weierophinney:
All changes are complete, including updated flow diagram. Ready for final review prior to merge!

weierophinney:
@danizord — What are your thoughts on this now? :smile:

doc/book/cookbook/setting-locale-depending-routing-parameter.md
```diff
@@ -99,247 +99,56 @@ return [
...
-            /* ... */
-        ],
-        'post_routing' => [
-            /* ... */
+        /* ... */
+        [
+            'middleware' => [
+                Zend\Expressive\Container\ApplicationFactory::ROUTING_MIDDLEWARE,
+                Helper\UrlHelperMiddleware::class,
+                [ 'middleware' => LocalizationMiddleware::class ],
```
harikt: actually I have a doubt. Is [] mean optional or we can write middleware aarays inside middleware ?

weierophinney: I need to fix that one; nice catch, @harikt !

ZergMark:
If I may make two comments/suggestions. The Architecture diagram is a nice addition, but I _think_ has extra dividing lines in the green and red sections. Right now it appears as if "high-priority" passes to "(run always/early)" passes to "pipeline" passes to "middleware" before hitting the routing middleware. As this is actually one step "high priority (run always/early) pipeline middleware" instead of four, I would recommend removing the horizontal lines and indicate somehow else (card stack?) that there can be multiple middleware called before routing.

Also, wasn't one reason behind the routing/dispatch split so that "listeners" could slide in between routing and dispatching? Perhaps an "(Optional) route-dependant helper middleware" layer could be added betwixt the two?

weierophinney:
@ZergMark Thanks for the feedback!

Yes, the reason for the routing/dispatch split is to allow placing middleware between the two. I wasn't sure how to illustrate that without complicating the diagram, however. If you have any suggestions, feel free to provide some sketches.

Regarding this:

> but I think has extra dividing lines in the green and red sections. Right now it appears as if "high-priority" passes to "(run always/early)" passes to "pipeline" passes to "middleware" before hitting the routing middleware. As this is actually one step "high priority (run always/early) pipeline middleware" instead of four, I would recommend removing the horizontal lines and indicate somehow else (card stack?) that there can be multiple middleware called before routing.

The idea here is that each box represents middleware. I colored the high-priority middleware as green and the error middleware as red simply to indicate a distinction with the routing/dispatch middleware, and to provide visual groupings so that the labels can be read as groups.

However, the point is that an application consists of layered middleware, with the routing/dispatch middleware typically running somewhere in the middle of the architecture.

Again, if you have ideas for further clarification, please share some sketches!

weierophinney:
@harikt —

> if this was wrong, I guess I noticed it in some other places

I've found audited the various places where `middleware` specifications are defined, and found one more such nested speficication, which I've fixed in 9a00c36. I think we're good now; thanks for the review!

harikt:
Thank you for your good work on this.

Now I need to see how I can upgrade without much issues :) .

weierophinney:
@harikt —

> Now I need to see how I can upgrade without much issues :) .

I wrote a migration guide as part of this PR. :)

weierophinney:
@harikt That said, I need to update zend-expressive-helpers after this is merged; I didn't want to remove the route result observer functionality until after this was accepted.

ZergMark:
@weierophinney I was thinking perhaps something along the below. It labels the areas correctly without the "first scan" problem of implying that each line is it's own type of middleware that comes before routing. Not a major issue of course.

architecture

weierophinney:
@ZergMark Awesome! I'll make that update shortly!

weierophinney:
@ZergMark Updated!

ZergMark:
@weierophinney Thanks. Looks great.

ezimuel:
@weierophinney it's ok for me, very nice feature!

src/Application.php
```diff
@@ -232,41 +251,47 @@ public function notifyRouteResultObservers(Router\RouteResult $result)
...
-            $middleware = $path;
-            $path       = '/';
+            $middleware = $this->prepareMiddleware($middleware, $this->container);
         }
 
         if ($middleware === [$this, 'routeMiddleware'] && $this->routeMiddlewareIsRegistered) {
             return $this;
         }
 
+        if ($middleware === [$this, 'dispatchMiddleware'] && $this->dispatchMiddlewareIsRegistered) {
```
mtymek: Why do we need a check if routing and dispatch middleware are registered? This adds extra logic for every call to `pipe()`.
If this check is needed, I'd rather throw an exception to show the user that something is not properly configured.

mtymek:
Wow, I love this changes! :+1: :+1: :+1: :+1:

src/Application.php
weierophinney: It's to prevent double-registration; most likely you would never want to double-route/double-dispatch, as that's expensive to do in the same request.

We can investigate making these exceptions instead, but that's out of scope for this patch; the behavior added for the dispatch middleware was done to mimic the existing behavior of registering the routing middleware.

doc/book/cookbook/using-zend-form-view-helpers.md
```diff
@@ -224,13 +224,17 @@ return [
         /* ... */
     ],
     'middleware_pipeline' => [
-        'pre_routing' => [
-            ['middleware' => Your\Application\FormHelpersMiddleware::class],
-            /* ... */
-        ],
-        'post_routing' => [
-            /* ... */
+        ['middleware' => Your\Application\FormHelpersMiddleware::class, 'priority' => 1000],
```
danizord: Maybe this recipe should be updated to suggest config merging instead?

doc/book/helpers/body-parse.md
```diff
@@ -43,13 +43,17 @@ return [
         ],
     ],
     'middleware_pipeline' => [
-        'pre_routing' => [
-            [ 'middleware' => Helper\BodyParams\BodyParamsMiddleware::class ],
-            /* ... */
-        ],
-        'post_routing' => [
-            /* ... */
+        [ 'middleware' => Helper\BodyParams\BodyParamsMiddleware::class, 'priority' => 100],
```
danizord: Can't we pipe this one after routing middleware? It should be a little performance optimization.

danizord: If yes, the examples of pre-routing middlewares in `features.md` should be updated as well.

doc/book/cookbook/using-zend-form-view-helpers.md
weierophinney: For standalone, priority is a good rule-of-thumb. Integrated, you'd want it likely inside a middleware pipeline definition.

doc/book/helpers/body-parse.md
weierophinney: There are a few places it could be pushed:

- prior to routing (demonstrated)
- between routing and dispatch (not demonstrated)
- in routing-specific middleware pipelines (demonstrated, and listed as the optimal path, as it ensures it's not executed unless your middleware is actually expecting body params)

We can add the middle one via another PR later; I merely updated the existing examples to the changes in this PR.

doc/book/cookbook/using-zend-form-view-helpers.md
danizord: I'm talking about merging config for `ViewHelperManager` instead of using middlewares to update it.

doc/book/helpers/server-url-helper.md
```diff
@@ -78,20 +79,22 @@ $container->set(
 );
 ```
 
-To register the `ServerUrlMiddleware` as pre-routing pipeline middleware:
+To register the `ServerUrlMiddleware` as pipeline middleware following the
```
danizord: Should be **before** routing middleware

doc/book/helpers/server-url-helper.md
```diff
@@ -47,7 +47,8 @@ As such, you will need to:
 
 - Register the `ServerUrlHelper` as a service in your container.
 - Register the `ServerUrlMiddleware` as a service in your container.
-- Register the `ServerUrlMiddleware` as pre_routing pipeline middleware.
+- Register the `ServerUrlMiddleware` as pipeline middleware, immediately
+  following the routing middleware.
```
danizord: **before**, so that you can take advantage of this one even if some error occurs in routing

doc/book/helpers/server-url-helper.md
```diff
@@ -78,20 +79,22 @@ $container->set(
...
+$app->pipeRoutingMiddleware();
+$app->pipeDispatchMiddleware();
 
 // Or use configuration:
 // [
 //     'middleware_pipeline' => [
-//         'pre_routing' => [
-//             ['middleware' => ServerUrlMiddleware::class],
-//         ],
+//         ['middleware' => ServerUrlMiddleware::class, 'priority' => PHP_INT_MAX],
```
danizord: `PHP_INT_MAX` is not a waste of memory?

danizord: Also, as said above it should be piped before routing, so `PHP_INT_MAX` won't work.

weierophinney: @danizord It may or may not be; I need to look up how PHP stores integers (i.e., if it allocates enough bytes to allow for the maximum, or only the amount needed for the integer itself). However, you're wrong about this:

> Also, as said above it should be piped before routing, so PHP_INT_MAX won't work.

Order is defined by using a *priority queue*; as such, *higher* numbers indicate that the middleware is piped to the application *earlier*. So this is correct here.

doc/book/helpers/server-url-helper.md
weierophinney: Can you send a PR with this change, please?

doc/book/helpers/server-url-helper.md
weierophinney: Please send a PR with this change, with the one from your following comment.

doc/book/helpers/server-url-helper.md
weierophinney: @danizord — I've created a PR for this now: #273.

doc/book/helpers/server-url-helper.md
danizord: Ah yes, my bad :D

PR #273, opened 2016-01-18
Fix verbiage about when to register ServerUrlMiddleware

weierophinney:
Examples were correct, but verbiage was incorrectly indicating the middleware should be registered following the routing middleware, instead of before.


weierophinney:
Ping @danizord — here you go. :smile:

PR #274, opened 2016-01-18
use --no-dev flag when installing

weierophinney:
Composer installs the dev requirements *as listed before running any scripts* when executing `create-project`. By specifying `--no-dev`, we prevent that from occurring.

This also adds a note that if you want development requirements, you *must* run `composer update` on completion.

See zendframework/zend-expressive-skeleton#60 for details.


samsonasik:
should be written in readme too https://github.com/zendframework/zend-expressive/blob/master/README.md#using-the-skeleton--installer

PR #277, opened 2016-01-20
Nested error middleware does not work correctly

weierophinney:
Given the following middleware specification:

```php
[
    'middleware' => [
        Unauthorized::class,
        Unauthenticated::class,
    ],
    'error' => true,
    'priority' => -10000,
]
```

Neither middleware listed triggers. Each will trigger, however, if you do them separately, not as part of a list:

```php
[
    'middleware' => Unauthorized::class,
    'error' => true,
    'priority' => -10000,
],
[
    'middleware' => Unauthenticated::class,
    'error' => true,
    'priority' => -10000,
],
```

The error is because the `MarshalMiddlewareTrait` ~~is not retaining the `error` flag when marshaling lazy middleware services.~~ composes a `MiddlewarePipe` instance when creating an error middleware pipeline, but that class does not implement the error middleware signature.

This patch creates a `Zend\Expressive\ErrorMiddlewarePipe` implementation that consumes a `MiddlewarePipe` instance while implementing the error middleware signature. `MarshalMiddlewareTrait` is updated to pass the generated `MiddlewarePipe` instance to `ErrorMiddlewarePipe` when generating an error pipeline.

### DIFF
--- a/src/ErrorMiddlewarePipe.php
+++ b/src/ErrorMiddlewarePipe.php
@@ -1,0 +1,87 @@
+<?php
+/**
+ * Zend Framework (http://framework.zend.com/)
+ *
+ * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
+ * @copyright Copyright (c) 2015-2016 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   https://github.com/zendframework/zend-expressive/blob/master/LICENSE.md New BSD License
+ */
+
+namespace Zend\Expressive;
+
+use Psr\Http\Message\ServerRequestInterface as Request;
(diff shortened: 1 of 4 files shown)


weierophinney:
On testing this in a real-world application, I'm finding it's still not working. Essentially, even though error middleware is being created, when `$next()` is being invoked with an `$error` argument, these are never being hit. I've worked up a full integration test that demonstrates the expected behavior, and am now turning to isolating the root cause.

weierophinney:
#278 resolved the issue I was seeing, and I can report that those changes definitely allow middleware pipelines representing error middleware to work! New RC coming tomorrow!

PR #294, opened 2016-01-27
Update naming modular-layout.md

bakura10:
Hi,

I'd suggest that we update the wording for better consistency and better reflect the fact that your App (the one merging is config), is mostly merging config of modules / standalone-libraries / non-app code (and not other app).

This better reflect the distinction between an app (which is actually the consuming code), and the reusable-code.


weierophinney:
Nice catch, @bakura10 !

PR #303, opened 2016-01-28
[Docs] Update locale regex in examples

moderndeveloperllc:
Update docs to have a better regex for locale discovery. This accounts for 2 or 3 character ISO-639 language codes and optional 2 character ISO-3166 country codes. Also accounts for BCP47-style dashes, or locale-style underscores. I.e.: `en-US`, `en_GB`, `ru`, or `ar-eg`. This keeps the documentation consistent with the example that is using `de_DE` as a locale. Yes this seems trivial, but many devs copy/paste examples like this, so might as well make them as good as possible.

### DIFF
--- a/CONDUCT.md
+++ b/CONDUCT.md
@@ -1,0 +1,43 @@
+# Contributor Code of Conduct
+
+The Zend Framework project adheres to [The Code Manifesto](http://codemanifesto.com)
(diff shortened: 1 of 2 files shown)


weierophinney:
Thanks, @moderndeveloperllc

PR #307, opened 2016-02-02
Handle routers which do not allow runtime configuration

DASPRiD:
Related PRs:

https://github.com/zendframework/zend-expressive-router/pull/6
https://github.com/zendframework/zend-expressive-helpers/pull/9

### DIFF
--- a/test/ApplicationTest.php
+++ b/test/ApplicationTest.php
@@ -31,6 +31,7 @@ use Zend\Expressive\Exception\InvalidMiddlewareException;
 use Zend\Expressive\Router\Route;
 use Zend\Expressive\Router\RouteResult;
 use Zend\Expressive\Router\RouterInterface;
+use Zend\Expressive\Router\RuntimeConfigurableRouterInterface;
 use Zend\Stratigility\MiddlewarePipe;
 use Zend\Stratigility\Route as StratigilityRoute;
 use ZendTest\Expressive\TestAsset\InvokableMiddleware;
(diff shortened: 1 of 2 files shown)


test/ApplicationTest.php
```diff
@@ -149,6 +150,14 @@ public function testCallingRouteWithOnlyAPathRaisesAnException()
         $app->route('/path');
     }
 
+    public function testCallingRouteWithConsumerOnlyRouter()
+    {
+        $this->router = $this->prophesize(RouteInterface::class);
```
samsonasik: s/`RouteInterface::class`/`RouterInterface::class`

DASPRiD: Sneaky little errors, hehe, thanks.

michaelmoussa:
@DASPRiD can you explain this PR a little more? I'm not clear on why it's needed (or if it is even needed anymore). If we take the approach I mention in [this comment](https://github.com/zendframework/zend-expressive-router/pull/6#issuecomment-240421488), then any `RouterInterface`'s `generateUri(...)` method will accept the optional `$options` param, and there'd be no need to distinguish between a plain `RouterInterface` and a `RuntimeConfigurableRouterInterface`.

michaelmoussa:
Closing w/o merge as this will not be necessary due to [zend-expressive-router#6](https://github.com/zendframework/zend-expressive-router/pull/6) - all `RouterInterface` instances will still have the `addRoute` method.

PR #308, opened 2016-02-02
remove extends AbstractPage when using trait

Repkit:
we can remove extends AbstractPage when using trait ActionBasedInvocation


weierophinney:
Thanks, @Repkit

PR #309, opened 2016-02-03
Create configured FinalHandler if one is not available in the container

michaelmoussa:
@weierophinney this piggybacks on what we discussed earlier with regard to the issue that inspired [this PR](https://github.com/zendframework/zend-stratigility/pull/46) on the Stratigility repo.

The first part of the other PR fixes the stack traces in production issue, and the second adds the `setOriginalResponse($response);` to the `FinalHandler` in order to allow someone to create a properly configured `FinalHandler` ahead of time. The `Application` would then inject the original response when it calls `->getFinalHandler()`.

The purpose of this new PR is to provide support for configuring the `FinalHandler` that Expressive automatically creates without needing to create a new factory class. Developers would just need to add the desired final handler options to their config.

**There's a caveat, though - it depends on the Stratigility PR.**

This version of this PR should not be merged until the Stratigility PR is merged, a new version is tagged, and Expressive's `composer.json` entry for `zendframework/zend-stratigility` reflects the dependency on that version. This implementation needs `FinalHandler` to support `setOriginalResponse`, otherwise, `200 OK` responses will return 404s instead (since no original response is being injected into the `FinalHandler`).

I don't think the full solution would be considered a BC break, though, since people who were providing their own `FinalHandler` will not be affected, and people who were relying on the default will not notice a thing.

Let me know if you think a proper `Zend\Expressive\FinalHandlerFactory` that folks need to explicitly add to their dependencies config would be a better solution, otherwise, I'll take care of the `composer.json` update once the other PR is merged and tagged and update this PR.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
         "zendframework/zend-diactoros": "^1.1",
         "zendframework/zend-expressive-router": "^1.1",
         "zendframework/zend-expressive-template": "^1.0.1",
-        "zendframework/zend-stratigility": "^1.1"
+        "zendframework/zend-stratigility": "^1.2.0"
     },
     "require-dev": {
         "filp/whoops": "^1.1",
--- a/src/Container/ApplicationFactory.php
+++ b/src/Container/ApplicationFactory.php
@@ -18,6 +18,7 @@ use Zend\Expressive\Container\Exception\InvalidArgumentException as ContainerInv
 use Zend\Expressive\Router\FastRouteRouter;
 use Zend\Expressive\Router\Route;
 use Zend\Expressive\Router\RouterInterface;
+use Zend\Stratigility\FinalHandler;
 use Zend\Stratigility\MiddlewarePipe;
 
 /**
@@ -144,7 +145,7 @@ class ApplicationFactory
 
         $finalHandler = $container->has('Zend\Expressive\FinalHandler')
             ? $container->get('Zend\Expressive\FinalHandler')
-            : null;
+            : $this->marshalFinalHandler($container);
 
         $emitter = $container->has(EmitterInterface::class)
             ? $container->get(EmitterInterface::class)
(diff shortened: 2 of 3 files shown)


michaelmoussa:
Closing & re-opening to trigger a new Travis build, since my pending `zendframework/zend-stratigility` PR has been merged.

michaelmoussa:
@mwop now that this build will pass once Travis finishes running again, what do you think?

Basically, instead of defaulting an unset `FinalHandler` to `null`, it'll create a new one based on the `final_handler` options available in the configuration.

Alternatively, I could create a `Zend\Expressive\FinalHandlerFactory` to encapsulate the creation-from-config logic, which will allow the default to remain `null`, but will allow users to just configure their DI container to use the Expressive default and not need to write their own to support options-based creation (that's probably the better option, actually, as these days I hate seeing `new` in constructors :smile:). Let me know and I'll go ahead and change it.

src/Container/ApplicationFactory.php
```diff
@@ -142,9 +143,13 @@ public function __invoke(ContainerInterface $container)
...
-        $finalHandler = $container->has('Zend\Expressive\FinalHandler')
-            ? $container->get('Zend\Expressive\FinalHandler')
-            : null;
+        if ($container->has('Zend\Expressive\FinalHandler')) {
+            $finalHandler = $container->get('Zend\Expressive\FinalHandler');
+        } else {
+            $config = $container->has('config') ? $container->get('config') : [];
+            $options = isset($config['final_handler']['options']) ? $config['final_handler']['options'] : [];
+            $finalHandler = new FinalHandler($options, null);
+        }
```
weierophinney: I'd extract the above logic to a private method, so that the logic here can be simply something like `$finalHandler = $this->marshalFinalHandler($container);`.

weierophinney:
> Alternatively, I could create a Zend\Expressive\FinalHandlerFactory to encapsulate the creation-from-config logic

I like this approach, but it would require existing users to update their configuration to add a dependency map for it. As such, let's stick to an internal, private method for this as suggested above.

weierophinney:
One more thing: update the `composer.json` to pin to zend-stratigility `^1.2.0`.

michaelmoussa:
@weierophinney done and done

composer.json
```diff
@@ -22,7 +22,7 @@
         "zendframework/zend-diactoros": "^1.1",
         "zendframework/zend-expressive-router": "^1.1",
         "zendframework/zend-expressive-template": "^1.0.1",
-        "zendframework/zend-stratigility": "^1.1"
+        "zendframework/zend-stratigility": "^1.2.0"
```
bakura10: ^1.2 instead?

weierophinney: They're semantically equivalent. `^` will allow from that version to next
major. (vs `~`, which limits to the same minor version.)

weierophinney:
Thanks for the patch, @michaelmoussa\! Merged to develop for release with 1.1.0.

PR #319, opened 2016-02-13
Fix: Reduce visibility to protected

localheinz:
This PR
- [x] removes visibility of `setUp()` in tests to `protected`

:information_desk_person: There's no need to increase the visibility, is there? For reference, see [`PHPUnit_Framework_TestCase::setUp()`](https://github.com/sebastianbergmann/phpunit/blob/4.7/src/Framework/TestCase.php#L1803-L1809).

### DIFF
--- a/test/ApplicationTest.php
+++ b/test/ApplicationTest.php
@@ -42,7 +42,7 @@ class ApplicationTest extends TestCase
 {
     use ContainerTrait;
 
-    public function setUp()
+    protected function setUp()
     {
         $this->noopMiddleware = function ($req, $res, $next) {
         };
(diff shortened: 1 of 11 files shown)


weierophinney:
Since there will be no classes extending these, there's no reason to change visibility. While the PHPUnit API defines them as protected, the versions in the test cases will still be called when defined as public. Since there's no *technical* rationale behind the change, I'd rather leave as-is, particularly as there are new test classes in the develop branch already that *also* define public `setup()/tearDown()` methods.

localheinz:
@weierophinney 

So why is the visibility changed in the first place?

PR #320, opened 2016-02-13
Fix: Use class keyword where possible

localheinz:
This PR
- [x] makes use of the `class` keyword where possible


weierophinney:
Thanks, @localheinz

PR #321, opened 2016-02-13
Fix: Docblock

localheinz:
This PR
- [x] fixes a docblock where the argument is actually named `$pipeline` instead of `$pipe`


weierophinney:
Thanks, @localheinz

PR #324, opened 2016-02-21
Fixed: the check for the changing of the response body.

overnilx:


weierophinney:
- What is the issue being fixed, exactly?
- Can you please provide unit tests that:
  - Demonstrate how to reproduce the problem?
  - Validate that the change introduced results in the expected behavior?

Thanks in advance!

overnilx:
Sorry, It's too long to remember exactly what is the issue being fixed.

PR #332, opened 2016-04-04
Make InvalidMiddlewareException more verbose

nightlinus:
Add $middleware to exception message if it is string to know problem middleware.

### DIFF
--- a/src/MarshalMiddlewareTrait.php
+++ b/src/MarshalMiddlewareTrait.php
@@ -58,12 +58,16 @@ trait MarshalMiddlewareTrait
         }
 
         if (! is_callable($callable)) {
+            if (is_object($middleware) === true) {
+                $type = get_class($middleware) . '[Object]';
+            } else {
+                $scalarType = gettype($middleware) . '[Scalar]';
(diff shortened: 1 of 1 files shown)


Ocramius:
@nightlinus this change is unclear to me. Can you add a test case showing why this was done?

nightlinus:
Ok, i will add. It is usefull when you use invokable class as middleware and it is not loadable via autoload.
for example in config:
```php
return [
  'path' => '/',
  'middleware' => ShowMain::class,
];
```
Here if something is wrong with autoload you get just ```Unable to resolve middleware "string[Scalar]"``` to a callable'.

harikt:
@nightlinus it looks to me your class `ShowMain` is not callable. And the error you got seems to be correct to me.

nightlinus:
It is callable, but not autoloadable.

harikt:
@nightlinus if that is the case, I don't think you change will mention this class is not auto loadable.

nightlinus:
But it will tell with what class there is a problem.

harikt:
ok. cool. Thank you for the explanation @nightlinus .

PR #340, opened 2016-05-01
Use more inclusive verbiage

i-am-tom:
This is made in reference to point 1 of the code manifesto, really, and updates documentation accordingly.

This might sound like nitpicking, but I think we'd probably be better off avoiding these references to mental health; thoughts?

### DIFF
--- a/features/application/index.html
+++ b/features/application/index.html
@@ -511,7 +511,7 @@ that composes:</p>
 <p>You can define the <code>Application</code> instance in several ways:</p>
 <ul>
 <li>Direct instantiation, which requires providing several dependencies.</li>
-<li>The <code>AppFactory</code>, which will use sane defaults, but allows injecting alternate
+<li>The <code>AppFactory</code>, which will use some common defaults, but allows injecting alternate
   container and/or router implementations.</li>
 <li>Via a dependency injection container; we provide a factory for setting up all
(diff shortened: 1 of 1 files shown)


grizzm0:
The context is not regarding a person. In this case sane would mean something more like reasonable/sensible and has nothing to do with mental health.

i-am-tom:
I totally understand what the word means in this context, and my point is just that there are better ways of phrasing it that don't have any potential to be exclusionary :) This same reasoning is why we avoid words like "retarded" and "dumb" - the context might not have anything to do with humans, but they're needless slurs, and there are plenty of other ways to say the same thing.

weierophinney:
The wording we use is verbatim from the code manifesto; if you have suggestions on wording, please make them against its own repository, and we will pull them in if merged.

akrabat:
This PR is against our text not the text of the code manifesto.

weierophinney:
This is what comes of me reading PRs on my phone... Re-opening, and will re-evaluate. Thanks for the PR, @i-am-tom !

weierophinney:
I back-ported the changes to the documentation source files, which are under the `doc/book/` tree of the master and develop branches, and written in markdown. Documentation rebuilds after pushes to master, so you'll see your change show up shortly.

Thanks for the patch, @i-am-tom !

PR #343, opened 2016-05-07
Improve compatible with PHP7 Errors

basz:
expressive itself was fixed #295 in 1.0, but until whoops 2 whoops couldn't handle it either.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "zendframework/zend-stratigility": "^1.1"
     },
     "require-dev": {
-        "filp/whoops": "^1.1",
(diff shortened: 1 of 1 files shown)


weierophinney:
There's a bit of a problem with just blindly upgrading to whoops 2.0: there *are* some incompatibilities in the new version with the API we were exercising in Expressive; if you look at the [build details](https://travis-ci.org/zendframework/zend-expressive/builds/128539918), you'll see the errors.

I'll see if I can use this as a starting point, however; thanks, @basz.

weierophinney:
#384 accomplishes this comprehensively; closing this in favor of that patch.

PR #346, opened 2016-05-19
fix delegators config key

pine3ree:


samsonasik:
I think the delegator factory class need to be updated as well to implements DelegatorFactoryInterface

pine3ree:
Hello @samsonasik, I agree...

doc/book/cookbook/route-specific-pipeline.md
```diff
@@ -62,7 +62,7 @@ use Zend\ServiceManager\DelegatorFactoryInterface;
 use Zend\ServiceManager\ServiceLocatorInterface;
 use Zend\Stratigility\MiddlewarePipe;
 
-class ApiResourcePipelineDelegatorFactory
+class ApiResourcePipelineDelegatorFactory implements DelegatorFactoryInterface
```
samsonasik: it is better to use `DelegatorFactoryInterface` from `ServiceManager\Factory` namespace so we use : 
```php
public function __invoke(
      ContainerInterface $container,
      $name,
      callable $callback,
      array $options = null
)
```

pine3ree:
@samsonasik 
Yes, right. That was still the deprecated delegator factory.

pine3ree:
@samsonasik 
I hope I didn't miss anything this time. Please verify, :-) thank You. Kind regards.

samsonasik:
it seems ok now 👍

weierophinney:
There's a problem with this patch: the original documentation was demonstrating the zend-servicemanager v2 API, but the changes introduced show only the v3 API. The Expressive installer, and the default configuration, supports *both*. Ideally, we should be:

- implementing `Zend\ServiceManager\DelegatorFactoryInterface`, as originally outlined, *but*
- implementing the methods in both the v2 and v3 versions of the interface.

This means more code in the examples, but will ensure that the code, if copy and pasted into an application, will work regardless of the zend-servicemanager version installed by the user.

If I can get back to this today, I'll use your changes as a starting point. If you do not see activity from me, and you have a chance to update, please do.

Thanks!

pine3ree:
@weierophinney , sure! kind regards!

weierophinney:
@pine3ree I'd updated the zend-form examples with a previous documentation patch, and have now pushed another patch that updates the route-specific pipeline examples. Both now demonstrate both v2 and v3 compatibility for the delegator factory examples.

Issue #350, opened 2016-06-03
Update cookbook for locale routing parameter


xtreamwayz:
I've followed the guidelines for setting up a locale depending on a [routing parameter](https://zendframework.github.io/zend-expressive/cookbook/setting-locale-depending-routing-parameter/).

The problem I have that if a locale is given in the url, I have to inject it inside every route. This has to be done after the LocalizationMiddleware. Since the config and routes are already loaded I found no way to set the current locale as the default locale for every route.

However by a simple hack this can be done automatically:

``` php
<?php

namespace App\Core\Router;

class FastRouteRouter extends \Zend\Expressive\Router\FastRouteRouter
{
    public function generateUri($name, array $substitutions = [])
    {
        if (!array_key_exists('locale', $substitutions)) {
            $substitutions['locale'] = substr(\Locale::getDefault(), 0, 2);
        }

        return parent::generateUri($name, $substitutions);
    }
}
```

Would this be useful to implement this directly or add it to the cookbook?

moderndeveloperllc:
They seem to want all the locale stuff in the cookbook. If they ever [merge this pull request](https://github.com/zendframework/zend-expressive/pull/303/files), you can take out the `substr()` as the locale logic will account for language-country locales. (But not script - sorry Serbians!)

xtreamwayz:
@Ocramius I'm setting it with middleware first:

```php
class LocalizationMiddleware
{
    public function __invoke(ServerRequestInterface $request, ResponseInterface $response, callable $next = null)
    {
        $locale = $request->getAttribute(
            'locale',
            Locale::acceptFromHttp($request->getServerParams()['HTTP_ACCEPT_LANGUAGE'])
        );
        Locale::setDefault($locale);

        if ($request->getUri()->getPath() === '/') {
            return new RedirectResponse('/' . substr(\Locale::getDefault(), 0, 2));
        }

        return $next($request, $response);
    }
}
```

Thanx for the advice on the delegator factory. I'll try that.

Ocramius:
> `Locale::setDefault($locale);`

Argh, we should really kill that stuff :-\

xtreamwayz:
As discussed on IRC, it's better to inject the locale as request attribute and not rely on a static global.

RalfEggert:
@xtreamwayz 

could you document that in the cookbook? Or is that already included in the recipies?

xtreamwayz:
[...]

PR #351, opened 2016-06-04
[docs] Use locale request attribute in favor of the static global locale

xtreamwayz:
As discussed on IRC and in #350.

This changes how the locale is stored in the cookbook example. Since it's not reliable to use the global static `Locale::setDefault()`, it's better to store it as a request attribute.

### DIFF
--- a/doc/book/cookbook/setting-locale-depending-routing-parameter.md
+++ b/doc/book/cookbook/setting-locale-depending-routing-parameter.md
@@ -108,21 +108,34 @@ registered in the pipeline immediately following the routing middleware.
 Such a `LocalizationMiddleware` class could look similar to this:
 
 ```php
+<?php
+
 namespace Application\I18n;
 
 use Locale;
-use Zend\Expressive\Router\RouteResult;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
 
 class LocalizationMiddleware
 {
-    public function __invoke($request, $response, $next)
+    const LOCALIZATION_ATTRIBUTE = 'locale';
+
+    public function __invoke(ServerRequestInterface $request, ResponseInterface $response, callable $next = null)
     {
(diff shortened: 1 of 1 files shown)


RalfEggert:
Great! Need to refactor my own expressive apps now, though.

doc/book/cookbook/setting-locale-depending-routing-parameter.md
```diff
@@ -108,22 +108,35 @@ registered in the pipeline immediately following the routing middleware.
...
+
+    public function __invoke(ServerRequestInterface $request, ResponseInterface $response, callable $next = null)
     {
-        $locale = $request->getAttribute('locale', 'de_DE');
-        Locale::setDefault($locale);
-        return $next($request, $response);
+        // Get locale from route, fallback to the user's browser preference
+        $locale = $request->getAttribute(
+            'locale',
+            Locale::acceptFromHttp($request->getServerParams()['HTTP_ACCEPT_LANGUAGE'] ?? 'en_US')
```
edigu: What happens if we write this as `current($request->getHeader('Accept-language'))` here? Also [Null Coalesce Operator](https://wiki.php.net/rfc/isset_ternary) (`??`) is available only PHP-7 and later, it worth to be noted in comments imho.

xtreamwayz: The Accept-Language can be something like this: `da, en-gb;q=0.8, en;q=0.7`. If you use current and the first element would have this format `en-gb;q=0.8` it would be useless. AFAIK Locale::acceptFromHttp parses the entire strings and gets the preferred language in the right format.

xtreamwayz:
On a side note: Locale::acceptFromHttp() doesn't compare the users preferred languages to the offered languages by the application. The function is pretty limited and only returns the first prefered language. Something like this might help out: http://www.thefutureoftheweb.com/blog/use-accept-language-header

weierophinney:
Thanks, @xtreamwayz

PR #353, opened 2016-06-07
Fix license header

malukenho:
If the tool to check license headers is considered overkill, just tell me, so I can remove it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,8 @@
         "zendframework/zend-expressive-aurarouter": "^1.0",
         "zendframework/zend-expressive-fastroute": "^1.0",
         "zendframework/zend-expressive-zendrouter": "^1.0",
-        "zendframework/zend-servicemanager": "^2.6"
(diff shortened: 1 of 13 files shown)


xtreamwayz:
Considering all the errors found I don't think it's overkill :)

composer.json
```diff
@@ -31,7 +31,8 @@
         "zendframework/zend-expressive-aurarouter": "^1.0",
         "zendframework/zend-expressive-fastroute": "^1.0",
         "zendframework/zend-expressive-zendrouter": "^1.0",
-        "zendframework/zend-servicemanager": "^2.6"
+        "zendframework/zend-servicemanager": "^2.6",
+        "malukenho/docheader": "^0.1.0"
```
samsonasik: should be sorted alphabetically ( before phpunit list )

weierophinney:
Thanks, @malukenho. I modified the docheader file to reflect the style we use, which omits the initial header lines, and then updated all files flagged.

PR #357, opened 2016-06-20
Router ZF2

CrossPlatformCoder:
Code example missing class method


StaveleyHead:
Is this an active project or is it dead? - I supplied this simple document fix a while ago and it is still here?

I am guessing the project is being ignored whilst they work on the 'next big thing' - dissapointing

weierophinney:
Thanks, @CrossPlatformCoder

PR #359, opened 2016-06-21
Added Coveralls Badge to README

Koopzington:
Why not display the badge if we already use Coveralls?


weierophinney:
Thanks, @Koopzington

PR #362, opened 2016-07-01
Change visibility of "injectRoutesAndPipeline" method to protected

mtymek:
Use case: my application is made of a few API endpoints, all of them share the same configuration. In order for them to work together, I need to customize `middleware_pipeline` and `routes` config keys. Easiest way to achieve this is to extend `ApplicationFactory` class, overriding `injectRoutesAndPipeline` method.

### DIFF
--- a/src/Container/ApplicationFactory.php
+++ b/src/Container/ApplicationFactory.php
@@ -163,7 +163,7 @@ class ApplicationFactory
      * @param Application $app
      * @param ContainerInterface $container
      */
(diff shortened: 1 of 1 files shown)


Ocramius:
This is basically the only thing that the `ApplicationFactory` does: please don't extend it to do this, and instead write your own `ApplicationFactory` implementation.

mtymek:
Yeah, I know it is not perfect OOP, however it means a lot of repeated code...

Issue #235, opened 2015-12-21
Allow to segregate middleware on host


bakura10:
Hi,

Similar to how we can segregate by path:

``` php
[
                'path'       => '/api',
                'middleware' => [
                    ApiAuthenticationMiddleware::class,
                    AuthorizationMiddleware::class
                ]
            ]
```

Would it be possible to also segregate by host?

``` php
[
                'host'       => 'api',
                'middleware' => [
                    ApiAuthenticationMiddleware::class,
                    AuthorizationMiddleware::class
                ]
            ]
```

weierophinney:
That's something you can do with nested middleware applications.

As an example, you can have pipeline middleware that checks against the host, and then dispatches another Expressive application based on the match:

```php
$host = $request->getHeaderLine('Host');

switch (strtolower($host)) {
    case 'api':
        $api = $this->container->get('ApiApplication');
        return $api($request, response, $next);
    default:
        return $next($request, $response);
}
```

Just a quick point of clarification: the path-based matching in pipeline middleware is a feature of Stratigility, which inherits the feature from Sencha Connect. Its primary purpose is path-based segregation; if you want to do host-based segregation, you either run the application on a separate host, or you use a solution like I outline above.

bakura10:
Sounds like a nice solution. But maybe Stratigility implementation could deviate a bit to add this feature? That sounds like a pretty simple feature, as today in Expressive there is basically no easy way through config to perform routing based on host.

weierophinney:
@bakura10 I don't want to bloat Stratigility too much, but that should be relatively easy to add. Want to contribute a PR?

bakura10:
Yep will try to do that when I have a bit of time :).

artwis:
Hey, has anyone worked on this feature? If not I'd be willing to submit PRs to Stratigility and Expressive to allow host based middleware pipe routing.

webimpress:
It's possible now with Expressive 3. See:
- https://github.com/zendframework/zend-stratigility/pull/142
- https://docs.zendframework.com/zend-stratigility/v3/api/#host

PR #369, opened 2016-08-10
Host routing

artwis:
This PR adds segregating middlewares by host in Application's middleware pipe, implements https://github.com/zendframework/zend-expressive/issues/235

```php
// in middleware-pipeline.global.php

'api' => [
    'path' => '/',
    'host' => 'api.example.com',
    'middleware' => [
        ...
    ],
],
```

Disclaimer: it depends on my PR for Stratigility https://github.com/zendframework/zend-stratigility/pull/62
Feedback would be very welcome :)


artwis:
Wrongly added, ignore

PR #370, opened 2016-08-10
Allow to segregate middleware on host

artwis:
This PR adds segregating middlewares by host in Application's middleware pipe, implements https://github.com/zendframework/zend-expressive/issues/235

``` php
// in middleware-pipeline.global.php

'api' => [
    'path' => '/',
    'host' => 'api.example.com',
    'middleware' => [
        ...
    ],
],
```

Disclaimer: it depends on my PR for Stratigility https://github.com/zendframework/zend-stratigility/pull/62
Feedback would be very welcome.

### DIFF
--- a/doc/book/features/application.md
+++ b/doc/book/features/application.md
@@ -145,19 +145,19 @@ and, more specifically, its `MiddlewarePipe` definition, you can also pipe
 should execute on each request, defining error handlers, and/or segregating
 applications by subpath.
 
-The signature of `pipe()` is:
+`pipe()` can be called with up to three arguments:
 
 ```php
-public function pipe($pathOrMiddleware, $middleware = null)
+public function pipe($middleware)
+public function pipe(string $path, $middleware)
+public function pipe(string $path, string $host, $middleware)
 ```
 
 where:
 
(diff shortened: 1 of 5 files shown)


weierophinney:
We have decided not to incorporate zendframework/zend-stratigility#62; as such, this would need to be re-worked to not depend on that feature.

weierophinney:
@artwis Do you want to re-work this to be standalone within Expressive, or should I close it?

artwis:
@weierophinney Yes, I'll take another shot on this one, hopefully next week!

weierophinney:
This is now possible starting with version 3.0.0alpha2, and using zend-stratigility 3.0.0alpha2, via the utility method `Zend\Stratigility\host()`:

```php
// For now, assume the following; we'll likely be updating the skeleton to make the factory
// generally available.
$container = require 'config/container.php';
$factory = new MiddlewareFactory(new MiddlewareContainer($container));

$app->pipe(host('api.example.com', $factory->prepare(SomeMiddlewareService::class)));
```

Essentially, we felt this was a generalized tool useful to zend-stratigility, and with the changes in Expressive v3, we can accommodate such tooling without needing to alter Expressive itself.

PR #373, opened 2016-08-31
Intercept exceptions from the ServerRequest factory

weierophinney:
As noted in zendframework/zend-diactoros#171, the `ServerRequestFactory::fromGlobals()` method raises exceptions on malformed information, such as:

- malformed file upload data
- invalid HTTP protocol versions

At the application level, we should catch these, and invoke the final handler with a 400 response when caught.

I've opened this against master, as I consider the current behavior a "bug"; however, in a way, it's technically a new feature. I'd appreciate comments as to the preferred release. :smile:

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -11,12 +11,15 @@ namespace Zend\Expressive;
 
 use Interop\Container\ContainerInterface;
 use Interop\Container\Exception\ContainerException;
+use InvalidArgumentException;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
+use UnexpectedValueException;
 use Zend\Diactoros\Request;
 use Zend\Diactoros\Response;
 use Zend\Diactoros\Response\EmitterInterface;
 use Zend\Diactoros\Response\SapiEmitter;
+use Zend\Diactoros\ServerRequest;
 use Zend\Diactoros\ServerRequestFactory;
 use Zend\Stratigility\MiddlewarePipe;
 
@@ -552,10 +555,19 @@ class Application extends MiddlewarePipe implements Router\RouteResultSubjectInt
      */
     public function run(ServerRequestInterface $request = null, ResponseInterface $response = null)
     {
-        $request  = $request ?: ServerRequestFactory::fromGlobals();
-        $response = $response ?: new Response();
+        try {
+            $request  = $request ?: ServerRequestFactory::fromGlobals();
+        } catch (InvalidArgumentException $e) {
+            // Unable to parse uploaded files
+            $this->emitMarshalServerRequestException($e);
+            return;
+        } catch (UnexpectedValueException $e) {
+            // Invalid request method
+            $this->emitMarshalServerRequestException($e);
+            return;
+        }
 
-        $response = $this($request, $response);
+        $response = $this($request, $response ?: new Response());
 
         $emitter = $this->getEmitter();
         $emitter->emit($response);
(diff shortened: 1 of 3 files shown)


weierophinney:
Ping @michaelmoussa — thoughts?

src/Application.php
```diff
@@ -100,6 +103,13 @@ class Application extends MiddlewarePipe implements Router\RouteResultSubjectInt
     private $routes = [];
 
     /**
+     * Internal; factory class for server request (for mocking)
+     *
+     * @var string
+     */
+    private $serverRequestFactory = ServerRequestFactory::class;
```
michaelmoussa: @weierophinney you _can_ mock this using Mockery without this private property + reflection if desired, by the way.

```php
/**
 * @runInSeparateProcess
 * @preserveGlobalState disabled
 */
public function testSomething()
{
    $serverRequestFactory = m::mock('alias:' . ServerRequestFactory::class);
    $serverRequestFactory->shouldReceive('fromGlobals')
        ->once()
        ->andReturn(['foo' => 'bar']);

    $this->assertSame(['foo' => 'bar'], ServerRequestFactory::fromGlobals());
}
```

Both annotations are required for this to work, else you will hit:

```
Mockery\Exception\RuntimeException: Could not load mock Zend\Diactoros\ServerRequestFactory, class already exists
```

No real harm in leaving it how it is now, but IMO the less "this is only used for testing" stuff in real classes, the better.

michaelmoussa:
@weierophinney PR looks good overall. Only suggestion was using Mockery alias mocks, which would allow us to remove an otherwise unnecessary private property as well as the two test asset classes.

Regarding bug vs feature, this looks like more of a "doing it this way is better than we are currently doing it" than "it is currently wrong and we must fix". Probably makes more sense for a minor release than a patch release.

weierophinney:
> Only suggestion was using Mockery alias mocks, which would allow us to remove an otherwise unnecessary private property as well as the two test asset classes.

I actually tried those first, but could not get them to work; I suspect the fact that the `ServerRequestFactory` was already autoloaded was getting in the way. One thing I did not try was running the test in a separate process, however; I'll give that a try shortly.

Regarding minor release vs patch, I'll rebase this against develop, and then update the PR to base it against there.

Thanks for the feedback, @michaelmoussa !

weierophinney:
Initial run worked, but then when I started removing the extraneous code, I started getting "serialization of closure is not allowed' errors from PHPUnit. Trying to debug...

weierophinney:
So, the solution was to wrap the test case in a try/catch block, which helped me discover that the response passed to the emitter was not the expected one... which brings me back to the original issue I had with `alias:`. I'll keep trying, though. I'd rather not have workarounds just for testing.

michaelmoussa:
@weierophinney "Serialization of Closure" in that context is usually due to what's being output by the test run in isolation. IIRC the tests in isolation communicate back to the main PHPUnit process by spitting out a serialized response which the main process reads.

If you get annoyed enough, feel free to share what you've got and I'll see if anything screams at me.

weierophinney:
@michaelmoussa Got it sorted; wrapped the internals in a try/catch, and the problem was with assertions in the closure failing. By wrapping in a try/catch, I was able to raise a failure assertion from the catch blocks.

The result: I discovered I wasn't importing `ServerRequestFactory`, which meant my double wasn't. :smile:

All is working now.

I rebased against develop, and squashed my commits so that we don't have a history of those unnecessary test assets, and then re-targeted this PR against the develop branch. Ready for review again!

test/ApplicationTest.php
```diff
@@ -622,4 +626,64 @@ public function testPipingNotInvokableErrorMiddlewareRisesException()
...
+     * @runInSeparateProcess
+     * @preserveGlobalState disabled
+     * @dataProvider invalidRequestExceptions
+     */
+    public function testRunInvokesFinalHandlerWhenServerRequestFactoryRaisesException(
+        $expectedException,
+        $message
+    ) {
+        // try/catch is necessary in the case that the test fails.
+        // Assertion exceptions raised inside prophecy expectations normally
```
danizord: s/prophecy/mockery :P

danizord: haaa nvm :D

michaelmoussa:
LGTM 👍

weierophinney:
@michaelmoussa I'll have you merge; I've got a few more proposals for 1.1.0 to sort out before we do a release, though, so don't tag. :)

michaelmoussa:
@weierophinney done. I used `feature/373` as the local merge branch rather than `hotfix` since you changed the base branch to `develop` and are planning to bundle in some other items before release. Hope that was correct.

weierophinney:
Perfect!

PR #375, opened 2016-09-07
Update setting-locale-without-routing-parameter.md

bartoszzajac:
There was missing ")" in line 49


weierophinney:
Fixed previously with #323.

Issue #380, opened 2016-09-16
Regarding composer suggestions...


Koopzington:
Since [zend-expressive-router suggests installing the routers](https://github.com/zendframework/zend-expressive-router/blob/master/composer.json#L37) and [zend-expressive-template suggests the template renderers](https://github.com/zendframework/zend-expressive-template/blob/master/composer.json#L33) and since [these 2 packages are in the require-section of zend-expressive's composer.json](https://github.com/zendframework/zend-expressive/blob/master/composer.json#L23) wouldn't it make sense to remove the suggestions in zend-expressive's composer.json for 
zend-expressive-aurarouter,
zend-expressive-fastroute,
zend-expressive-zendrouter,
zend-expressive-platesrenderer,
zend-expressive-twigrenderer,
and zend-expressive-zendviewrenderer?

harikt:
I believe you are correct. Probably you should consider sending a PR. If busy I can send one.

weierophinney:
Closed with #381.

PR #381, opened 2016-09-17
Remove unnecessary suggestions

Koopzington:
As mentioned in #380 These suggestions are already getting suggested by zend-expressive-template and zend-expressive-router.


weierophinney:
Thanks, @Koopzington

PR #382, opened 2016-09-18
Correct the documentation for configuring a container-driven container

settermjd:
I'm not sure if the documentation was correct previously, but whenever
I've tried to follow it, to the letter, it's resulted in segmentation
faults and the routes never being loaded. Given that, after some further
research, I believe that the previous recommendation was never going to
work. So, I've migrated some of the configuration from a setup using the
skeleton installer and updated the docs to reflect that.

I don't mean what I've said in any kind of mean or uncharitable way. 
Perhaps the configuration worked up until a particular revision, but was
not updated to reflect subsequent changes. Or, perhaps the fault lies with
me, and I wasn't as thorough as I thought I was. Either way, I offer this for 
consideration and review.

### DIFF
--- a/doc/book/reference/usage-examples.md
+++ b/doc/book/reference/usage-examples.md
@@ -276,81 +276,86 @@ In `config/config.php`, place the following:
 
 ```php
 <?php
-use Zend\Config\Factory as ConfigFactory;
+
+use Zend\Stdlib\ArrayUtils;
 use Zend\Stdlib\Glob;
 
-$files = Glob::glob('config/autoload/{{,*.}global,{,*.}local}.php', Glob::GLOB_BRACE);
-if (0 === count($files)) {
-    return [];
+$config = [];
+// Load configuration from autoload path
+foreach (Glob::glob('config/autoload/{{,*.}global,{,*.}local}.php', Glob::GLOB_BRACE) as $file) {
+    $config = ArrayUtils::merge($config, include $file);
 }
-return ConfigFactory::fromFiles($files);
+
+// Return an ArrayObject so we can inject the config as a service in Aura.Di
+// and still use array checks like ``is_array``.
+return new ArrayObject($config, ArrayObject::ARRAY_AS_PROPS);
 ```
 
-In `config/autoload/global.php`, place the following:
+In `config/container.php`, place the following:
 
 ```php
 <?php
-return [
-    'routes' => [
-        [
-            'path' => '/',
-            'middleware' => 'Application\HelloWorld',
-            'allowed_methods' => [ 'GET' ],
-        ],
-        [
-            'path' => '/ping',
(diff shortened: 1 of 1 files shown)


doc/book/reference/usage-examples.md
```diff
@@ -276,81 +276,86 @@ In `config/config.php`, place the following:
...
+            'path' => '/ping',
+            'middleware' => \Application\PingAction::class,
+            'allowed_methods' => [ 'GET' ],
+        ],
+    ],
+];
 ```
 
-In `src/HelloWorld.php`, place the following:
+In `src/Application/HelloWorld.php`, place the following:
```
samsonasik: s/Application/App

doc/book/reference/usage-examples.md
```diff
@@ -276,81 +276,86 @@ In `config/config.php`, place the following:
...
-    ],
-    'invokables' => [
-        'Application\HelloWorld' => 'Application\HelloWorld',
-        'Application\Ping' => 'Application\Ping',
-    ],
-    'factories' => [
-        'Zend\Expressive\Application' => 'Zend\Expressive\Container\ApplicationFactory',
-    ],
+    'dependencies' => [
+        'invokables' => [
```
samsonasik: I'm thinking about encouraging to use `zend-servicemanager` v3 with `InvokableFactory::class` usage

doc/book/reference/usage-examples.md
```diff
@@ -379,7 +384,7 @@ class Ping
 }
 ```
 
-Finally, in `public/index.php`, place the following:
+After that’s done run `composer dump-autoload` from the command-line, in the root directory of your project. Finally, in `public/index.php`, place the following:
```
samsonasik: `composer dump-autoload` should be placed with different paragraph imo, so something like:

After that’s done, run:
```
 composer dump-autoload
```

settermjd:
@samsonasik all these seem perfectly reasonable to me.

doc/book/reference/usage-examples.md
settermjd: That's not the correct spot to make that change. I had a further look and needed to change App to Application in the class files (Ping, HelloWorld) to match the rest of the code.

doc/book/reference/usage-examples.md
settermjd: Nice and concise.

settermjd:
@samsonasik all changes made.

doc/book/reference/usage-examples.md
weierophinney: This, however, is cross-component compatible.

weierophinney:
Thanks, @settermjd\!

settermjd:
No worries @weierophinney.

Issue #383, opened 2016-09-21
Issue with how WhoopsErrorHandler processes a Whoops object with multiple handlers in its stack


alrav:
Hello all - I believe this issue becomes apparent when you use the `WhoopsErrorHandler` along with `WhoopsFactory` - which is configured to output AJAX errors in JSON format, based on the below configuration:

```
    'whoops' => [
        'json_exceptions' => [
            'display'    => true,
            'show_trace' => true,
            'ajax_only'  => true,
        ],
    ],
```

However, it always outputs in HTML instead.

The reason is because the `WhoopsErrorHandler` always pushes the injected `PrettyPageHandler` ($this->whoopsHandler) object to the top of the stack, even if other handlers were already configured in Whoops:

```
// in WhoopsErrorHandler::handleException
$this->whoops->pushHandler($this->whoopsHandler);
```

So you might have a `Whoops` handler stack like the one created by `WhoopsFactory`:
1. `PrettyPageHandler`
2. `JsonResponseHandler`

But, `WhoopsErrorHandler` will always add `PrettyPageHandler` to the top, which gets popped first and, therefore, not allowing `JsonResponseHandler` to be called. The final stack would look like:
1. `PrettyPageHandler`
2. `JsonResponseHandler`
3. `PrettyPageHandler`

xtreamwayz:
I have tried to debug this as well I didn't get anywhere with `WhoopsFactory`. The only way I got this working properly was hacking the `WhoopsErrorHandler`. It needs some more work to add the json exception config.

```php
// Config
'dependencies' => [
    'invokables' => [
        'Zend\Expressive\Whoops' => Whoops\Run::class,
        'Zend\Expressive\WhoopsPageHandler' => Whoops\Handler\PrettyPageHandler::class,
    ],
    'factories' => [
        'Zend\Expressive\FinalHandler' => Zend\Expressive\Container\WhoopsErrorHandlerFactory::class,
    ],
],
```

```php
// Zend\Expressive\WhoopsErrorHandler
protected function handleException($exception, Request $request, Response $response)
{
    if (in_array('xmlhttprequest', $request->getHeader('x-requested-with'))) {
        $this->whoops->writeToOutput(false);
        $jsonHandler = new JsonResponseHandler();
        $jsonHandler->addTraceToOutput(true);
        $this->whoops->pushHandler($jsonHandler);
    } else {
        $this->prepareWhoopsHandler($request);
[...]

PR #384, opened 2016-09-21
Add support for Whoops 2.0

xtreamwayz:
This PR adds support for Whoops 2.0 as discussed in zendframework/zend-expressive-skeleton#115 and #383.

Falling back to 1.1 is no option since it doesn't support PHP 7.

It's not optimal but it adds support fro 2.0 while maintaining BC.
- [x] Add whoops 2.0 support
- [x] Suggest whoops 2.0
- [x] Fix whoops tests

### DIFF
--- a/src/WhoopsErrorHandler.php
+++ b/src/WhoopsErrorHandler.php
@@ -51,7 +51,7 @@ class WhoopsErrorHandler extends TemplatedErrorHandler
      */
     public function __construct(
         Whoops $whoops,
-        PrettyPageHandler $whoopsHandler,
+        PrettyPageHandler $whoopsHandler = null,
         Template\TemplateRendererInterface $renderer = null,
         $template404 = 'error/404',
         $templateError = 'error/error',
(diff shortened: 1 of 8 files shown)


xtreamwayz:
@michaelmoussa Could you review this code please. If this is a valid solution I'll fix the tests too. I don't want to spend time on the tests if the solution isn't good.

src/WhoopsErrorHandler.php
```diff
@@ -76,9 +76,18 @@ public function __construct(
      */
     protected function handleException($exception, Request $request, Response $response)
     {
-        $this->prepareWhoopsHandler($request);
+        // Push the whoops handler if any
+        if (null !== $this->whoopsHandler) {
```
michaelmoussa: Can you reverse the order here for the comparison for consistency? I haven't seen the Yoda condition style being used elsewhere in the project.

xtreamwayz:
The tests are now compatible for Whoops 1.x and 2.x.

For Expressive version 2 I suggest to drop whoops 1.x support. All Whoops related tests can then be rewritten and properly use the Whoops\RunInterface (this isn't available in Whoops 1.x).

One more thing I discovered: The Zend\Expressive\WhoopsErrorHandler expects as the second parameter the PrettyPageHandler. However there is also a [Whoops\Handler\HandlerInterface](https://github.com/filp/whoops/blob/master/src/Whoops/Handler/HandlerInterface.php). This would allow to inject any error handler like the JsonResponseHandler without breaking BC.

michaelmoussa:
Great! I'll see about merging and doing a patch release early next week, as I won't have access to a computer for a few days... unless @weierophinney or someone else wants to wrap things up.

michaelmoussa:
@weierophinney this looks good to me, but I noticed that `master` is way ahead of the latest `1.0.0` tag for this repo. It seems to be mostly documentation and CI related commits, though. Do you see any issue with me merging this PR and releasing `1.0.1`? (it's a bug fix and is BC, so I think a patch version is appropriate)

kynx:
This is working well for me :) Would be good to see it merged.

weierophinney:
Thanks, @xtreamwayz

Issue #390, opened 2016-11-11
Application::run incompatible with zend-stratigility 1.3


mryvlin:
When invoking the Application, the third param is missing in the request, but as of zend-stratigility 1.3 the third param is now required.
Here ist the error message:

```
[Fri Nov 11 11:30:33 2016] PHP Deprecated:  The third argument to Zend\Stratigility\MiddlewarePipe() ($out) will be required starting with Stratigility version 2; please see https://docs.zendframework.com/zend-stratigility/migration/to-v2/ for more details on how to update your application to remove this message. in /tmp/test-app/vendor/zendframework/zend-stratigility/src/MiddlewarePipe.php on line 101
[Fri Nov 11 11:30:33 2016] PHP Stack trace:
[Fri Nov 11 11:30:33 2016] PHP   1. {main}() /tmp/test-app/public/index.php:0
[Fri Nov 11 11:30:33 2016] PHP   2. Zend\Expressive\Application->run() /tmp/test-app/public/index.php:21
[Fri Nov 11 11:30:33 2016] PHP   3. Zend\Expressive\Application->__invoke() /tmp/test-app/vendor/zendframework/zend-expressive/src/Application.php:558
[Fri Nov 11 11:30:33 2016] PHP   4. Zend\Stratigility\MiddlewarePipe->__invoke() /tmp/test-app/vendor/zendframework/zend-expressive/src/Application.php:141
[Fri Nov 11 11:30:33 2016] PHP   5. trigger_error() /tmp/test-app/vendor/zendframework/zend-stratigility/src/MiddlewarePipe.php:101
[Fri Nov 11 11:30:33 2016] PHP Deprecated:  The third argument to Zend\Stratigility\MiddlewarePipe() ($out) will be required starting with Stratigility version 2; please see https://docs.zendframework.com/zend-stratigility/migration/to-v2/ for more details on how to update your application to remove this message. in /tmp/test-app/vendor/zendframework/zend-stratigility/src/MiddlewarePipe.php on line 101
[Fri Nov 11 11:30:33 2016] PHP Stack trace:
[Fri Nov 11 11:30:33 2016] PHP   1. {main}() /tmp/test-app/public/index.php:0
[Fri Nov 11 11:30:33 2016] PHP   2. Zend\Expressive\Application->run() /tmp/test-app/public/index.php:21
[...]

PR #391, opened 2016-11-11
Ensure `Application::__invoke()` does not trigger deprecations

weierophinney:
Stratigility 1.3 deprecates having an optional `$out` argument to `__invoke()`. However, `Application::run()` was calling without the argument, to allow lazy-creation of a final handler if none was injected at instantiation.

In order to mimic the behavior of `MiddlewarePipe`, this meant testing:

- For a null `$out` argument
- A null return value from `getFinalHandler()`
- Decorating a PSR response with a Stratigility response
- Seeding a Stratigility `FinalHandler` with the decorated response

I removed a condition from an error handler that was checking for deprecations in order to ensure that deprecation notices *were* being emitted previously, and after changes, were not. Interestingly, these were not triggering failures for me locally; I was only seeing the deprecation notices within the PHPUnit progress output. As such, verification of this patch will involve checking the Travis logs.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -15,6 +15,8 @@ use Zend\Diactoros\Response;
 use Zend\Diactoros\Response\EmitterInterface;
 use Zend\Diactoros\Response\SapiEmitter;
 use Zend\Diactoros\ServerRequestFactory;
+use Zend\Stratigility\FinalHandler;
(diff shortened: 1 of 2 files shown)


weierophinney:
Fixes #390.

PR #392, opened 2016-11-11
Add the new error handler classes

xtreamwayz:
These are the new error handler classes for Stratigility 1.3+.

PR #376, opened 2016-09-10
Prepare 1.1 release

bakura10:
Hi !

This PR removes all features that was deprecated in pre 1 version (as per comment in the code).

Now that 5.5 is not supported, I've raised dependency to 5.6 and switch to variadic where possible.

PR #396, opened 2016-11-11
Implement programmatic pipeline features

weierophinney:
This patch implements the [Programmatic Pipelines RFC requirements](https://gist.github.com/weierophinney/a983f733c2ff3c9b3a687625e012e6e1#requirements-for-expressive-11).

TODO
----

- [x] Create `Zend\Expressive\Middleware\ErrorResponseGenerator`
- [x] Create `Zend\Expressive\Middleware\WhoopsErrorResponseGenerator`
- [x] Create `Zend\Expressive\Middleware\NotFoundHandler`
- [x] Create `Zend\Expressive\Container\ErrorResponseGeneratorFactory`
- [x] Create `Zend\Expressive\Container\WhoopsErrorResponseGeneratorFactory`
- [x] Create `Zend\Expressive\Container\NotFoundHandlerFactory`
- [x] Create `Zend\Expressive\Container\ErrorHandlerFactory`
- [x] Create `Zend\Expressive\ApplicationConfigInjectionTrait`, and compose it into `Application`.
- [x] Update `Zend\Expressive\Container\ApplicationFactory` to vary creation and return of the `Application` instance based on the `zend-expressive.programmatic_pipeline` configuration flag.
- [x] Update `Zend\Expressive\Container\ApplicationFactory` to use the methods defined in `ApplicationConfigInjectionTrait` to inject pipeline and routed middleware from configuration into the `Application` instance.
- [x] Update `Zend\Expressive\Application` to emit a deprecation notice from `pipeErrorMiddleware()`.
- [x] Update `Zend\Expressive\MarshalMiddlewareTrait` to allow marshalling http-interop middleware.
- [x] Add support for "raise throwables" flag to `ApplicationFactory`.
- [x] Documentation

## Tooling support

The following tools were written in [zendframework/zend-expressive-tooling](https://github.com/zendframework/zend-expressive-tooling) to support user migrations to work with the above changes:

- [x] Tool for generating programmatic pipelines and routing from existing configuration.
- [x] Tool for migrating `getOriginal*()` calls to `getAttribute(*)` calls on the request object.
- [x] Tool for identifying error middleware in an existing application.

References
----------

- [Programmatic Pipelines RFC](https://gist.github.com/weierophinney/a983f733c2ff3c9b3a687625e012e6e1)
- #392 (original draft of error handler classes)
- [Stratigility 1.3/2.0 migration docs](https://docs.zendframework.com/zend-stratigility/migration/to-v2/)


weierophinney:
Ready for review! Pinging:

- @zendframework/community-review-team 
- @michaelmoussa 
- @xtreamwayz 
- @ezimuel 

Thanks in advance!

composer.json
```diff
@@ -52,15 +54,18 @@
...
         "check": [
             "@cs-check",
             "@test"
         ],
         "upload-coverage": "coveralls -v",
-        "cs-check": "phpcs",
-        "cs-fix": "phpcbf",
-        "test": "phpunit",
+        "cs-check": "phpcs --colors",
+        "cs-fix": "phpcbf --colors",
```
Koopzington: --colors for phpcs and phpcbf is not needed since the new coding standard will have them enabled via the ruleset.xml

doc/book/reference/migration/to-v1.1.md
```diff
@@ -0,0 +1,560 @@
...
+include:
+
+- `get($path, $middleware, $name = null)`
+- `post($path, $middleware, $name = null)`
+- `put($path, $middleware, $name = null)`
+- `patch($path, $middleware, $name = null)`
+- `delete($path, $middleware, $name = null)`
+- `route($path, $middleware, array $methods = null, $name = null)`
+
+Each returns a `Zend\Expressive\Router\Route` instance; this is useful if you
```
xtreamwayz: Thanx for mentioning this. No need anymore for routes via config.

doc/book/reference/migration/to-v1.1.md
```diff
@@ -0,0 +1,560 @@
...
+
+Starting in Stratigility 2.0 and Expressive 2.0, `Application` will continue to
+accept the legacy double-pass signature, but will require that you either:
+
+- Provide a `$responsePrototype` (a `ResponseInterface` instance) to the
+  `Application` instance prior to piping or routing such middleware.
+- Decorate the middleware in a `Zend\Stratigility\Middleware\CallableMiddlewareWrapper`
+  instance (which also requires a `$responsePrototype`).
+
+We recommend that you begin writing middleware to follow the http-interop
```
xtreamwayz: Maybe add an example here to clarify it a bit. I think it's something like this? (Disclaimer: from memory and untested)

```php
<?php

namespace App\Middleware;

use Interop\Http\Middleware\ServerMiddlewareInterface;
use Interop\Http\Middleware\DelegateInterface;

class MyMiddleware implements ServerMiddlewareInterface
{
    public function __construct()
    {
    }

    /**
     * @param Request           $request
     * @param DelegateInterface $delegate
     *
     * @return Response
     */
    public function process(ServerRequestInterface $request, DelegateInterface $delegate): ResponseInterface
    {
        // Do something (with the request) first

        // Call the next middleware and wait for the response
        $response = $delegate->process($request);

        // Do something (with the response) before returning the response

        // Return the response
        return $response;
    }
}
```

doc/book/reference/migration/to-v1.1.md
weierophinney: This has actually been true all along... just not well documented.

doc/book/reference/migration/to-v1.1.md
weierophinney: Excellent feedback! I'll incorporate that shortly!

doc/book/reference/migration/to-v1.1.md
```diff
@@ -0,0 +1,560 @@
...
+  callable $next)`). This middleware would be invoked when calling `$next()`
+  with a third argument indicating an error, and would be expected to handle it
+  or delegate to the next error middleware.
+
+  Internally, Stratigility would execute each middleware within a try/catch
+  block; if an exception were caught, it would then delegate to the next _error
+  middleware_ using the caught exception as the `$err` argument.
+
+- The "Final Handler". This is a special middleware type with the signature
+  `function ServerRequestInterface $request, ResponseInterface $response, $err = null)`,
```
harikt: Probably missed a `(`

doc/book/reference/migration/to-v1.1.md
```diff
@@ -0,0 +1,560 @@
...
+
+  Once enabled, any `middleware_pipeline` or `routes` configuration will be
+  ignored when creating the `Application` instance.
+
+- Second, update your `public/index.php` to add the following lines immediately
+  prior to calling `$app->run();`:
+
+  ```php
+  include 'config/pipeline.php';
+  include 'config/routes.php';
```
glen-84: Shouldn't these use `require`?

test/ApplicationConfigInjectionTest.php
```diff
@@ -0,0 +1,414 @@
...
+/**
+ * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
+ * @copyright Copyright (c) 2015-2016 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   https://github.com/zendframework/zend-expressive/blob/master/LICENSE.md New BSD License
+ */
+
+namespace ZendTest\Expressive;
+
+use Closure;
+use InvalidArgumentException;
```
webimpress: It is never used.

src/ApplicationConfigInjectionTrait.php
```diff
@@ -0,0 +1,307 @@
+<?php
+/**
+ * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
+ * @copyright Copyright (c) 2015-2016 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   https://github.com/zendframework/zend-expressive/blob/master/LICENSE.md New BSD License
+ */
+
+namespace Zend\Expressive;
+
+use ReflectionProperty;
```
webimpress: It is never used.

src/Container/ApplicationFactory.php
```diff
@@ -7,16 +7,16 @@
 
 namespace Zend\Expressive\Container;
 
+use ArrayObject;
 use Interop\Container\ContainerInterface;
-use SplPriorityQueue;
 use Zend\Diactoros\Response\EmitterInterface;
 use Zend\Expressive\Application;
+use Zend\Expressive\ApplicationUtils;
 use Zend\Expressive\Exception;
```
webimpress: It is no longer used here, can be removed.

test/Container/ErrorResponseGeneratorFactoryTest.php
```diff
@@ -0,0 +1,82 @@
...
+ * @copyright Copyright (c) 2016-2016 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   http://framework.zend.com/license/new-bsd New BSD License
+ */
+
+namespace ZendTest\Expressive\Container;
+
+use Interop\Container\ContainerInterface;
+use PHPUnit_Framework_TestCase as TestCase;
+use Zend\Expressive\Container\ErrorResponseGeneratorFactory;
+use Zend\Expressive\Middleware\ErrorResponseGenerator;
```
webimpress: It is never used.

bin/expressive-pipeline-from-config
```diff
@@ -0,0 +1,27 @@
...
+ * Script for migrating configuration-driven pipelines/routes to programmatic usage.
+ *
+ * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
+ * @copyright Copyright (c) 2015-2016 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   https://github.com/zendframework/zend-expressive/blob/master/LICENSE.md New BSD License
+ */
+
+namespace Zend\Expressive;
+
+use Zend\Stdlib\ConsoleHelper;
```
webimpress: Seems to be unused in this file.

test/Command/GenerateProgrammaticPipelineFromConfig/GeneratorTest.php
```diff
@@ -0,0 +1,158 @@
...
+ * @copyright Copyright (c) 2015-2016 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   https://github.com/zendframework/zend-expressive/blob/master/LICENSE.md New BSD License
+ */
+
+namespace ZendTest\Expressive\Command\GenerateProgrammaticPipelineFromConfig;
+
+use org\bovigo\vfs\vfsStream;
+use org\bovigo\vfs\vfsStreamDirectory;
+use PHPUnit_Framework_TestCase as TestCase;
+use Prophecy\Argument;
```
webimpress: Unused, please remove.

test/Command/GenerateProgrammaticPipelineFromConfig/GeneratorTest.php
```diff
@@ -0,0 +1,158 @@
...
+/**
+ * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
+ * @copyright Copyright (c) 2015-2016 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   https://github.com/zendframework/zend-expressive/blob/master/LICENSE.md New BSD License
+ */
+
+namespace ZendTest\Expressive\Command\GenerateProgrammaticPipelineFromConfig;
+
+use org\bovigo\vfs\vfsStream;
+use org\bovigo\vfs\vfsStreamDirectory;
```
webimpress: Unused, please remove.

test/Command/GenerateProgrammaticPipelineFromConfig/CommandTest.php
```diff
@@ -0,0 +1,138 @@
...
+/**
+ * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
+ * @copyright Copyright (c) 2015-2016 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   https://github.com/zendframework/zend-expressive/blob/master/LICENSE.md New BSD License
+ */
+
+namespace ZendTest\Expressive\Command\GenerateProgrammaticPipelineFromConfig;
+
+use org\bovigo\vfs\vfsStream;
+use org\bovigo\vfs\vfsStreamDirectory;
```
webimpress: Unused, please remove.

src/Command/GenerateProgrammaticPipelineFromConfig/Command.php
```diff
@@ -0,0 +1,165 @@
...
+
+    /**
+     * @var ConsoleHelper
+     */
+    private $helper;
+
+    /**
+     * @param ConsoleHelper $helper
+     */
+    public function __construct($command = self::DEFAULT_COMMAND_NAME, ConsoleHelper $helper = null)
```
webimpress: Missing `$command` in PHPDoc

src/Command/GenerateProgrammaticPipelineFromConfig/Help.php
```diff
@@ -0,0 +1,71 @@
...
+
+    /**
+     * @var ConsoleHelper
+     */
+    private $helper;
+
+    /**
+     * @param ConsoleHelper $helper
+     */
+    public function __construct($command, ConsoleHelper $helper)
```
webimpress: Missing `$command` in PHPDoc.

src/Command/GenerateProgrammaticPipelineFromConfig/Help.php
```diff
@@ -0,0 +1,71 @@
+<?php
+/**
+ * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
+ * @copyright Copyright (c) 2015-2016 Zend Technologies USA Inc. (http://www.zend.com)
```
webimpress: This is general one: shouldn't we just use `2016` year for new files?

src/Container/NotFoundHandlerFactory.php
```diff
@@ -0,0 +1,37 @@
...
+
+use Interop\Container\ContainerInterface;
+use Zend\Diactoros\Response;
+use Zend\Expressive\Middleware\NotFoundHandler;
+use Zend\Expressive\Template\TemplateRendererInterface;
+
+class NotFoundHandlerFactory
+{
+    /**
+     * @param ContainerInterface
```
webimpress: Missing param name `$container`

src/Middleware/WhoopsErrorResponseGenerator.php
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        $this->whoops = $whoops;
+    }
+
+    /**
+     * @param \Throwable|\Exception $e
+     * @param ServerRequestInterface $request
+     * @param ResponseInterface $response
+     * @return $response
```
webimpress: Should be here return __type__, not a variable name.

src/Container/ApplicationFactory.php
```diff
@@ -7,16 +7,16 @@
 
 namespace Zend\Expressive\Container;
 
+use ArrayObject;
 use Interop\Container\ContainerInterface;
-use SplPriorityQueue;
 use Zend\Diactoros\Response\EmitterInterface;
 use Zend\Expressive\Application;
+use Zend\Expressive\ApplicationUtils;
```
webimpress: I can't see this class in the library. Where is it?

src/Container/ApplicationFactory.php
```diff
@@ -27,105 +27,41 @@
...
+ *   creating an empty instance if none is found.
  * - 'Zend\Diactoros\Response\EmitterInterface'. If missing, an EmitterStack is
  *   created, adding a SapiEmitter to the bottom of the stack.
  * - 'config' (an array or ArrayAccess object). If present, and it contains route
  *   definitions, these will be used to seed routes in the Application instance
  *   before returning it.
  *
- * When introspecting the `config` service, the following structure can be used
- * to define routes:
+ * Please see the `Zend\Expressive\ApplicationUtils` class for details on how
```
webimpress: As above, I can't see this class in the library. Am I missing something?

src/Command/GenerateProgrammaticPipelineFromConfig/Generator.php
```diff
@@ -0,0 +1,458 @@
...
+     *
+     * If all other cases, it is returned as the var_export value.
+     *
+     * @param mixed $value
+     * @param int $indentLevel
+     * @return string
+     */
+    private function createOptionValue($value, $indentLevel = 1)
+    {
+        if (is_array($value) || $value instanceof Traversable) {
```
webimpress: `Traversable` should be imported or `\Traversable` here.

src/Middleware/WhoopsErrorResponseGenerator.php
```diff
@@ -0,0 +1,88 @@
...
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Whoops\Handler\PrettyPageHandler;
+use Whoops\Run;
+use Whoops\RunInterface;
+
+class WhoopsErrorResponseGenerator
+{
+    /**
+     * @var Whoops
```
webimpress: Seems to be wrong type here.

src/Command/GenerateProgrammaticPipelineFromConfig/Generator.php
```diff
@@ -0,0 +1,458 @@
...
+
+        return sprintf("[\n    %s,\n]", implode(",\n    ", $middleware));
+    }
+
+    /**
+     * @param array $options
+     * @param int $indentLevel
+     * @return string
+     */
+    private function formatOptions($options, $indentLevel = 1)
```
webimpress: Is `$options` arg always an array? Can we add type hint here?

doc/book/reference/migration/to-v1.1.md
```diff
@@ -0,0 +1,560 @@
...
+  that middleware re-throw for exceptions it cannot handle.
+
+- Consider providing your own `Zend\Stratigility\NoopFinalHandler`
+  implementation; this will now only be invoked if the queue is exhausted, and
+  could return a generic 404 page, raise an exception, etc.
+
+## http-interop
+
+Stratigility 1.3 provides the ability to work with [http-interop middleware
+0.2.0](https://github.com/http-interop/http-middleware/tree/ff545c87e97bf4d88f0cb7eb3e89f99aaa53d7a9).
```
webimpress: Maybe better here link to tagged version:
https://github.com/http-interop/http-middleware/tree/0.2.0

webimpress:
In `Zend\Expressive\Application` there are some methods and properties marked as:
```
@deprecated This property will be removed in v1.1.
```

and also:

```
@todo For 1.1, remove the RouteResultSubjectInterface implementation, and
      all deprecated properties and methods.
```

Is it gonna be done in this PR, or PR https://github.com/zendframework/zend-expressive/pull/376 will be merged before release 1.1?

webimpress:
Migration script:
I would expect that after using migration script my application will work straight away, but I've got error:

```
Fatal error: Uncaught exception 'Zend\Expressive\Exception\InvalidMiddlewareException' with message 'Unable to resolve middleware "string[Scalar]" to a callable' in my-app\vendor\zendframework\zend-expressive\src\MarshalMiddlewareTrait.php on line 65
```

This is because migration script is not adding:
```
    'dependencies' => [
        'invokables' => [
            OriginalMessages::class => OriginalMessages::class,
```

Into configuration, but:

```
$app->pipe(\Zend\Stratigility\Middleware\OriginalMessages::class);
```
is added into `config/pipeline.php`.

I think it should be done automatically (in migration script) or it should be noted in migration notes, that this action is required to be done manually.

webimpress:
One more issue. This was my config based configuration - `config/autoload/middleware-pipeline.global.php`:
```php
[...]
    'middleware_pipeline' => 
        'my-path' => [
            'path' => '/my-path',
            'middleware' => [
                ApplicationFactory::ROUTING_MIDDLEWARE,
                \App\Middleware\MyMiddleware::class,
                \App\Middleware\OtherMiddleware::class,
            ],
        ],
```

it was successfully converted to middleware pipeline (via migration script) - `config/pipeline.php`:

```php
[...]
$app->pipe('/my-path', [
    \Zend\Expressive\Application::ROUTING_MIDDLEWARE,
    \App\Middleware\MyMiddleware::class,
    \App\Middleware\OtherMiddleware::class,
]);
```

And then piping is failing, because `pipe` method from Application call `prepareMiddleware` method from `Zend\Expressive\MarshalMiddlewareTrait` with `$middleware = "EXPRESSIVE_ROUTING_MIDDLEWARE"`, and we got fatal:
```
Fatal error: Uncaught exception 'Zend\Expressive\Exception\InvalidMiddlewareException' with message 'Unable to resolve middleware "string[Scalar]" to a callable' in my-app\vendor\zendframework\zend-expressive\src\MarshalMiddlewareTrait.php on line 65
```

Am I doing something wrong or there are missing some scenarios, when we have some routing/dispatch middlewares only for some paths?

webimpress:
Another issue. "Method not allowed". I have routing, let say:
```php
$app->post('/foo', \App\Action\Foo::class, 'foo');
```

So only POST request are allowed. If I try to do `GET /foo` I'm getting 200 response with no body.
I should get 405.

I think the problem is in `\Zend\Expressive\Application::routeMiddleware:452` where we have:
```php
        $result = $this->router->match($request);

        if ($result->isFailure()) {
            if ($result->isMethodFailure()) {
                $response = $response->withStatus(405)
                    ->withHeader('Allow', implode(',', $result->getAllowedMethods()));
                return $next($request, $response, 405); // <-------
            }
[...]
```

and then, the 3rd argument (`405`) is passed through `\Zend\Stratigility\Next:119` where `$err = 405`. Error is not thrown (because `$err` is 405 not an instance of `\Exception|\Throwable`).
I can write here more, but I think it should be enough to investigate and fix this issue.

doc/book/reference/migration/to-v1.1.md
weierophinney: Yeah, probably. Thanks for the suggestion; will implement shortly.

test/Command/GenerateProgrammaticPipelineFromConfig/GeneratorTest.php
weierophinney: Odd, I needed to have this in another test suite whenever I used the `vfsStream::setup()` method; evidently that's no longer true in recent versions. Removing.

src/Command/GenerateProgrammaticPipelineFromConfig/Generator.php
weierophinney: It could be an `ArrayObject`.

weierophinney:
> Is it gonna be done in this PR, or PR #376 will be merged before release 1.1?

I'll be merging #376 before release, which covers these.

src/Command/GenerateProgrammaticPipelineFromConfig/Generator.php
webimpress: So then PHPDoc should be updated.

weierophinney:
> Another issue. "Method not allowed". I

Should be resolved now; thanks for finding the issue!

weierophinney:
> And then piping is failing, because pipe method from Application call prepareMiddleware method from Zend\Expressive\MarshalMiddlewareTrait with $middleware = "EXPRESSIVE_ROUTING_MIDDLEWARE", and we got fatal

I think this is an edge case we never tested.

When you use configuration-based pipelines, as it loops through nested pipelines, if it encounters either of those special constants, it replaces the middleware with the appropriate callable (`[$this, 'routeMiddleware]` and `[$this, 'dispatchMiddleware]`).

However, in the `MarshalMiddlewareTrait`, we do not.

I'm going to see if I can create a test case and resolve this; I think it should be relatively easy to accomplish.

src/Command/GenerateProgrammaticPipelineFromConfig/Generator.php
```diff
@@ -0,0 +1,458 @@
...
+    ],
+];
+
+EOT;
+
+    const TEMPLATE_PIPELINE = <<< 'EOT'
+<?php
+/**
+ * Expressive middleware pipeline
+ */
```
weierophinney: Done!

src/Command/GenerateProgrammaticPipelineFromConfig/Generator.php
weierophinney: Done!

src/Container/ErrorHandlerFactory.php
```diff
@@ -0,0 +1,29 @@
...
+     * @param ContainerInterface $container
+     * @return ErrorHandler
+     */
+    public function __invoke(ContainerInterface $container)
+    {
+        $generator = $container->has(ErrorResponseGenerator::class)
+            ? $container->get(ErrorResponseGenerator::class)
+            : null;
+
+        return new ErrorHandler(new Response(), $generator);
```
danizord: @weierophinney any chance to make it possible to configure `ErrorHandler` listeners via config? Or should I override this factory to attach my custom listeners?

xtreamwayz:
@weierophinney I only see the config setting for the new error handling in the migration guide:
`'zend-expressive' => ['raise_throwables' => true,],`

Did you forgot `$app->raiseThrowables();` or is the config setting preferred?

doc/book/reference/migration/to-v1.1.md
```diff
@@ -0,0 +1,604 @@
...
+- Original request and response messages
+- Error handling
+- Programmatic middleware pipelines
+- Usage of [http-interop middleware](https://github.com/http-interop/http-middleware)
+
+## Original messages
+
+Stratigility 1.3 deprecates its internal request and response decorators,
+`Zend\Stratigility\Http\Request` and `Zend\Stratigility\Http\Response`,
+respsectively. The main utility of these instances was to provide access in
```
xtreamwayz: Typo: respectively

xtreamwayz:
Also phpunit can be upgraded to 5.x since PHP 5.5 support is removed.

weierophinney:
> Did you forgot $app->raiseThrowables(); or is the config setting preferred?

You can do either; I didn't document the method, as technically it's part of the Stratigility 1.3 API, and documented there.

doc/book/reference/migration/to-v1.1.md
```diff
@@ -0,0 +1,604 @@
...
+```php
+$app->get('/', \App\Action\HomePageAction::class, 'home');
+$app->get('/api.ping', \App\Action\PingAction::class, 'api.ping');
+```
+
+We recommend rewriting your middleware pipeline and routing configuration into
+programmatic/declarative statements. Specifically:
+
+- We recommend putting the pipeline declarations into `config/pipeline.php`.
+- We recommend putting the pipeline declarations into `config/routes.php`.
```
weierophinney: Fixed, thanks!

doc/book/reference/migration/to-v1.1.md
weierophinney: Fixed, thanks!

weierophinney:
> Also phpunit can be upgraded to 5.x since PHP 5.5 support is removed.

Done in latest commit; thanks!

weierophinney:
@webimpress On Friday, I resolved your issues with using the `ROUTING_MIDDLEWARE`/`DISPATCH_MIDDLEWARE` inside arrays of middleware when defining a programmatic pipeline; please test again!

bin/expressive-migrate-original-messages
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env php
+<?php // @codingStandardsIgnoreFile
+/**
+ * Script for migrating configuration-driven pipelines/routes to programmatic usage.
+ *
+ * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
+ * @copyright Copyright (c) 2015-2016 Zend Technologies USA Inc. (http://www.zend.com)
```
webimpress: Should be just `2016` in new files.

bin/expressive-pipeline-from-config
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env php
+<?php // @codingStandardsIgnoreFile
+/**
+ * Script for migrating configuration-driven pipelines/routes to programmatic usage.
+ *
+ * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
+ * @copyright Copyright (c) 2015-2016 Zend Technologies USA Inc. (http://www.zend.com)
```
webimpress: As above, just `2016`.

src/Command/MigrateOriginalMessageCalls/Help.php
```diff
@@ -0,0 +1,82 @@
+<?php
+/**
+ * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
+ * @copyright Copyright (c) 2015-2016 Zend Technologies USA Inc. (http://www.zend.com)
```
webimpress: Only `2016` in new files, please.

test/Command/MigrateOriginalMessageCalls/HelpTest.php
```diff
@@ -0,0 +1,39 @@
+<?php
+/**
+ * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
+ * @copyright Copyright (c) 2015-2016 Zend Technologies USA Inc. (http://www.zend.com)
```
webimpress: Only `2016` in new files, please.

test/ApplicationConfigInjectionTest.php
```diff
@@ -0,0 +1,413 @@
...
+     * @dataProvider callableMiddlewares
+     */
+    public function testFactorySetsUpRoutesFromConfig($middleware)
+    {
+        $config = [
+            'routes' => [
+                [
+                    'path' => '/',
+                    'middleware' => $middleware,
+                    'allowed_methods' => [ 'GET' ],
```
webimpress: We shouldn't have spaces after opening array brackets and before closing:
`['GET']` instead of `[ 'GET' ]`

More places in this file.

test/ApplicationConfigInjectionTest.php
```diff
@@ -0,0 +1,413 @@
...
+
+    public function testMiddlewareWithoutPriorityIsGivenDefaultPriorityAndRegisteredInOrderReceived()
+    {
+        // @codingStandardsIgnoreStart
+        $middleware = function ($request, $response, $next) {};
+        // @codingStandardsIgnoreEnd
+
+        $pipeline1 = [ [ 'middleware' => clone $middleware ] ];
+        $pipeline2 = [ [ 'middleware' => clone $middleware ] ];
+        $pipeline3 = [ [ 'middleware' => clone $middleware ] ];
```
webimpress: Again, no spaces after opening array brackets, and before closing.

doc/book/reference/migration/to-v1.1.md
```diff
@@ -0,0 +1,623 @@
...
+  that middleware re-throw for exceptions it cannot handle.
+
+- Consider providing your own `Zend\Stratigility\NoopFinalHandler`
+  implementation; this will now only be invoked if the queue is exhausted, and
+  could return a generic 404 page, raise an exception, etc.
+
+## http-interop
+
+Stratigility 1.3 provides the ability to work with [http-interop middleware
+0.2.0](https://github.com/http-interop/http-middleware/tree/0.2.0).
```
webimpress: Please note that 0.2.1 is released with REMOVED `MiddlewareInterface`, there is still `ServerMiddlewareInterface`, but I think this (remove) shouldn't be done in 0.2.X release...

src/Command/MigrateOriginalMessageCalls/ConvertOriginalMessageCalls.php
```diff
@@ -0,0 +1,150 @@
+<?php
+/**
+ * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
+ * @copyright Copyright (c) 2016-2016 Zend Technologies USA Inc. (http://www.zend.com)
```
webimpress: Please fix year. Should be only `2016`, the same in some other new files (12 times)

src/Command/MigrateOriginalMessageCalls/ConvertOriginalMessageCalls.php
```diff
@@ -0,0 +1,150 @@
...
+     */
+    public function __construct(ConsoleHelper $console)
+    {
+        $this->console = $console;
+    }
+
+    /**
+     * Process a directory
+     */
+    public function process($directory)
```
webimpress: Missing `$directory` param in PHPDoc.

src/Command/MigrateOriginalMessageCalls/Help.php
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * @var ConsoleHelper
+     */
+    private $helper;
+
+    /**
+     * @param ConsoleHelper $helper
+     */
+    public function __construct($command, ConsoleHelper $helper)
```
webimpress: Missing param `$command` in PHPDoc.

src/Container/ApplicationFactory.php
```diff
@@ -7,16 +7,15 @@
 
 namespace Zend\Expressive\Container;
 
+use ArrayObject;
 use Interop\Container\ContainerInterface;
-use SplPriorityQueue;
 use Zend\Diactoros\Response\EmitterInterface;
 use Zend\Expressive\Application;
 use Zend\Expressive\Exception;
```
webimpress: `Zend\Expressive\Exception` seems to be no longer used and can be removed.

src/Container/NotFoundHandlerFactory.php
webimpress: Please add param name `$container` in PHPDoc after type `ContainerInterface`.

src/Container/WhoopsErrorResponseGeneratorFactory.php
```diff
@@ -0,0 +1,25 @@
...
+class WhoopsErrorResponseGeneratorFactory
+{
+    /**
+     * @param ContainerInterface $container
+     * @return WhoopsErrorResponseGenerator
+     */
+    public function __invoke(ContainerInterface $container)
+    {
+        return new WhoopsErrorResponseGenerator(
+            $container->get('Zend\Expressive\Whoops')
```
webimpress: Should we use `::class` here?

bin/expressive-scan-for-error-middleware
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env php
+<?php // @codingStandardsIgnoreFile
+/**
+ * Script for migrating configuration-driven pipelines/routes to programmatic usage.
+ *
+ * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
+ * @copyright Copyright (c) 2015-2016 Zend Technologies USA Inc. (http://www.zend.com)
```
webimpress: As it is a new file, there should be just `2016` year.

src/Command/ScanForErrorMiddleware/ErrorMiddlewareFilter.php
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        return array_reduce(
+            $classes,
+            $this->scanClass($file),
+            false
+        );
+    }
+
+    private function scanClass(SplFileInfo $file)
```
webimpress: Missing PHPDoc.

src/Command/ScanForErrorMiddleware/Scanner.php
```diff
@@ -0,0 +1,129 @@
...
+
+    /**
+     * @var string Path to scan.
+     */
+    private $path;
+
+    /**
+     * @param string path
+     * @param ConsoleHelper $console
+     * @throws ScannerException if the path is not a directory or not readable.
```
webimpress: There is no any throws in construct.

composer.json
```diff
@@ -27,14 +27,15 @@
     "require-dev": {
         "filp/whoops": "^1.1 || ^2.0",
         "malukenho/docheader": "^0.1.5",
+        "mikey179/vfsStream": "^1.6",
         "mockery/mockery": "^0.9.5",
         "phpunit/phpunit": "^5.0",
+        "zendframework/zend-code": "^3.1",
```
webimpress: New script "expressive-scan-for-error-middleware" requires `zend-code` to work, but this is only in `require-dev` section. So to use the script in project it's required to install `zend-code`. I think it should be noted in documentation, because this tool is not gonna work out of the box.

src/Command/ScanForErrorMiddleware/Scanner.php
```diff
@@ -0,0 +1,129 @@
...
+     */
+    private $count = 0;
+
+    /**
+     * @var string Path to scan.
+     */
+    private $path;
+
+    /**
+     * @param string path
```
webimpress: Missing `$` before variable name (`path`).

src/Command/ScanForErrorMiddleware/Scanner.php
```diff
@@ -0,0 +1,129 @@
...
+    {
+        $rdi = new RecursiveDirectoryIterator($this->path);
+        $rii = new RecursiveIteratorIterator($rdi);
+        return new ErrorMiddlewareFilter($rii);
+    }
+
+    /**
+     * Recursively scan the given path for error middleware.
+     *
+     * @return void
```
webimpress: It will be nice to keep consistency. For some methods `return void` is missing.

src/Command/ScanForErrorMiddleware/ErrorMiddlewareFilter.php
```diff
@@ -0,0 +1,123 @@
...
+        return 4 === $method->getNumberOfParameters();
+    }
+
+    /**
+     * @param ClassScanner $class
+     * @return bool
+     */
+    private function callsNextWithError(ClassScanner $class)
+    {
+        return array_reduce($class->getMethods(), function ($found, $method) {
```
webimpress: Can we add PHPDoc for params `$found` and `$methods`?

src/Command/ScanForErrorMiddleware/ErrorMiddlewareFilter.php
```diff
@@ -0,0 +1,123 @@
...
+        );
+    }
+
+    private function scanClass(SplFileInfo $file)
+    {
+        $file->implementsInterface      = false;
+        $file->invokableErrorMiddleware = false;
+        $file->callsNextWithError       = false;
+
+        return function ($found, $class) use ($file) {
```
webimpress: Can we add PHPDoc for params `$found` and `$class`?

src/Command/ScanForErrorMiddleware/NextInvocationScanner.php
```diff
@@ -0,0 +1,120 @@
...
+ * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
+ * @copyright Copyright (c) 2016 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   https://github.com/zendframework/zend-expressive/blob/master/LICENSE.md New BSD License
+ */
+
+namespace Zend\Expressive\Command\ScanForErrorMiddleware;
+
+class NextInvocationScanner
+{
+    const TEMPLATE = <<< 'EOT'
```
webimpress: I've just checked why single quotes here:
http://php.net/manual/en/language.types.string.php#language.types.string.syntax.nowdoc
I didn't know about it. Nice 👍

src/Command/ScanForErrorMiddleware/NextInvocationScanner.php
```diff
@@ -0,0 +1,120 @@
...
+            }
+
+            if (is_array($token)) {
+                // Hit a token; continue to next
+                continue;
+            }
+
+            if ($token === '(') {
+                // Hit an opening paren; increase paren depth
+                $parenDepth += 1;
```
webimpress: Maybe just `++`?

src/Command/ScanForErrorMiddleware/NextInvocationScanner.php
```diff
@@ -0,0 +1,120 @@
...
+                continue;
+            }
+
+            if ($token === ')') {
+                // Hit an closing paren; decrease paren depth
+                $parenDepth -= 1;
```
webimpress: And here `--`?

test/Container/ErrorResponseGeneratorFactoryTest.php
weierophinney: It is - but using the `::class` notation, for `assertInstanceOf` checks.

src/Container/WhoopsErrorResponseGeneratorFactory.php
weierophinney: No, because it's a virtual (vs defined) class name, and resolves to a `Whoops\Run` instance. Having it as a string helps signal that fact.

src/Command/ScanForErrorMiddleware/Scanner.php
weierophinney: Fixed.

src/Command/ScanForErrorMiddleware/NextInvocationScanner.php
weierophinney: I've been using `+=` and `-=` for a few years, since seeing a Douglas Crockford talk where he demonstrated how it's more optimized both for the language interpreter (while he was talking about JavaScript, the optimizations are present for most C-style languages), as well as human comprehension (as `++` and `--` vary behavior based on whether or not they prefix or suffix the variable). Using the `+=` and `-=` syntax makes it clear that we are _re-assigning_ the value following the increment/decrement operation.

It essentially comes down to author preference; I've not been changing them during code reviews unless doing so aids in understanding how the code works.

weierophinney:
I've just tried running each of the scripts against an older version of my own website to see how the migration would work.

The migration script for converting to a programmatic pipeline _just worked_, and my application was able to work immediately, with a few exceptions (a few deprecation notices, and places where I was using the legacy error handling).

The script for updating calls to `getOriginal*()` worked, fixing several `getOriginalRequest()` and `getOriginalUri()` calls. It also flagged one `getOriginalResponse()` call, which I was able to update to pull out of the request attribute. Again, after these changes, everything continued to _just work_.

Finally, I ran the script for identifying error middleware and/or invocation. I had a few false positives, primarily in classes implementing `__invoke()` with more than 3 arguments (delegator factories, in particular!), but it otherwise found them. This particular step is more difficult, as anything flagged needs to be manually updated. My workflow was:

- Open file, locate invocation(s).
- Determine if this is a generic error, a 404, or an app-specific error.
  - For generic errors, I simply threw a `RuntimeException`, including a status code if one was originaly given.
  - For 404 errors, I updated the middleware to compose the `NotFoundHandler`, and returned the invocation of that (see more below).
  - I had one app-specific error, for authorization. For this, I did two thingS:
    - Created a custom exception
    - Rewrote my `Unauthorized` error middleware to be standard middleware which wrapped a call to `$next` in a try/catch block, returning a response if one was returned. If the new exception was caught, I continued with my original middleware logic; for any other exception, I re-threw, so the standard error handler would catch it.

As noted above, I discovered a new use case for the `NotFoundHandler`: injecting it in middleware that may need to return a 404 response! Previously, I had several middleware that invoked `$next` with a `404` error argument, which my own final handler was picking up; now, I have these same middleware compose a `ServerMiddlewareInterface` instance that is intended to generate and return a 404 response; my factories then inject the `NotFoundHandler` for this argument.

I'm going to also try this on the example application @ezimuel used for a recent workshop, but I feel that my own website was sufficiently complex enough to demonstrate that the migration tools definitely assist in updating an application to switch to programmatic pipelines, switch to the new error handling, and help developers update their apps to remove deprecated functionality.

src/Container/ApplicationFactory.php
```diff
@@ -136,352 +70,75 @@ class ApplicationFactory
...
+
         $router = $container->has(RouterInterface::class)
             ? $container->get(RouterInterface::class)
             : new FastRouteRouter();
 
-        $finalHandler = $container->has('Zend\Expressive\FinalHandler')
-            ? $container->get('Zend\Expressive\FinalHandler')
-            : $this->marshalFinalHandler($container);
+        $finalHandler = (isset($config['zend-expressive']['raise_throwables'])
+            && $config['zend-expressive']['raise_throwables'])
```
webimpress: `! empty` should be enough here

src/Container/ApplicationFactory.php
```diff
@@ -136,352 +70,75 @@ class ApplicationFactory
...
-        if (isset($pipeline['pre_routing'])) {
-            if (! is_array($pipeline['pre_routing'])) {
-                throw new ContainerInvalidArgumentException(sprintf(
-                    'Pre-routing middleware collection must be an array; received "%s"',
-                    gettype($pipeline['pre_routing'])
-                ));
-            }
-            $this->injectMiddleware($pipeline['pre_routing'], $app);
+        if (isset($config['zend-expressive']['raise_throwables'])
+            && $config['zend-expressive']['raise_throwables']
```
webimpress: As above - `! empty`

src/Container/ApplicationFactory.php
```diff
@@ -136,352 +70,75 @@ class ApplicationFactory
...
-        if (isset($pipeline['post_routing'])) {
-            if (! is_array($pipeline['post_routing'])) {
-                throw new ContainerInvalidArgumentException(sprintf(
-                    'Post-routing middleware collection must be an array; received "%s"',
-                    gettype($pipeline['post_routing'])
-                ));
-            }
-            $this->injectMiddleware($pipeline['post_routing'], $app);
+        if (! isset($config['zend-expressive']['programmatic_pipeline'])
+            || ! $config['zend-expressive']['programmatic_pipeline']
```
webimpress: `empty` here

test/Command/GenerateProgrammaticPipelineFromConfig/GeneratorTest.php
```diff
@@ -0,0 +1,156 @@
...
+        [
+            'middleware' => %s,
+        ]
+    ],
+];
+EOT;
+        return [
+            'int'        => [sprintf($pipelineConfigTemplate, '1')],
+            'float'      => [sprintf($pipelineConfigTemplate, '1.1')],
+            'object'     => [sprintf($pipelineConfigTemplate, '(object) [\'foo\' => \'bar\']')],
```
webimpress: Can we have less spaces here before `=>`, please?

test/Command/GenerateProgrammaticPipelineFromConfig/GeneratorTest.php
```diff
@@ -0,0 +1,156 @@
...
+
+    public function invalidMiddleware()
+    {
+        $pipelineConfigTemplate = <<< 'EOT'
+<?php
+return [
+    'middleware_pipeline' => [
+        [
+            'middleware' => %s,
+        ]
```
webimpress: Would be nice to have also here trailing comma.

src/Command/MigrateOriginalMessageCalls/ConvertOriginalMessageCalls.php
```diff
@@ -0,0 +1,153 @@
...
+     */
+    public function originalResponseFound()
+    {
+        return $this->originalResponseCount > 0;
+    }
+
+    /**
+     * Process a single file
+     *
+     * @param $filename
```
webimpress: Missing parameter type.

src/Middleware/ErrorResponseGenerator.php
```diff
@@ -0,0 +1,145 @@
...
+
+    /**
+     * @var string
+     */
+    private $template;
+
+    /**
+     * @var bool $isDevelopmentMode
+     * @var null|TemplateRendererInterface $renderer
+     * @var string $template
```
webimpress: `@param` instead of `@var` please

webimpress:
I've tested it on my application and here are my notes:

- [ ] as I've noted [before](https://github.com/zendframework/zend-expressive/pull/396#discussion_r89659704) to use "expressive-scan-for-error-middleware" script requires `zend-code` to work. It is required but in __dev__ section of composer.json and I'm not using it in my project, so I have got fatal:
  ```
  PHP Fatal error:  Class 'Zend\Code\Scanner\FileScanner' not found in C:\path-to-my-project\vendor\zendframework\zend-expressive\src\Command\ScanForErrorMiddleware\ErrorMiddlewareFilter.php on line 35
  ```
- [ ] Scripts on Windows show in help absolute path in Usage, for example:
  ```bash
  $ vendor/bin/expressive-migrate-original-messages
  Usage:

    C:/path-to-my-project/vendor/zendframework/zend-expressive/bin/expressive-migrate-original-messages [command] [options]

  Commands:
  ...
  ```
  Would be nice to have there just `vendor/bin/expressive-migrate-original-messages` instead. I don't know how it is on linux/osx.
- [ ] Consistent messages in scripts. Here we are current results (when nothing is changed):
  ```bash
  $ vendor/bin/expressive-scan-for-error-middleware scan
  Scanning for error middleware or error middleware invocation...
  Done!
  ```
  ```bash
  $ vendor/bin/expressive-migrate-original-messages scan
  // <!---------- MISSING short description here
  Done!
  ```
  If nothing is changed we see message "Done!", and I'm not sure if it is clear enough that nothing has changed. Not important, but it could improve user experience.
- [ ] Should be allowed to run "expressive-pipeline-from-config" script just once. I mean the next run should be blocked or clever detected. Files `config/pipeline.php` and `config/routes.php` can be already there, and maybe we shouldn't force override them. Also when we run it again we have got duplicated entries in `public/index.php`: `error_reporting(error_reporting() & ~E_USER_DEPRECATED);` and `require 'config/pipeline.php'; require 'config/routes.php';`, then we have got fatal:
  ```
  Fatal error: Uncaught exception 'Zend\Expressive\Exception\DuplicateRouteException' with message 'Duplicate route detected; same name or path, and one or more HTTP methods intersect'
  ```
   I know, the script is designed to run just ONCE, so maybe it's not important.
- [ ] It seems we don't need `Zend\Stratigility\Middleware\OriginalMessages` middleware right now. The functionality it delivers is duplicated, because:
  - [`Application:run()` inject `originalResponse`](https://github.com/weierophinney/zend-expressive/blob/feature/programmatic/src/Application.php#L610),
  - [`MiddlewarePipe::__invoke()` decorate request](https://github.com/zendframework/zend-stratigility/blob/master/src/MiddlewarePipe.php#L91) and [inject `originalUri` and `originalRequest` attributes](https://github.com/zendframework/zend-stratigility/blob/master/src/MiddlewarePipe.php#L254-L261) - so request is _always_ instance of `Zend\Stratigility\Http\Request` (deprecated) and request has always `original*` attributes (without using `OriginalMessages` middleware). So right now, with Stratigility 1.3, we don't need to add `OriginalMessages` middleware into our pipeline.


I tried to migrate my application using notes from [previous @weierophinney comment](https://github.com/zendframework/zend-expressive/pull/396#issuecomment-264991054) and I'm a bit confused. I'd like to use single-pass middlewares. My application checks if user is authorized at the beginning (routing middleware) - based on headers/request. This is the only place where app checks for permissions. If user is not authorized should I:
  - break pipeline and just `return new Response()->withStatusCode(403)`?
  - throw exception and catch it later and return 403?
  - pass `UnauthorizedHandler` to the middleware and process it (the same approach as suggested for `NotFoundHandler` in above mwop comment)?

If user is authorized my application returns usually `204` (no content), `200`/`201` (json responses - create/update),  `404` (not found) - it's kinda REST api. How then in single-pass approach I should return these responses:
  - pass handlers for each type of response (`NoContentHandler`, `JsonHandler`, `NotFoundHandler`...)?
  - create own responses `NoContentResponse`, `JsonResponse`, `NotFoundResponse` and just return it (break pipeline)?
  - anything else ... ?

I've got confused with single-pass approach, maybe I'm doing something wrong, or not understand it correctly. I hope it could be clarified here.

weierophinney:
> "expressive-scan-for-error-middleware" script requires zend-code to work.

I noticed this yesterday, but originally thought it was a fluke, so thanks for confirming! I'd forgotten that Composer no longer installs dev dependencies of dependencies when installing (it did in early versions, and we got bit by this change in other components later).

I'm not quite sure what to do about this. Expressive itself does not _require_ zend-code; only this particular script does. And I'm a bit loathe to add a _required_ dependency for something that is only to be used during _development_.

One possibility would be to create a separate package with these migration tools which users can then install as a development dependency in their projects. The problem I see with that is it's an additional step: they have to (a) read the migration documentation, (b) install the tooling, and then (c) use it, versus just (a) and (c) as we currently have. The benefit, though, is that they could then remove it when done. Additionally, they could install it globally, and run it on any project.

The only command that has any dependencies on Expressive classes is `Zend\Expressive\Command\GenerateProgrammaticPipelineFromConfig\Generator`, and this is due to constants it references. These could easily be updated to inline those values locally, which would allow the tooling to exist with only dependencies on Stratigility (for identifying ErrorMiddlewareInterface instances), zend-code, and either zend-stdlib or zf-console or symfony/console for console tooling. As a development dependency, none of these would then be installed for production deployments, which would also ensure there is no vector for abuse on production systems.

What do you think about this approach, @webimpress? /cc @zendframework/community-review-team @michaelmoussa @xtreamwayz

webimpress:
> One possibility would be to create a separate package with these migration tools

@weierophinney I kinda like this approach, because these tools are actually "development tools" only and we don't want them to be available in live/production environment. As you mention it will add a bit more work on migration, but as you suggest, will be possible to use it as global tools (in case someone needs migrate more projects). In expressive library we can have then entries in "suggest" and/or "conflicts" of composer.json (for example we will allow only `expressive-migration-tools ^2.0` to work with `expressive ^2.0` etc).

The more I think about it, it just seems to be right solution :)

weierophinney:
> Scripts on Windows show in help absolute path in Usage

In your example, I notice an additional detail: it's not the path to the symlink under `vendor/bin/`, but to the script under the package itself; it's resolving the link. (Under *nix systems, it displays `./vendor/bin/{command name}`; in other words, it's exactly what you provide on the CLI, no more, no less.)

My thought on this, then, is to do the following: if the command is a resolvable filename, and it is identical to the _realpath_ of that filename, we truncate it to only the filename. The user already has invoked it, so the full path is not necessary. If it is not a realpath, we leave it as is.

I'll make this change shortly.

weierophinney:
> It seems we don't need Zend\Stratigility\Middleware\OriginalMessages middleware right now.

For now, yes, but when Stratigility 2.0 is ready, it will no longer perform these injections. Having this middleware in place makes an application _future-proof_.

webimpress:
Maybe it's too crazy but, what if we extract migrations tools into separate library and prepare here simply _interactive migration script_, which will install composer dependencies (migrations-tool library, if needed) and display available migrations options (we can detect current version, check what is to do, etc...):
```bash
$ ./vendor/bin/expressive-migration
Installing required dependencies....

Please choose what you want to do:
  [1] migrate to pipeline from config
  [2] migrate original messages
  [3] scan for error middlewares
  [Q] quit from migration tool

Your choose ___
```

@weierophinney What do you think?
 interactive migration tool.

weierophinney:
@webimpress —

Regarding:

> I'd like to use single-pass middlewares. My application checks if user is authorized at the beginning (routing middleware) - based on headers/request. This is the only place where app checks for permissions. If user is not authorized should I:

and:

> How then in single-pass approach I should return these responses:

There are a few approaches to this, and you actually touch on all of them.

The reason single pass is being recommended as part of PSR-15/http-middleware is due to a pretty big issue with passing the response: if one middleware alters the response it pushes into another layer (by invoking `$next` or the delegate), there's no guarantee that those changes will be reflected in the response returned by inner layers as any of these layers could potentially create and return a brand new response. Additionally, because any given layer could write to the response body, developers need to jump a number of hurdles to ensure anything *they* write properly overwrites the original (see [Rob Allen's post on this subject](https://akrabat.com/filtering-the-psr-7-body-in-middleware/).  Anthony Ferrara has written a [comprehensive post covering these response issues](http://blog.ircmaxell.com/2016/05/all-about-middleware.html) if you want more detail.

So, knowing that passing the response via the `$next`/`$delegate` argument is not safe, how do we return a response?

You've noted several options already, and these are the suggested approaches:

- Inject an empty or pre-populated response via the middleware's constructor.  This approach allows you to typehint against PSR-7, and then use your DiC to provide the concrete implementation.

- Inject a _factory_ that is capable of producing a response via the middleware's constructor. (This is essentially the idea behind [PSR-17](https://github.com/php-fig/fig-standards/tree/master/proposed/http-factory) ([HTTP message factories](https://github.com/http-interop/http-factory))). This approach allows you to typehint against a callable or a generic interface, and, again, use your DiC to provide the concrete implementation. Such a factory could even be _other middleware_, as I noted in a previous comment with regards to re-using the `NotFoundHandler` from Expressive for generating my 404 response.

- Instantiate a concrete response within your middleware. In this case, your middleware would _depend_ on a concrete implementation of PSR-7. This is perfectly acceptable _within your application_, but for middleware you want to distribute as a library, you should likely go with the previous approaches.

I personally like the second/middle approach above a ton, as it allows me to create templated responses, or even alter responses based on `Accept` header and the like.

Now, regarding your "user not authorized" scenario, you have a couple of options:

- As noted already, you could return a response somehow, using one of the above methods. This is likely the fastest and easiest possiblity, and it's what I'm doing on the current incarnation of my website. (See [this example](https://github.com/weierophinney/mwop.net/blob/db4d5a7c1c2d2524eb7c4713f287e75e849ae5da/src/ComicsPage.php#L22-L33).) This approach is essentially the first and last bullet points you presented.

- Create a custom middleware for handling "unauthorized" exceptions, and have other middleware raise such an exception for those situations. That ends up looking like:

  ```php
  class UnauthorizedHandler
  {
      public function __invoke($request, DelegateInterface $delegate)
      {
          try {
              $response = $delegate->process($request);
              return $response;
          } catch (UnauthorizedException $e)
          } catch (\Throwable $e)
              throw $e;
          } catch (\Exception $e)
              throw $e;
          }

          // create 403 or related response and return it here...
      }
  }

  class MiddlewareRequiringAuthorization
  {
      public function __invoke($request, DelegateInterface $delegate)
      {
          if (! $this->isAuthorized($request)) {
               throw new UnauthorizedException();
          }

          // ...
      }
  }
  ```
  
It basically boils down to your own personal preference. The other thing to consider, though, is that having nests of try/catch blocks can be potentially expensive; having a lot of error handlers like the above might slow down your application. As always, test and measure to determine the impact.

weierophinney:
@webimpress — regarding having an interactive migration tool, that may be a bit more than I want to bite off. However, I definitely like the idea of a script that will _install_ the migration tooling; it could even detail what the newly installed tools are and their purpose following installation, which would provide the user with the information they need to proceed.

I'll work on that after separating out the tooling; thanks for the idea!

webimpress:
@weierophinney 2a825f4 didn't resolve the issue

```
$ composer show
zendframework/zend-expressive           dev-feature/programmatic be7224a 

$ vendor/bin/expressive-migrate-original-messages
Usage:

  C:/path-to-my-project/vendor/zendframework/zend-expressive/bin/expressive-migrate-original-messages [command] [options]
```

so still it's not working.

I debugged it and problem is with slashes/backslashes:
```
$this->command is 'C:/path-to-my-project/vendor/zendframework/zend-expressive/bin/expressive-migrate-original-messages';

file_exists($this->command) // <-- true
realpath($this->command) === $this->command // <-- false

realpath($this->command) is C:\path-to-my-project\vendor\zendframework\zend-expressive\bin\expressive-migrate-original-messages
```

weierophinney:
Tooling is now separated into the [zendframework/zend-expressive-tooling repository](https://github.com/zendframework/zend-expressive-tooling). I've also added a new vendor binary, `bin/expressive-tooling`, which can be used to install and remove the tooling support:

```bash
$ ./vendor/bin/expressive-tooling # installs it as a dev dependency
$ ./vendor/bin/expressive-tooling remove # removes it
```

I will update the documentation shortly to reflect these changes; otherwise, the docs are still correct in the tooling usage and what they each do.

weierophinney:
> I debugged it and problem is with slashes/backslashes:

If you can, could you submit a PR against the new repo to fix this? I don't have a working PHP install in any of my windows envs currently, so I cannot easily test this.

bin/expressive-tooling
```diff
@@ -0,0 +1,90 @@
...
+}
+
+$command = $argc > 1 ? $argv[1] : '';
+$remove  = false;
+
+switch (strtolower($command)) {
+    case 'remove':
+        echo "Removing zend-expressive-tooling\n";
+        $remove  = true;
+        $command = sprintf('%s remove --dev zendframework/zend-expressive-tooling', $composer);
```
weierophinney: Yes - otherwise composer will prompt you, indicating it did not find the required package, but one exists as a dev requirement, and ask if you want to remove that. Having the flag in the command allows it to run without user intervention.

weierophinney:
Rebased to remove commits related to the tooling support, which is now in [zendframework/zend-expressive-tooling](https://github.com/zendframework/zend-expressive-tooling). I'll add a commit shortly for the documentation to detail how to install the tooling prior to using the various scripts it defines.

weierophinney:
All documentation is complete; ready for final review!

src/ApplicationConfigInjectionTrait.php
```diff
@@ -0,0 +1,306 @@
...
+            return;
+        }
+
+        foreach ($config['routes'] as $spec) {
+            if (! isset($spec['path']) || ! isset($spec['middleware'])) {
+                continue;
+            }
+
+            if (isset($spec['allowed_methods'])) {
+                $methods = $spec['allowed_methods'];
```
malukenho: why not 
```php
$methods = isset($spec['allowed_methods']) ? $spec['allowed_methods'] : Route::HTTP_METHOD_ANY;

if (! is_array($methods)) {
    throw new Container\Exception\InvalidArgumentException(sprintf(
        'Allowed HTTP methods for a route must be in form of an array; received "%s"',
        gettype($methods)
    ));
}
```

src/ApplicationConfigInjectionTrait.php
```diff
@@ -0,0 +1,306 @@
...
+     */
+    private function createCollectionMapper()
+    {
+        $pipelineMap = $this->createPipelineMapper();
+        $appMiddlewares = [
+            Application::ROUTING_MIDDLEWARE,
+            Application::DISPATCH_MIDDLEWARE,
+        ];
+
+        return function ($item) use ($pipelineMap, $appMiddlewares) {
```
malukenho: mark as `static` to protect against `$this` access. I'm not sure if there's some gain on it on that context btw.

src/ApplicationConfigInjectionTrait.php
```diff
@@ -0,0 +1,306 @@
...
+            if (in_array($item, $appMiddlewares, true)) {
+                return ['middleware' => $pipelineMap($item)];
+            }
+
+            if (! is_array($item) || ! array_key_exists('middleware', $item)) {
+                throw new Container\Exception\InvalidArgumentException(sprintf(
+                    'Invalid pipeline specification received; must be an array containing a middleware '
+                    . 'key, or one of the Application::*_MIDDLEWARE constants; received %s',
+                    (is_object($item) ? get_class($item) : gettype($item))
```
malukenho: unneeded `()`

src/MarshalMiddlewareTrait.php
```diff
@@ -159,4 +212,46 @@ private function marshalLazyErrorMiddlewareService($middleware, ContainerInterfa
...
+     *
+     * @param callable $middleware
+     * @return bool
+     */
+    private function isCallableInteropMiddleware(callable $middleware)
+    {
+        $r = $this->reflectMiddleware($middleware);
+        $paramsCount = $r->getNumberOfParameters();
+
+        if ($paramsCount !== 2) {
```
malukenho: Can be `return $paramsCount === 2`?

src/Middleware/ErrorResponseGenerator.php
```diff
@@ -0,0 +1,145 @@
...
+        $this->template  = $template;
+    }
+
+    /**
+     * @param \Throwable|\Exception $e
+     * @param ServerRequestInterface $request
+     * @param ResponseInterface $response
+     * @return ResponseInterface
+     */
+    public function __invoke($e, ServerRequestInterface $request, ResponseInterface $response)
```
malukenho: `\Exception $e`? I saw the `\Throwable` on the docblock, but still better rely on `\Exception` them nothing as it's not checking for it in the code.

src/Middleware/NotFoundHandler.php
```diff
@@ -0,0 +1,75 @@
...
+     *
+     * @return ResponseInterface
+     */
+    public function process(ServerRequestInterface $request, DelegateInterface $delegate)
+    {
+        if (! $this->renderer) {
+            return parent::process($request, $delegate);
+        }
+
+        $response = $this->responsePrototype->withStatus(404);
```
malukenho: No constant for `404`?

weierophinney:
Just did the following with [Adam Culp's Expressive workshop repo](https://github.com/adamculp/expressive-workshop) to see how upgrade might work:

- Added a repository entry to the `composer.json` pointing at my repository.
- Updated the `zend-expressive` requirement to `dev-feature/programmatic as 1.1.0`.
- Ran `composer update zendframework/zend-expressive zendframework/zend-stratigility` (the second package needed to be listed, as composer does not update nested dependencies when specifying a specific package).

At this point, the application continued to work, but emitted some deprecation notices due to the fact that, out-of-the-box, we use Stratigility error middleware.

I then ran `./vendor/bin/expressive-tooling` to install the zend-expressive-tooling package. Once installed, I ran `./vendor/bin/expressive-pipeline-from-config generate`. This ran without errors, and at this point, the application worked perfectly, emitting no errors.

I then ran each of the other migration tools (original messages and error middleware scanners); these caught nothing, meaning no further changes needed to be made.

I hand-edited the generated `config/pipeline.php` file, as the project defines a number of middleware that modify the returned request; I moved these to run before the error handler. Everything continues to work!

As such, I think we're in pretty good shape, but I'd like to see if anybody else can do similarly with their own apps and report back regarding the experience they had.

src/ApplicationConfigInjectionTrait.php
weierophinney: Because `Route::HTTP_METHOD_ANY` is not an array, but an integer.

src/ApplicationConfigInjectionTrait.php
weierophinney: Can't be static, as it's closing over provided values. Additionally, I don't see what value that would add; this is returned from a private method, so those extending would not be able to re-bind anyways.

src/Middleware/ErrorResponseGenerator.php
weierophinney: Except that `\Throwable` does not inherit from `\Exception` in PHP 7, which would make things like `TypeError` `$e` values create _new_ `TypeError`s.

This is a necessary evil until we can move to PHP 7+ only.

src/Middleware/NotFoundHandler.php
weierophinney: Not within our libraries, but it's a good point. I'll add `fig/http-message-util` as a dependency so we can use a constant here.

src/MarshalMiddlewareTrait.php
weierophinney: Updated.

src/ApplicationConfigInjectionTrait.php
weierophinney: Fixed.

malukenho:
@weierophinney I did the upgrade as you said above, and worked fine for me.

src/Middleware/ErrorResponseGenerator.php
weierophinney: Context is key here; it's being invoked from `Zend\Stratigility\Middleware\ErrorHandler`, which is invoking it with an exception or throwable that it has caught. Adding a check here is unnecessary.

danizord:
Worked fine here as well :+1:

bin/expressive-tooling
```diff
@@ -0,0 +1,90 @@
...
+
+if (false === ($paths = getenv('PATH'))) {
+    fwrite(STDERR, "No PATH environment provided; cannot execute.\n");
+    fwrite(STDERR, "Please manually run 'composer require --dev zendframework/zend-expressive-tooling'\n");
+    fwrite(STDERR, "or 'composer remove zendframework/zend-expressive-tooling'.\n");
+    exit(1);
+}
+
+$paths    = explode(PATH_SEPARATOR, $paths);
+$bins     = ['composer', 'composer.bat', 'composer.phar'];
```
xtreamwayz: Windows can have .cmd as well, which I have :)
```
$bins     = ['composer', 'composer.bat', 'composer.cmd', 'composer.phar'];
```

xtreamwayz:
I did this:
- Added a repository entry to the composer.json pointing at my repository.
- Updated the zend-expressive requirement to dev-feature/programmatic as 1.1.0.
Followed by `composer update` and it updated every package, including stratigility to version 1.3.1

I've converted 2 websites successfully.

src/ApplicationConfigInjectionTrait.php
```diff
@@ -0,0 +1,306 @@
...
+    private function createPriorityQueueReducer()
+    {
+        // $serial is used to ensure that items of the same priority are enqueued
+        // in the order in which they are inserted.
+        $serial = PHP_INT_MAX;
+        return function ($queue, $item) use (&$serial) {
+            $priority = isset($item['priority']) && is_int($item['priority'])
+                ? $item['priority']
+                : 1;
+            $queue->insert($item, [$priority, $serial--]);
```
webimpress: We discuss before `++` operator and here it could be confusing.
I'd prefer to have here just `$serial` and then in new line `$serial--` (or `$serial -= 1`).
Ofc this is post decrementation so value is returned first and then value is changed.

test/ApplicationConfigInjectionTest.php
```diff
@@ -0,0 +1,413 @@
...
+                return false;
+            }
+
+            if (isset($spec['allowed_methods'])) {
+                if ($route->getAllowedMethods() !== $spec['allowed_methods']) {
+                    return false;
+                }
+            }
+
+            if (! isset($spec['allowed_methods'])) {
```
webimpress: `else` to above previous condition?

src/Application.php
```diff
@@ -426,7 +446,10 @@ public function routeMiddleware(ServerRequestInterface $request, ResponseInterfa
             if ($result->isMethodFailure()) {
                 $response = $response->withStatus(405)
                     ->withHeader('Allow', implode(',', $result->getAllowedMethods()));
-                return $next($request, $response, 405);
+
+                return $this->raiseThrowables
+                    ? $response
+                    : $next($request, $response, 405);
```
webimpress: Should we use constant also here? (`StatusCodeInterface::STATUS_METHOD_NOT_ALLOWED`)

src/ApplicationConfigInjectionTrait.php
weierophinney: This code was originally copy-pasted from zend-stdlib's priority queue implementation (which ensures that items injected with the same priority execute in the order in which they are injected); likely need to revisit in that library as well. Updating now.

src/Application.php
weierophinney: Yep. Also needed to update TemplatedErrorHandler, as it was doing 404s.

src/ApplicationConfigInjectionTrait.php
```diff
@@ -0,0 +1,306 @@
...
+                    throw new Container\Exception\InvalidArgumentException(sprintf(
+                        'Allowed HTTP methods for a route must be in form of an array; received "%s"',
+                        gettype($methods)
+                    ));
+                }
+            } else {
+                $methods = Route::HTTP_METHOD_ANY;
+            }
+            $name    = isset($spec['name']) ? $spec['name'] : null;
+            $route   = new Route($spec['path'], $spec['middleware'], $methods, $name);
```
weierophinney: Done.

src/ApplicationConfigInjectionTrait.php
```diff
@@ -0,0 +1,306 @@
...
+     *
+     * The following configuration structure can be used to define routes:
+     *
+     * <code>
+     * return [
+     *     'routes' => [
+     *         [
+     *             'path' => '/path/to/match',
+     *             'middleware' => 'Middleware Service Name or Callable',
+     *             'allowed_methods' => [ 'GET', 'POST', 'PATCH' ],
```
weierophinney: Done.

test/ApplicationConfigInjectionTest.php
```diff
@@ -0,0 +1,413 @@
...
+        }, false));
+    }
+
+    public function callableMiddlewares()
+    {
+        return [
+           ['HelloWorld'],
+           [
+                function () {
+                }
```
weierophinney: Done

test/Container/ApplicationFactoryTest.php
```diff
@@ -1308,4 +877,141 @@ public function testProperlyRegistersNestedErrorMiddlewareAsLazyErrorMiddleware(
...
+        $noPath = clone $api;
+        $goodbye = clone $api;
+        $pipelineFirst = clone $api;
+        $hello = clone $api;
+        $pipelineLast = clone $api;
+
+        $config = [
+            'middleware_pipeline' => [
+                [ 'path' => '/api', 'middleware' => $api ],
+                [ 'path' => '/dynamic-path', 'middleware' => 'DynamicPath' ],
```
weierophinney: Done, throughout (there were a lot of them!).

test/Container/NotFoundHandlerFactoryTest.php
```diff
@@ -0,0 +1,72 @@
...
+        $this->assertAttributeEquals('error::404', 'template', $handler);
+    }
+
+    public function testTemplateIsSetWhenFoundInConfiguration()
+    {
+        $this->container->has('config')->willReturn(true);
+        $this->container->get('config')->willReturn(['zend-expressive' => [
+            'error_handler' => [
+                'template_404' => 'error::custom404'
+            ]
```
weierophinney: Done.

PR #398, opened 2016-11-28
HEAD and OPTIONS requests

webimpress:
Documentation says:

> // Matching NO HTTP methods
> // Pass an empty array to the HTTP methods. HEAD and OPTIONS will still be
> // honored. (In FastRoute, GET is also honored.)
> $app->route('/post', 'WillThisHandlePost', []);
> https://github.com/zendframework/zend-expressive/blob/master/doc/book/reference/usage-examples.md

As I understand `HEAD` and `OPTIONS` requests should be available when "no http" methods are set (an empty array), and also `GET` request when using FastRoute.
I've added test that shows it's not working correctly.

Also added two other tests:
- check if HEAD/OPTIONS requests are __allowed__ when list of allowed methods is set,
- check if GET*/POST/PUT/DELETE/PATCH request are __not allowed__ when "no http" methods is set.

GET* - as docs says in FastRoute GET method is honored (as we can see in my tests it's not working as well).

We can discuss what is the correct way and what should contain body for HEAD and OPTIONS requests (probably we need more tests for it). My tests shows that for different route providers it works different way. IMHO, it should be consistent, doesn't matter what router is used.

### DIFF
--- a/test/RouteMiddlewareTest.php
+++ b/test/RouteMiddlewareTest.php
@@ -522,6 +522,126 @@ class RouteMiddlewareTest extends TestCase
         $this->assertEquals('Middleware', (string) $result->getBody());
     }
 
+    public function allowedMethod()
+    {
+        return [
+            'aura-head'          => [AuraRouter::class, 'HEAD'],
+            'aura-options'       => [AuraRouter::class, 'OPTIONS'],
+            'fast-route-head'    => [FastRouteRouter::class, 'HEAD'],
+            'fast-route-options' => [FastRouteRouter::class, 'OPTIONS'],
+            'fast-route-get'     => [FastRouteRouter::class, 'GET'],
+            'zf2-head'           => [ZendRouter::class, 'HEAD'],
+            'zf2-options'        => [ZendRouter::class, 'OPTIONS'],
+        ];
+    }
+
+    public function notAllowedMethod()
+    {
+        return [
+            'aura-get'          => [AuraRouter::class, 'GET'],
+            'aura-post'         => [AuraRouter::class, 'POST'],
+            'aura-put'          => [AuraRouter::class, 'PUT'],
+            'aura-delete'       => [AuraRouter::class, 'DELETE'],
+            'aura-patch'        => [AuraRouter::class, 'PATCH'],
+            'fast-route-post'   => [FastRouteRouter::class, 'POST'],
+            'fast-route-put'    => [FastRouteRouter::class, 'PUT'],
+            'fast-route-delete' => [FastRouteRouter::class, 'DELETE'],
+            'fast-route-patch'  => [FastRouteRouter::class, 'PATCH'],
+            'zf2-get'           => [ZendRouter::class, 'GET'],
+            'zf2-post'          => [ZendRouter::class, 'POST'],
+            'zf2-put'           => [ZendRouter::class, 'PUT'],
+            'zf2-delete'        => [ZendRouter::class, 'DELETE'],
+            'zf2-patch'         => [ZendRouter::class, 'PATCH'],
+        ];
+    }
+
+    /**
+     * @dataProvider allowedMethod
+     *
+     * @param string $adapter
+     * @param string $method
+     */
+    public function testAllowedMethodsWhenOnlyPutMethodSet($adapter, $method)
+    {
+        $app = new Application(new $adapter);
+
+        // Add a PUT route
+        $app->put('/foo', function ($req, $res, $next) {
+            $res->getBody()->write('Middleware');
+            return $res;
+        });
+
+        $next = function ($req, $res) {
+            return $res;
+        };
+
+        $request  = new ServerRequest(['REQUEST_METHOD' => $method], [], '/foo', $method);
+        $response = new Response();
+        $result   = $app->routeMiddleware($request, $response, function ($request, $response) use ($app, $next) {
+            return $app->dispatchMiddleware($request, $response, $next);
+        });
+        $this->assertEquals(200, $result->getStatusCode());
+        $this->assertEquals('Middleware', (string) $result->getBody());
+    }
+
+    /**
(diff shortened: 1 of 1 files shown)


weierophinney:
> as docs says in FastRoute GET method is honored (as we can see in my tests it's not working as well).

I'd argue this is incorrect behavior on the part of FastRoute. While it's unusual to not honor GET for a given path, the various HTTP specifications do not say that servers MUST honor it. As such, I'd argue that we should not include the test case for that in this set.

weierophinney:
I'm looking at the code now, and thinking about how this should work.

First, I agree that `HEAD` and `OPTIONS` requests should always be honored. The question is, what should that look like?

When you do a `HEAD` request, the request is generally expected to send the same headers as if you had made a `GET` request to the resource (though this may not be true if the server does not support `GET` to that particular resource; in that case, it may be something else entirely). This might then imply that the routing middleware should consider this a successful route match, delegate to the next, and, once a response is received, return the response _with an empty body_.

When you do an `OPTIONS` request, you _generally_ are mainly interested in emitting the `Allow` header, which is what we do on a routing _failure_, but not what we do currently on a routing _success_, even when it's based on an `OPTIONS` method. However, `OPTIONS` can be handled in many ways, and provide additional detail; [Zac Stewart has a nice post on using it for self-documenting APIs](http://zacstewart.com/2012/04/14/http-options-method.html). So, then, the question becomes: should we just go ahead and pass `OPTIONS` requests on to the next middleware? or should we return a response immediately with an `Allow` header and 200 status, and be done?

In reviewing the existing routers, we have an interesting situation:

- FastRoute always honors `HEAD` and `OPTIONS` requests for a matched route; this means, however, that whatever middleware is mapped will be invoked for these types of requests, always.
- Neither the zend-router or aura.router support `HEAD` or `OPTIONS` requests out of the box, unless `HTTP_METHOD_ANY` is specified. As such, specifying either of these methods currently results in either a 405 or a 500 response.

Here are my thoughts:

- We should honor `HEAD` and `OPTIONS` requests always.
- For `HEAD`:
    - In individual routers, if the matched route DOES NOT ALLOW the `HEAD` OR `GET` methods, we would need to flag the returned route result somehow to indicate this, and the routing middleware would need to return a 200 response immediately.  This would ensure we do not have a 404.

    - If the matched route DOES ALLOW the `HEAD` method, we would delegate handling on to the next middleware, and then need to ensure that the returned body is empty. (More on that on the next point.)

    - In individual routers, if the matched route DOES NOT ALLOW the `HEAD` method, but DOES ALLOW the `GET` method, we would need to flag the returned route result somehow to indicate this, and we would then proceed as normal for a routing match, potentially altering the request method sent to the next layer to `GET`. Once the response is returned, we would replace the response body with an empty stream. I would argue that we should do this via new middleware, which users can OPTIONALLY register in an outer layer (and which, in 1.1, we'd register by default), and which would look something like the following:

      ```php
      use Zend\Diactoros\Stream;

      function ($request, $delegate) {
          $response = $delegate->process($request);
          if ('HEAD' !== $request->getMethod()) {
              return $response;
          }

          return $response->withBody(new Stream('php://temp'));
      }
      ```

      Note, this also means that the middleware author would be required to test the incoming method, particularly if they allow more than just GET to reach the middleware.

- For `OPTIONS`:
    - In individual routers, if the matched route DOES NOT ALLOW the `OPTIONS` method, we would need to flag the route result somehow to indicate this, and the routing middleware would then return a 200 response with an `Allow` header. (Essentially, this is the same as our current 405 handling, but with a 200 status instead.)

    - If a route DOES ALLOW the `OPTIONS` method, we would delegate to the next middleware as-is, and simply return the response.

My inclination is to handle this via a combination of new functionality in zend-expressive-router, the invididual router implementations, and in Expressive itself.

First, we'd update the constructor logic of `Zend\Expressive\Router\Route` to add `HEAD` and `OPTIONS` to the list of accepted HTTP methods for the route if they are not provided; we'd then also provide the following additional methods on that class:

- `implicitHead()`
- `implicitOptions()`

These would allow the router implementations to check if a match for a `HEAD` or an `OPTIONS` request was due to the user specifying these as acceptable methods, or the `Route` class adding them during instantiation.

Additionally, `Zend\Expressive\Router\RouteResult` would need to be modified to allow storing the actual matched `Route` instance in the `RouteResult`, instead of the matched route name, middleware, and params. This could be done via a new method `fromRoute()` (vs the current `fromRouteMatch()`), and it would inject a `$route` property.  The existing methods `getMatchedRouteName()`, `getMatchedMiddleware()`, and `getMatchedParams()` would be updated to proxy to the underlying `Route` instance if present, and a new method, `getMatchedRoute()` would be added to retrieve the `Route` instance, allowing consumers to test for `implictHead()` and/or `implicitOptions()`.

Once those changes were made, individual routing implementations could then be updated to use `RouteResult::fromRoute()` with the matched `Route` instance.

- For `FastRouteRouter`, this is relatively straight-forward, as we are already pulling data from the `Route` instance when creating the `RouteResult`.
- For `AuraRouter`, we could store the `Route` instance as one of the params provided to the generated `AuraRoute` instance; when marshaling the `RouteResult`, we'd use that value from the `AuraRoute`.
- For `ZendRouter`, we could store the `Route` instance as one of the `defaults`, and then access it via the matched route params.

These changes could be done in an immediate minor release of zend-expressive-router, and subsequent minor releases of the individual router implementations, without BC concerns as the current `RouteResult` API would still work for existing Expressive users; a new minor release of Expressive (I'd like to target 1.1 with this, as I think it's quite important!) could then depend on the new zend-expressive-router version, and update the routing and/or dispatch middleware as noted above in order to provide implicit `OPTIONS` and `HEAD` handling.

So, consider this a request for comments. Pinging @michaelmoussa and @xtreamwayz.

weierophinney:
Closing in favor of #413, which incorporates the commits from this pull request.

PR #399, opened 2016-11-28
Documentation fixes

webimpress:
`$response->getBody()->write(...)` instead of `$response->write(...)`


weierophinney:
Thanks, @webimpress

PR #404, opened 2016-12-07
Register only one error handler in `swallowDeprecationNotices()`

adrium:
Handlers are not popped correctly in `__invoke()`:

`$handler` from `swallowDeprecationNotices()` is still active after running the application which again swallows all errors after `Zend\Expressive\Application` finished execution.

The modified test passes for this commit, but it does not in the previous (but it should).

Edit: Would it be desired to also open a corresponding issue?

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -677,24 +677,16 @@ class Application extends MiddlewarePipe implements Router\RouteResultSubjectInt
      */
     private function swallowDeprecationNotices()
     {
-        $handler = function ($errno, $errstr) {
-            if ($errno !== E_USER_DEPRECATED) {
-                return false;
+        $handler = function ($errno, $errstr, $errfile, $errline, $errcontext) use (&$previous) {
+            $swallow = $errno === E_USER_DEPRECATED && false !== strstr($errstr, 'error middleware is deprecated');
+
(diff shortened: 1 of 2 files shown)


src/Application.php
```diff
@@ -677,24 +677,16 @@ private function checkForDuplicateRoute($path, $methods = null)
      */
     private function swallowDeprecationNotices()
     {
-        $handler = function ($errno, $errstr) {
-            if ($errno !== E_USER_DEPRECATED) {
-                return false;
+        $handler = function ($errno, $errstr, $errfile, $errline, $errcontext) use (&$previous) {
```
weierophinney: You'll need to define `$previous` prior to declaring this handler; references can sometimes lead to issues otherwise.

weierophinney: I can make this change on merge.

weierophinney:
Thanks, @adrium

adrium:
The PR might seem a bit fussy, but I like Expressive and that was something I could contribute :)

> You'll need to define `$previous` prior to declaring this handler; references can sometimes lead to issues otherwise.

I am interested in what cases? I tested it with `error_reporting(~0)` and it was ok... Actually, I had `$previous = null` previously in a commit I force push overwrote.

PR #406, opened 2016-12-12
Remove route result observers

danizord:
Closes #376


danizord:
@weierophinney should we remove the `result-observers.md` completely from docs?

weierophinney:
> should we remove the result-observers.md completely from docs?

Yes.

weierophinney:
Thanks, @danizord\!

PR #407, opened 2016-12-13
Add tests for exception classes

malukenho:


test/Exception/BadMethodCallExceptionTest.php
```diff
@@ -0,0 +1,26 @@
...
+use Zend\Expressive\Exception\ExceptionInterface;
+
+final class BadMethodCallExceptionTest extends \PHPUnit_Framework_TestCase
+{
+    public function testException()
+    {
+        $exception = new BadMethodCallException();
+
+        $this->assertInstanceOf(ExceptionInterface::class, $exception);
+        $this->assertInstanceOf(BadMethodCallException::class, $exception);
```
danizord: This one is useless given the behavior of `new` in PHP? :)

test/Exception/BadMethodCallExceptionTest.php
```diff
@@ -0,0 +1,25 @@
...
+{
+    public function testException()
+    {
+        $exception = new BadMethodCallException();
+
+        $this->assertInstanceOf(ExceptionInterface::class, $exception);
+
+        $this->setExpectedException(BadMethodCallException::class);
+
+        throw $exception;
```
Ocramius: As discussed before, this is not the right way to test an exception in my opinion, but I'll have to think about it.

weierophinney:
In the majority of cases, these are simply _types_ that are thrown and/or caught internally, and are thus covered elsewhere. I'd argue that if we need coverage for specific exceptions, we would want to write tests that ensure that code that would raise or catch one of these types is exercised, and then annotate that the test covers that exception type.

weierophinney:
BTW, #414 now covers several of these exceptions via new unit tests.

malukenho:
Closing as not necessary

PR #408, opened 2016-12-13
fast-route caching doc draft

pine3ree:
Hello @weierophinney,

a few quick considerations that came to my mind (under the influence of the flu)

1. I am not a native english speaker....so please ... have mercy :-)...no, really, pls edit, edit....
   ...maybe I was over naive in some explainations.

2. Should we add a short intro about what fast-route dispatch data is, how fast.route caching works?

3. Would you prefer to have the class constant suggested by @Ocramius as the config keys?
  namely `FastRouteRouter::CONFIG_CACHE_ENABLED` and `FastRouteRouter::CONFIG_CACHE_FILE`

4. I used the default `cache_file` option value to be (both in the config file and in the router class)  `'data/cache/fastroute.php.cache'`. I use to prefix cache files with `php.cache` for php-array/serialized array, `json.cache` for json array/serialized data etc etc. If you prefer or think that can be a source of confusion we could just use `'data/cache/fastroute.cache'`.
(edited: or `'data/cache/fastroute.php'` to be aligned with the new `app_config.php` name)

5. in the `loadDispatchData()` method [src](https://github.com/zendframework/zend-expressive-fastroute/blob/develop/src/FastRouteRouter.php#L450): is there a way to have the error suppressor handler in one line pass the CS checker? I am quite sure i saw a one-line version somewhere in zend code.

kind regards
maks

### DIFF
--- a/doc/book/features/router/fast-route.md
+++ b/doc/book/features/router/fast-route.md
@@ -274,8 +274,8 @@ $container['Zend\Expressive\Router\RouterInterface'] = new RouterFactory();
 
 ### FastRoute caching support
 
-Starting from version 1.3.0 zend-expressive-fastroute supports fast-route 
-dispatch data caching.
+Starting from version 1.3.0 zend-expressive-fastroute comes with support 
+for fast-route native dispatch data caching.
 
 To enable this feature a couple of modifications are needed to be made in 
 the `routes.global.php` configuration file. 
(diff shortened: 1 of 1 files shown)


pine3ree:
close and reopen to trigger travis check

weierophinney:
Thanks, @pine3ree\!

PR #413, opened 2016-12-15
Allow HEAD and OPTIONS requests for any matched route

weierophinney:
Updates zend-expressive-router dependency to 1.3.2+, and related new versions of router implementations to those supporting that version.

This patch also introduces two new classes, `Zend\Expressive\Middleware\ImplicitHeadMiddleware` and `ImplicitOptionsMiddleware`. These can be placed after the routing middleware, but before the dispatch middleware, in order to handle HEAD and OPTIONS requests when the matched `Route` does not explicitly support them. In each case, they trigger only when:

- the method is the one they are specifically interested in;
- the request composes a `RouteResult` attribute;
- the `RouteResult` composes a `Route` instance (per the changes in zend-expressive-router and the various implementations);
- the `Route` instance indicates that the method is only supported implicitly.

Additionally, for routes supporting `GET` requests, the `ImplicitHeadMiddleware` will dispatch the next middleware in the layer, using `GET` as the request method, and ensure the returned response has an empty body.

This approach ensures that if a route _explicitly_ supports the given method, then the matched middleware is dispatched as normal.

Fixes #398

### DIFF
--- a/src/Middleware/ImplicitHeadMiddleware.php
+++ b/src/Middleware/ImplicitHeadMiddleware.php
@@ -1,0 +1,111 @@
+<?php
+/**
+ * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
+ * @copyright Copyright (c) 2016 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   https://github.com/zendframework/zend-expressive/blob/master/LICENSE.md New BSD License
+ */
+
+namespace Zend\Expressive\Middleware;
+
+use Fig\Http\Message\RequestMethodInterface as RequestMethod;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Zend\Diactoros\Response;
+use Zend\Diactoros\Stream;
+use Zend\Expressive\Router\RouteResult;
+
+/**
+ * Handle implicit HEAD requests.
+ *
+ * Place this middleware after the routing middleware so that it can handle
+ * implicit HEAD requests -- requests where HEAD is used, but the route does
+ * not explicitly handle that request method.
+ *
+ * When invoked, it will create an empty response with status code 200.
+ *
+ * You may optionally pass a response prototype to the constructor; when
+ * present, that instance will be returned instead.
+ *
+ * The middleware is only invoked in these specific conditions:
+ *
+ * - a HEAD request
+ * - with a `RouteResult` present
+ * - where the `RouteResult` contains a `Route` instance
+ * - and the `Route` instance defines implicit HEAD.
+ *
+ * In all other circumstances, it will return the result of the delegate.
+ *
+ * If the route instance supports GET requests, the middleware dispatches
+ * the next layer, but alters the request passed to use the GET method;
+ * it then provides an empty response body to the returned response.
+ */
+class ImplicitHeadMiddleware
+{
+    /**
+     * @var null|ResponseInterface
+     */
+    private $response;
+
+    /**
+     * @param null|ResponseInterface $response Response prototype to return
+     *     for implicit HEAD requests; if none provided, an empty zend-diactoros
+     *     instance will be created.
+     */
+    public function __construct(ResponseInterface $response = null)
+    {
+        $this->response = $response;
+    }
+
+    /**
+     * Handle an implicit HEAD request.
+     *
+     * If the route allows GET requests, dispatches as a GET request and
+     * resets the response body to be empty; otherwise, creates a new empty
+     * response.
+     *
+     * @param ServerRequestInterface $request
+     * @param ResponseInterface $response
+     * @param callable $next
+     * @return ResponseInterface
+     */
+    public function __invoke(ServerRequestInterface $request, ResponseInterface $response, callable $next)
+    {
+        if ($request->getMethod() !== RequestMethod::METHOD_HEAD) {
+            return $next($request, $response);
+        }
+
+        if (false === ($result = $request->getAttribute(RouteResult::class, false))) {
+            return $next($request, $response);
+        }
+
+        $route = $result->getMatchedRoute();
+        if (! $route || ! $route->implicitHead()) {
+            return $next($request, $response);
+        }
+
+        if (! $route->allowsMethod(RequestMethod::METHOD_GET)) {
+            return $this->getResponse();
+        }
+
+        $response = $next(
+            $request->withMethod(RequestMethod::METHOD_GET),
+            $response
+        );
+
+        return $response->withBody(new Stream('php://temp/', 'wb+'));
+    }
+
+    /**
+     * Return the response prototype to use for an implicit HEAD request.
+     *
+     * @return ResponseInterface
+     */
+    private function getResponse()
+    {
+        if ($this->response) {
+            return $this->response;
+        }
+
+        return new Response();
+    }
+}
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         "container-interop/container-interop": "^1.1",
         "psr/http-message": "^1.0",
         "zendframework/zend-diactoros": "^1.1",
-        "zendframework/zend-expressive-router": "^1.1",
+        "zendframework/zend-expressive-router": "^1.3.2",
         "zendframework/zend-expressive-template": "^1.0.1",
         "zendframework/zend-stratigility": "^1.3.1",
         "fig/http-message-util": "^1.1"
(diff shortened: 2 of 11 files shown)


src/Application.php
```diff
@@ -753,4 +776,53 @@ private function swallowDeprecationNotices()
...
+        return $response->withBody(new Stream('php://temp', 'wb+'));
+    }
+
+    /**
+     * Handle an implicit OPTIONS request.
+     *
+     * Returns an empty 200 response with an Allow header.
+     *
+     * @param Router\Route $route
+     * @return ResponseInterface
```
webimpress: Method always returns `Response`.

test/DispatchMiddlewareTest.php
```diff
@@ -0,0 +1,253 @@
...
+            ->withMethod(RequestMethod::METHOD_HEAD)
+            ->withAttribute(RouteResult::class, $result->reveal());
+        $response = clone $expected;
+        $delegate = $this->getDelegate()->reveal();
+
+        $app = $this->getApplication();
+        $test = $app->dispatchMiddleware($request, $response, $delegate);
+
+        $this->assertInstanceOf(Response::class, $test);
+        $this->assertEquals(200, $test->getStatusCode());
```
webimpress: Should we use constant also here?

danizord:
@weierophinney what about shipping them as standalone, optional middlewares?

src/Application.php
```diff
@@ -753,4 +776,53 @@ private function swallowDeprecationNotices()
...
+     * @param ServerRequestInterface $request
+     * @param ResponseInterface $response
+     * @return ResponseInterface
+     */
+    private function handleImplicitHeadRequest(
+        Router\Route $route,
+        callable $middleware,
+        ServerRequestInterface $request,
+        ResponseInterface $response,
+        callable $next
```
weierophinney: Fixed; thanks!

test/DispatchMiddlewareTest.php
weierophinney: Yep; fixed, thanks!

weierophinney:
> @weierophinney what about shipping them as standalone, optional middlewares?

Considering it, but need to figure out how that would work within the `Application` instance. (Same issue I have with the routing/dispatch middleware, to a degree.)

weierophinney:
> Are you going to include some tests from #398 as well or they are redundant now?

Reviewing this now. I merged your changes from that PR locally, and have a ton of failures, and I'm trying to analyze if they're due to the changes I've made, or invalid expectations. I'll keep you posted.

weierophinney:
@webimpress I'm reviewing the tests, and the main issue is that those that trigger the implicit HEAD/OPTIONS usage were checking for a populated response body, when it should not be; updating those tests to look for an empty string for the body makes them pass.

The one I'm struggling with right now is `testNotAllowedMethodWhenNoHttpMethodsSet` when used with the Aura.Router. Evidently, in v3, if you do not specify any HTTP methods, *these routes match any*. The solution is to look for `empty($route->allows)` on the matched route. I'm going to push a 1.1.2 release of that library that does that, and all of your tests from #398 will then pass (with the changes I made for testing the response body contents). As such, I'll close #398 once I've pushed the updates to this PR.

webimpress:
Thanks, @weierophinney !

weierophinney:
@danizord —

> what about shipping them as standalone, optional middlewares?

I've separated them out at this point locally. For the implicit options middleware, it's not a problem at all, as it's only working with the route, which is obtainable from the route result, which is obtainable as a request option.

For the implicit head middleware, things get more dicey, as we need the matched middleware in the case that the route allows GET, but HEAD is only implied. This requires passing it as a constructor argument, which means it _must_ be composed as part of routing or dispatching middleware, since this is typically marshaled from a container.

Ideally, these would operate _between_ the routing and dispatch middleware. However, I do not want to force end-users to add these to their pipeline. As such, still debating; I'll try and make a decision in the next day or two.

danizord:
>For the implicit head middleware, things get more dicey, as we need the matched middleware in the case that the route allows GET, but HEAD is only implied. This requires passing it as a constructor argument, which means it must be composed as part of routing or dispatching middleware, since this is typically marshaled from a container.

If we pipe `ImplicitHeadMiddleware` before dispatch middleware, we don't need to know the routed middleware, we just call `$response = $delegate->process();` to run the routed middleware.

>Ideally, these would operate between the routing and dispatch middleware. However, I do not want to force end-users to add these to their pipeline. As such, still debating; I'll try and make a decision in the next day or two.

Mhmm, why not? We already have some basic functionally such as `ServerUrlMiddleware` and `BodyParamsMiddleware` that are piped manually, and I like this approach because it makes these behaviors explicit and optional, so that developers don't need to wonder why the framework did something or why some header was added to the response.

weierophinney:
@danizord Thought about this over the weekend, and I agree with you. I've updated the patch to provide optional middleware, `Zend\Expressive\Middleware\ImplicitHeadMiddleware` and `ImplicitOptionsMiddleware`, each of which may be slip-streamed between routing and dispatch middleware.

In the skeleton, we can enable these in the default pipeline. I will also likely add some support for them into the tooling for creating a pipeline from existing configuration.

I've pushed the changes at this time, but need to document them. Otherwise, the code is ready for review.

weierophinney:
I've rebased this patch to remove the commits that were updating the dispatch middleware to add the implicit HEAD/OPTIONS support.

src/Application.php
```diff
@@ -19,6 +20,7 @@
 use Zend\Diactoros\Response\SapiEmitter;
 use Zend\Diactoros\ServerRequest;
 use Zend\Diactoros\ServerRequestFactory;
+use Zend\Diactoros\Stream;
```
webimpress: No longer used in that class.

src/Application.php
```diff
@@ -39,6 +41,14 @@ class Application extends MiddlewarePipe implements Router\RouteResultSubjectInt
     use ApplicationConfigInjectionTrait;
     use MarshalMiddlewareTrait;
 
+    /**
+     * HTTP methods that are implicitly allowed for any matched request.
+     */
+    const HTTP_METHODS_IMPLICIT = [
```
webimpress: Seems to be unused. Do we still need it?

weierophinney:
@xtreamwayz This is another one to consider for the skeleton: the `ImplicitHeadMiddleware` and `ImplicitOptionsMiddleware` should be part of the default pipeline, and occur between the routing and dispatch middleware. This could occur either before or after the UrlHelperMiddleware; the patch I have for the tooling puts it before, but that's because I cannot know for certain what else is present between the two.

moderndeveloperllc:
@weierophinney I see these are using the `__invoke()` pattern. Considering that Stratigility 1.3.x is required, would it be better to use the http-interop DelegateInterface pattern, or are y'all not wanting to switch over to that in Expressive right now?

weierophinney:
> I see these are using the __invoke() pattern.

There are changes pending for a 0.4.0 release of http-middleware that will impact the signatures (new namespace, and the `DelegateInterface` now typehints against `ServerRequestInterface` instead of `RequestInterface`), and since I do not know how soon that will drop, I'm waiting to update until a release is made. Once it has, I'll be needing to make changes to Stratigility as well, before we even up date Expressive.

weierophinney:
> I see these are using the __invoke() pattern.

Also, Stratigility provides functionality for wrapping double-pass middleware such as this to work with http-middleware, so it can still be utilized in that context; Expressive will be (eventually) doing this out-of-the-box once moving to a Stratigility version that _only_ supports http-middleware. However, that will not happen until PSR-15 is accepted.

moderndeveloperllc:
@weierophinney Good to know. I guess that's Stratigility 2.0 that will be going http-middleware only.  FYI, it looks like [0.4.0](https://github.com/http-interop/http-middleware/tree/0.4.0) dropped a couple of days ago. Not a huge fan of going from `Interop\Http\Middleware\ServerMiddlewareInterface` to `Interop\Http\ServerMiddleware\MiddlewareInterface`, but I can understand why.

weierophinney:
> Not a huge fan of going from

That was prompted by me, actually. :smile: Before this, we had:

- A generic `Interop\Http\Middleware` interface, with
- a `ServerMiddlewareInterface` that depended on a
- `DelegateInterface`, which depended originally on
- `RequestInterface`, but later was updated to depend on
- `ServerRequestInterface`

In other words, huge inconsistencies in naming that, I felt, would lead to difficulties understanding the roles each played. I provided a number of options forward, and, ultimately, most liked having a different namespace for the server-side middleware artifacts.

As you note, however, it causes us some issues with upgrading. I wasn't aware 0.4.0 had dropped, I need to see what the latest changes are, and make sure there are none others pending before we start adapting support for it.

weierophinney:
Hmmm... there's one open pull request on http-middleware that could still have ramifications, as it would result in at least a rename of `DelegateInterface` to `RequestHandlerInterface`, and potentially a rename of that interface's `process()` method to `__invoke()`. As such, I need to wait until that pull request is resolved before we move forward on Stratigility. In the meantime, as noted above, we can implement `__invoke()` safely.

src/Middleware/ImplicitHeadMiddleware.php
```diff
@@ -0,0 +1,111 @@
...
+
+    /**
+     * Return the response prototype to use for an implicit HEAD request.
+     *
+     * @return ResponseInterface
+     */
+    private function getResponse()
+    {
+        if ($this->response) {
+            return $this->response;
```
malukenho: Why not check and assign it on `__contructor`?

PR #414, opened 2016-12-22
Mutation tests

snapshotpl:


xtreamwayz:
It makes more sense to do this for the develop branch since a lot is changed in that branch.

test/AppFactoryTest.php
```diff
@@ -9,6 +9,20 @@
 
 namespace ZendTest\Expressive;
 
+namespace Zend\Expressive;
+
+use ZendTest\Expressive\AppFactoryTest;
+
+function class_exists($classname)
```
weierophinney: This should be extracted into a separate file; I can do that on merge.

src/Application.php
```diff
@@ -188,7 +196,7 @@ public function __call($method, $args)
         }
 
         // @TODO: we can use variadic parameters when dependency is raised to PHP 5.6
```
weierophinney: Note to self: update the code to use variadics when merging.

test/ApplicationTest.php
```diff
@@ -268,6 +268,17 @@ public function testThrowsExceptionWhenDispatchingUncallableMiddleware($middlewa
         });
     }
 
+    public function testExecuteMiddlewareIfNotRouteMatch()
```
weierophinney: Note to self: rename this to `testDispatchMiddlewareExecutesNextMiddlewareInStackIfNoRouteMatchPresent`.

test/Emitter/EmitterStackTest.php
```diff
@@ -110,16 +110,31 @@ public function testEmitLoopsThroughEmittersUntilOneReturnsNonFalseValue()
         $second->emit(Argument::type(ResponseInterface::class))
             ->willReturn(null);
 
-        $third = $this->prophesize(EmitterInterface::class);
-        $third->emit(Argument::type(ResponseInterface::class))
-            ->willReturn(false);
```
weierophinney: The above was in the test for a specific reason: to ensure that the stack does not execute the third emitter once the second returns a non-false value. Perhaps the better solution would be to have it raise an exception:

```php
->will(function () {
    $this->fail('Should not reach third emitter');
})
```

I can make that change on merge.

src/Application.php
weierophinney: Interestingly, already done there, which means this was out-of-date.

test/Emitter/EmitterStackTest.php
weierophinney: Forgot this was a stack. The actual workflow is:

- Hits `$third`, which returns `false`.
- Hits `$second`, which returns `null`, and should halt execution of the stack.
- Should **not** hit `$first`.

So, the point is to note that the first non-`false` value will stop execution of the stack. We want to ensure that we can go more than one deep in the stack, and that it does not just halt on the first encountered.

weierophinney:
Thanks, @snapshotpl\!

PR #415, opened 2016-12-29
Clean up 

malukenho:


src/Emitter/EmitterStack.php
```diff
@@ -38,15 +38,13 @@ class EmitterStack extends SplStack implements EmitterInterface
      */
     public function emit(ResponseInterface $response)
     {
-        $completed = false;
         foreach ($this as $emitter) {
             if (false !== $emitter->emit($response)) {
-                $completed = true;
-                break;
+                return null;
```
Ocramius: This will stop emitting mid-way: you introduced a bug

malukenho: @Ocramius  there was a `break` on it anyway.

Ocramius:
Good point!

weierophinney:
Build failures are due to a change in filp/whoops, which adds a new method to the pretty page handler, which is called by the whoops runner... and thus is not expected by the mocking library. I've updated it previously on the develop branch, so we can ignore those failures for now.

src/Middleware/ErrorResponseGenerator.php
```diff
@@ -103,11 +103,10 @@ private function prepareTemplatedResponse($e, ServerRequestInterface $request, R
 
     /**
      * @param \Throwable|\Exception $e
-     * @param ServerRequestInterface $request
      * @param ResponseInterface $response
      * @return ResponseInterface
      */
-    private function prepareDefaultResponse($e, ServerRequestInterface $request, ResponseInterface $response)
```
weierophinney: I did this originally for symmetry; all `prepare*()` methods received the same arguments. However, being marked `private`, there's no reason to include the argument if it's not being used.

weierophinney:
Thanks, @malukenho\!

PR #418, opened 2017-01-03
Update year license header

malukenho:
Happy new year 🍻

### DIFF
--- a/src/AppFactory.php
+++ b/src/AppFactory.php
@@ -1,7 +1,7 @@
 <?php
 /**
  * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
- * @copyright Copyright (c) 2015-2016 Zend Technologies USA Inc. (http://www.zend.com)
+ * @copyright Copyright (c) 2015-2017 Zend Technologies USA Inc. (http://www.zend.com)
  * @license   https://github.com/zendframework/zend-expressive/blob/master/LICENSE.md New BSD License
  */
 
--- a/src/Application.php
+++ b/src/Application.php
@@ -1,7 +1,7 @@
 <?php
 /**
  * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
- * @copyright Copyright (c) 2015-2016 Zend Technologies USA Inc. (http://www.zend.com)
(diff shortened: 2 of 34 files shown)


Ocramius:
As suggested previously (can't remember in which project): can we simply drop the end year from licence headers? I did that for my projects, and nobody died.

xtreamwayz:
I'm in favor of dropping the end year. We had a discussion about this on #zftalk a few weeks ago and I've looked up some of the copyright rules. If I understand it correctly the copyright lasts for 95 years or more, but I'm not sure how this works for opensource projects. 

> For works published after 1977, the copyright lasts for the life of the author plus 70 years. However, if the work is a work for hire (that is, the work is done in the course of employment or has been specifically commissioned) or is published anonymously or under a pseudonym, the copyright lasts between 95 and 120 years, depending on the date the work is published.
http://fairuse.stanford.edu/overview/faqs/copyright-basics/

If ZF still exists in 90 years you could renew the license header then. Or maybe set the copyright date to the file creation date. Someone with extended copyright knowledge should figure this out.

weierophinney:
I did research on this a couple years ago. In the US, the copyright notice:

- MUST include the author/entity for which copyright is declared
- SHOULD include the date when it was created
- SHOULD also include the date last modified, if different than the date of creation (thus, the date range format used: 2005-2016, etc.)

Copyright lifetime is based on the last date listed. If you only list the original date, the copyright period will be based on that, but if a date range is present, the copyright extends from the last date listed.

Technically, the date can be omitted entirely, if it can be gleaned from other sources; however, it's always better to be _explicit_. As an example, if a dispute arises regarding whether or not your work is derivative, having the date of origination explicitly listed prevents ambiguity. When a work evolves over time, specifying the range from when it was first declared to last worked on helps define the date at which the copyright will expire; otherwise, expiration is based on date of origination.

What that means for code is this: we do not need to update the copyright on every file in the component. We only need to update _if we make change to a file in a year not present in the copyright date range already present_; we only need to do it _for files we are changing currently_.

This, of course, makes automating checks difficult, as we would need to check if a file changed, and then if the copyright header reflects the current year.

However, it _does_ mean that we do not need to honor pull requests such as this one, as we should be updating copyright dates as new code is written or changes are introduced, and then only if the range is no longer correct.

Issue #419, opened 2017-01-05
Router method failure gives deprecation notice when using raiseThrowables


moderndeveloperllc:
Not sure if this is something that can/will be worked out in the 1.0.x series.

Setup:
expressive 1.0.5
stratigility 1.3.1
Procedural configuration with raiseThrowables()
Using ErrorHandler-style catching - no ErrorMiddleware

If I have a request that hits `Application.php:423`, it returns a `$next($request, $response, 405);` which in `Next.php` will display the deprecation notice. This is the stacktrace from the point of failure:
```
"/blah/vendor/zendframework/zend-stratigility/src/Next.php:435",
"/blah/vendor/zendframework/zend-stratigility/src/Next.php:422",
"/blah/vendor/zendframework/zend-stratigility/src/Next.php:119",
"/blah/vendor/zendframework/zend-expressive/src/Application.php:426",
```
The code for 1.1.x is checking against $this->raiseThrowables. Not sure this could be backported without breaking those who are using stratigility 1.2.x.

weierophinney:
This one is tricky.

My original thought is that we change the invocation of `$next` to instead raise an exception. However, that could be a BC break for existing users. By invoking legacy-style error middleware, technically a consumer *could* write error middleware specifically for handling 405 conditions:

```php
function ($err, $request, $response, $next)
{
    if ($err !== 405) {
        return $next($request, $response, $err);
    }
    // handle the 405 condition...
}
```

This may or may not be useful; however, it *would* allow doing something like injecting a custom response body (such as Problem Details).

Raising an exception would break that use case.

Another use case that would be broken is for developers who have written legacy error middleware for general purpose error handling, instead of configuring or overriding the final handler. In such cases, again, raising an exception would bypass this middleware, breaking the application.

What we _may_ be able to do is call `swallowDeprecationNotices()` immediately before invoking `$next()` within that block. What I'm a bit surprised at, however, is the fact that the tests are not picking up the deprecation notice. I don't see any code within `RouteMiddlewareTest` that performs error handling on deprecation notices, which means it _should_ display the behavior you're indicating. I'll see if I can resolve that, which should allow me to then test a fix using `swallowDeprecationNotices()`.

weierophinney:
> I don't see any code within RouteMiddlewareTest that performs error handling on deprecation notices, which means it should display the behavior you're indicating. I'll see if I can resolve that

[...]

PR #420, opened 2017-01-09
Do not emit deprecation notices for 405 errors

weierophinney:
As reported in #419, when using zend-stratigility 1.3 with Expressive 1.0, 405 errors are also now accompanied by deprecation notices. Since this is default functionality, we need to swallow those deprecation notices for now.

### DIFF
--- a/src/Exception/MissingDependencyException.php
+++ b/src/Exception/MissingDependencyException.php
@@ -1,7 +1,7 @@
 <?php
 /**
  * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
- * @copyright Copyright (c) 2016 Zend Technologies USA Inc. (http://www.zend.com)
+ * @copyright Copyright (c) 2016-2017 Zend Technologies USA Inc. (http://www.zend.com)
  * @license   https://github.com/zendframework/zend-expressive/blob/master/LICENSE.md New BSD License
  */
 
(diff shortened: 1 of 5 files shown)


src/Exception/MissingDependencyException.php
```diff
@@ -1,7 +1,7 @@
 <?php
 /**
  * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
- * @copyright Copyright (c) 2016 Zend Technologies USA Inc. (http://www.zend.com)
+ * @copyright Copyright (c) 2016-2017 Zend Technologies USA Inc. (http://www.zend.com)
```
webimpress: File is not changed, year shouldn't be then changed here. I hope we can do some auto checks on it, or at least some scripts to update it automatically - maybe on merge?

test/WhoopsErrorHandlerTest.php
```diff
@@ -15,6 +15,7 @@
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Message\StreamInterface;
+use ReflectionClass;
```
webimpress: As this file is changed, year in header license notes should be updated to 2017.

src/Exception/MissingDependencyException.php
weierophinney: For some reason, this one kept getting flagged until I changed the date range. I have no idea why, and I didn't want to waste more time on it.

test/WhoopsErrorHandlerTest.php
weierophinney: Thanks, updating now.

PR #422, opened 2017-01-12
Updated to zend-expressive-router 2.0

weierophinney:
Updated the depedency on zend-expressive-router to `^2.0`, and, in development, each router implementation to that same constraint.

Doing so has an intersting side-effect: the route result is guaranteed to have valid middleware composed, as the result is created from a `Route` instance, which will raise an exception much earlier in cases of invalid middleware being passed during its own instantiation. This allowed removing two unit tests, as well as a conditional within `dispatchMiddleware()` for verifying we have callable middleware.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
         "filp/whoops": "^1.1 || ^2.0",
         "malukenho/docheader": "^0.1.5",
         "mockery/mockery": "^0.9.5",
-        "phpunit/phpunit": "^5.7",
+        "phpunit/phpunit": "^5.7.1",
         "zendframework/zend-coding-standard": "~1.0.0",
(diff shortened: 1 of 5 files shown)


Ocramius:
> `Missing argument 1 for ZendTest\Expressive\RouteMiddlewareTest::testMatchWithAllHttpMethods()`

This is happening with low deps, so something else needs bumping too.

weierophinney:
> This is happening with low deps, so something else needs bumping too.

It was a bug in PHPUnit 5.7.0, and fixed in 5.7.1; I've bumped the minimum PHPUnit version to 5.7.1 just now.

weierophinney:
(Reference: https://github.com/sebastianbergmann/phpunit/pull/2381)

PR #424, opened 2017-01-20
Remove unused use ($app) at anonymous function at ApplicationFactory::createCollectionMapper()

samsonasik:


weierophinney:
I've made this change already in #428, which will land in develop this week; since that patch contains substantial additional refactoring, I'm closing this one to prevent merge issues later.

Thanks for finding and reporting this, though!

PR #425, opened 2017-01-24
[Proposal] make Application::run() return a ResponseInterface

stefanotorresi:
Ok, I know some folks won't like it, but hey, proposals don't hurt.

I'd like to change `Zend\Expressive\Application` api so that `run()` **returns a Response**.

The main argument for it is so that functional testing is made easier, as it doesn't require stubbing the `EmitterInterface` and then inspect the response passed to it. The way I see it, the emitter is just an internal detail of `run()`, and as such I should be able to ignore it.

Also, there are other frameworks that do this (e.g. Slim), and some don't even output the response right away but do that during an entirely different method call.

Ultimately, I'd very much like if any test of any web application could be as simple as this:

```php
// setup
$app     = new App(/* etc */);
$request = new Request(/* etc */);

// exec
$response = $app->run($request); 

// assert
assertThat($response, /* etc */); 
```

Is there any compelling reason why `run()` should be void?

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -493,7 +493,9 @@ class Application extends MiddlewarePipe
      * the returned response using the composed emitter.
      *
      * @param null|ServerRequestInterface $request
-     * @param null|ResponseInterface $response
+     * @param null|ResponseInterface      $response
+     *
+     * @return ResponseInterface
      */
     public function run(ServerRequestInterface $request = null, ResponseInterface $response = null)
     {
@@ -501,12 +503,10 @@ class Application extends MiddlewarePipe
             $request  = $request ?: ServerRequestFactory::fromGlobals();
         } catch (InvalidArgumentException $e) {
             // Unable to parse uploaded files
-            $this->emitMarshalServerRequestException($e);
-            return;
+            return $this->emitMarshalServerRequestException($e);
         } catch (UnexpectedValueException $e) {
             // Invalid request method
-            $this->emitMarshalServerRequestException($e);
-            return;
+            return $this->emitMarshalServerRequestException($e);
         }
 
         $response = $response ?: new Response();
@@ -516,6 +516,8 @@ class Application extends MiddlewarePipe
 
         $emitter = $this->getEmitter();
         $emitter->emit($response);
+
+        return $response;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Ocramius:
> Is there any compelling reason why `run()` should be void?

Yes, this is something that was extremely confusing in ZF2.

Specifically, people would EMIT a response inside ZF2, and then further on pass it to a subsequent emitter too.

I suggest keeping this method as `void`. If the users want to do something with the response, they can either not use `run()`, or replace the `run()` implementation (which is trivial, btw).

Unless there is a *strong* reason for not making it `void`, I wouldn't go that way. `run()` is still a magic box.

If you want to retrieve the response from the application, then use it [as a `callable` instead](https://github.com/zendframework/zend-expressive/blob/f0d7feb8bc62d0d323b5113026f000f438ba3b8f/src/Application.php#L139)

stefanotorresi:
:cry:

weierophinney:
@stefanotorresi I agree with @Ocramius here. If you look at the logic of `run()`, it:

- creates request and response instances if none were provided
- invokes itself
- passes the returned response to the emitter

If you want to execute the application and manipulate the returned response, you can do so already by invoking the application:

```php
$response = $app(ServerRequestFactory::fromGlobals(), new Response());
```

The only _extra_ logic `app()` does beyond that is catch exceptions raised by the `ServerRequestFactory`, and, when detected, emit specialized, canned, error responses. If you feel confident that the factory will not fail, the above one-liner accomplishes what you requested.

stefanotorresi:
@weierophinney yep, I get where you're coming from, even if I'd still prefer if the emit phase was separated from producing the response, but @Ocramius made a point for that too in IRC.

Anyway, I was wondering if [this line](https://github.com/zendframework/zend-expressive/blob/f9e631ac946c593939652ee2a26c3ab46d01ac9d/src/Application.php#L513) is functionally relevant and if I should mock that too.

weierophinney:
> Anyway, I was wondering if this line is functionally relevant and if I should mock that too.

Depends. :smile:

We use it internally within the `FinalHandler` in order to determine if the response has changed or not when it is invoked. That said, that functionality goes away with the upcoming 2.0 version, as you will instead use the new middleware-based error handling mechanisms.

The other reason for it, however, is that it can act as a response prototype for your middleware, which you can then use as a basis for creating a response to return:

```php
$response = $request->getAttribute('originalResponse');
$response->getBody()->write($this->marshalProblemDetail($filter->getMessages());
return $response->withStatus(422);
```

PR #428, opened 2017-01-26
Provide Stratigility 2.0 compatibility

weierophinney:
This patch updates Expressive to work with Stratigility 2.0, and thus http-interop/http-middleware 0.4.1+.

In order to accomplish this, a large number of changes were needed:

- Extracted routing middleware to `Zend\Expressive\Middleware\RouteMiddleware`, and removed the `Application::routeMiddleware()` method.
- Extracted dispatch middleware to `Zend\Expressive\Middleware\DispatchMiddleware`, and removed the `Application::dispatchMiddleware()` method.
- Extracted the closure for lazy-loading middleware from the `MarshalMiddlewareTrait` into `Zend\Expressive\Middleware\LazyLoadingMiddleware`.
- Extracted the logic for `isCallableInteropMiddleware()` into the trait `IsCallableInteropMiddlewareTrait`.
- Rewrote the `MarshalMiddlewareTrait` essentially from the ground up, to ensure any middleware that `prepareMiddleware()` returns is http-interop/http-middleware compatible; it also removes all support for the Stratigility v1 "error middleware". Additionally, it now creates and returns instances of `RouteMiddleware`, `DispatchMiddleware`, and `LazyLoadingMiddleware`.
- Updated the `ApplicationConfigInjectionTrait` to no longer worry about Stratigility v1 error middleware, and to use the `Application` constants for routing and dispatch middleware instead of the legacy callables referencing `Application` methods.
- Removed the `ErrorMiddlewarePipe` class and the `Application::pipeErrorMiddleware()` method.
- Refactored all double-pass middleware to be http-interop/http-middleware.
- Moved the `ROUTING_MIDDLEWARE` and `DISPATCH_MIDDLEWARE` constants into the `Application` class; the `ApplicationFactory` class now sets its own to the `Application` constant values.
- Renamed `finalHandler` to `defaultDelegate`, and it is now a `NoopFinalHandler` by default.
- Removed the `__invoke()` method, as it no longer had or required logic different from `Zend\Stratigility\MiddlewarePipe`.

I also did a fair bit of test refactoring, due to the changes listed above. There's likely more to do; the `ApplicationFactoryTest`, for example, has a fair bit of logic mimicing that of the `MarshalMiddlewareTraitTest` which could be removed.

With these changes, the develop branch will now target a new _major_ version, 2.0.0, as the changes required to accommodate Stratigility 2.0 required BC breaks. This also afforded the opportunity to provide some nice architectural changes, such as extracting the routing and dispatch middleware.


snapshotpl:
Great changes! +1,829 −2,677 :heart: 
Good idea to move everything into middlewares (routing and dispatch), I think before about it. Maybe it's something else what we can extract to middewares?

weierophinney:
> Maybe it's something else what we can extract to middewares?

Umm... what do you mean? Any concrete suggestions?

snapshotpl:
I'm just thinking out loud ;)

snapshotpl:
What about migrate to php7 in 2.0?

weierophinney:
> What about migrate to php7 in 2.0?

Not yet; we're still committed to supporting PHP 5.6 until it goes EOL... unless we determine that a majority of users have migrated to 7 before then.

composer.json
```diff
@@ -12,17 +12,17 @@
...
+            "dev-develop": "2.0-dev"
         }
     },
     "require": {
         "php": "^5.6 || ^7.0",
         "container-interop/container-interop": "^1.1",
         "psr/http-message": "^1.0",
         "zendframework/zend-diactoros": "^1.1",
-        "zendframework/zend-expressive-router": "^2.0",
+        "zendframework/zend-expressive-router": "^2.1",
```
weierophinney: This was necessary, because before 2.1, we only allowed callables for the `$middleware` argument to the `Route` constructor. 2.1 allows http-interop/http-middleware 0.4.1 values as well.

composer.json
```diff
@@ -12,17 +12,17 @@
...
         "zendframework/zend-expressive-template": "^1.0.4",
-        "zendframework/zend-stratigility": "^1.3.1",
+        "zendframework/zend-stratigility": "^2.0.1",
```
weierophinney: 2.0.1 contains a fix for how nesting middleware pipes works.

src/ApplicationConfigInjectionTrait.php
```diff
@@ -165,24 +162,24 @@ public function injectRoutesFromConfig(array $config = null)
                 continue;
             }
 
+            $methods = Route::HTTP_METHOD_ANY;
             if (isset($spec['allowed_methods'])) {
                 $methods = $spec['allowed_methods'];
                 if (! is_array($methods)) {
-                    throw new Container\Exception\InvalidArgumentException(sprintf(
+                    throw new Exception\InvalidArgumentException(sprintf(
```
weierophinney: This is technically a BC break (part of the reason for targeting 2.0 instead of 1.1), but will likely never affect anyone.

src/ApplicationConfigInjectionTrait.php
```diff
@@ -196,36 +193,6 @@ public function injectRoutesFromConfig(array $config = null)
...
-     * function ($item) : callable|string
-     * </code>
-     *
-     * It is suitable for mapping pipeline middleware representing the application
-     * routing o dispatching middleware to a callable; if the provided item does not
-     * match either, the item is returned verbatim.
-     *
-     * @return callable
-     */
-    private function createPipelineMapper()
```
weierophinney: This functionality could be removed as `pipe()` now uses `prepareMiddleware()` throughout, and recognizes the `*_MIDDLEWARE` constants properly.

src/Container/ApplicationFactory.php
```diff
@@ -41,23 +39,11 @@
...
- * Additionally, when raise_throwables is enabled, you will need to provide an
- * innermost middleware to invoke when the queue is exhausted; we suggest
- * Zend\Expressive\Middleware\NotFoundHandler.
  */
 class ApplicationFactory
 {
-    const DISPATCH_MIDDLEWARE = 'EXPRESSIVE_DISPATCH_MIDDLEWARE';
-    const ROUTING_MIDDLEWARE = 'EXPRESSIVE_ROUTING_MIDDLEWARE';
+    const DISPATCH_MIDDLEWARE = Application::DISPATCH_MIDDLEWARE;
+    const ROUTING_MIDDLEWARE = Application::ROUTING_MIDDLEWARE;
```
weierophinney: Not technically a BC break; just switches the definitive source for the definitions. I'm keeping these, as anybody who was using configuration-based pipelines would have their code break immediately otherwise.

src/Container/ApplicationFactory.php
```diff
@@ -77,19 +63,13 @@ public function __invoke(ContainerInterface $container)
             ? $container->get(RouterInterface::class)
             : new FastRouteRouter();
 
-        $finalHandler = ! empty($config['zend-expressive']['raise_throwables'])
-            ? $this->marshalNoopFinalHandler($container)
-            : $this->marshalLegacyFinalHandler($container, $config);
+        $delegate = $this->marshalNoopFinalHandler($container);
```
weierophinney: The legacy final handler implementation can be retired with a major version bump. This will be a BC break in behavior, but can be fixed by pointing the `NoopFinalHandler` service to a factory for creating a `FinalHandler` instance if desired.

src/Container/ApplicationFactory.php
```diff
@@ -77,19 +63,13 @@ public function __invoke(ContainerInterface $container)
...
 
         $emitter = $container->has(EmitterInterface::class)
             ? $container->get(EmitterInterface::class)
             : null;
 
-        $app = new Application($router, $container, $finalHandler, $emitter);
-
-        if (! empty($config['zend-expressive']['raise_throwables'])) {
```
weierophinney: `raiseThrowables` has no semantic meaning in Stratigility 2.0; it's just how the `Next` implementation works.

src/IsCallableInteropMiddlewareTrait.php
```diff
@@ -0,0 +1,56 @@
...
+ * @copyright Copyright (c) 2017 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   https://github.com/zendframework/zend-expressive/blob/master/LICENSE.md New BSD License
+ */
+
+namespace Zend\Expressive;
+
+use ReflectionFunction;
+use ReflectionMethod;
+
+trait IsCallableInteropMiddlewareTrait
```
weierophinney: I extracted this so the logic could be re-used in the `LazyLoadingMiddleware`.

src/MarshalMiddlewareTrait.php
```diff
@@ -35,52 +38,55 @@
...
      */
-    private function prepareMiddleware($middleware, ContainerInterface $container = null, $forError = false)
-    {
-        if (is_callable($middleware)) {
+    private function prepareMiddleware(
+        $middleware,
+        Router\RouterInterface $router,
+        ResponseInterface $responsePrototype,
+        ContainerInterface $container = null
+    ) {
```
weierophinney: I modified the signature to require passing the router and response prototype; this prevents the trait from relying on instance state.

src/MarshalMiddlewareTrait.php
```diff
@@ -96,158 +102,63 @@ private function prepareMiddleware($middleware, ContainerInterface $container =
...
-        return new $middleware();
-    }
+        $instance = new $middleware();
 
-    /**
-     * @param string $middleware
-     * @param ContainerInterface $container
-     * @return callable
-     */
-    private function marshalLazyMiddlewareService($middleware, ContainerInterface $container)
```
weierophinney: This method was extracted to `LazyLoadingMiddleware`.

src/Middleware/DispatchMiddleware.php
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * Uses the composed container to marshal the middleware matched during
+ * routing, and then dispatches it.
+ *
+ * If no route result is present in the request, delegates to the next
+ * middleware.
+ *
+ * @internal
+ */
+class DispatchMiddleware implements ServerMiddlewareInterface
```
weierophinney: This was extracted from `Application::dispatchMiddleware()`. The class has no dependencies, as it can reliably assume that any middleware stored in a `RouteResult` is of the http-interop/http-middleware type, and thus only needs to `process()` it.

src/Middleware/RouteMiddleware.php
```diff
@@ -0,0 +1,82 @@
...
+ * the composed response prototype to generate a 405 response; otherwise,
+ * it delegates to the next middleware.
+ *
+ * If routing succeeds, injects the route result into the request (under the
+ * RouteResult class name), as well as any matched parameters, before
+ * delegating to the next middleware.
+ *
+ * @internal
+ */
+class RouteMiddleware implements ServerMiddlewareInterface
```
weierophinney: This was extracted from `Application::routeMiddleware()`. It requires the router instance, as well as a response prototype; the latter is used to create a 405 response when necessary.

test/Application/MarshalMiddlewareTraitTest.php
```diff
@@ -0,0 +1,298 @@
...
+use Zend\Expressive\Application;
+use Zend\Expressive\Exception\InvalidMiddlewareException;
+use Zend\Expressive\Middleware;
+use Zend\Expressive\Router\RouterInterface;
+use Zend\Stratigility\Delegate\CallableDelegateDecorator;
+use Zend\Stratigility\MiddlewarePipe;
+use Zend\Stratigility\Middleware\CallableInteropMiddlewareWrapper;
+use Zend\Stratigility\Middleware\CallableMiddlewareWrapper;
+
+class MarshalMiddlewareTraitTest extends TestCase
```
weierophinney: This test covers all behaviors of the `MarshalMiddlewareTrait` following the refactor.

src/Middleware/LazyLoadingMiddleware.php
```diff
@@ -0,0 +1,82 @@
...
+        if (! is_callable($middleware)) {
+            throw new InvalidMiddlewareException(sprintf(
+                'Lazy-loaded middleware "%s" is neither invokable nor implements %s',
+                $this->middlewareName,
+                ServerMiddlewareInterface::class
+            ));
+        }
+
+        // Callable http-interop middleware
+        if ($this->isCallableInteropMiddleware($middleware)) {
```
moderndeveloperllc: Are there enough users for this "edge case" to be included in a major version change? It would seem that this is targeting a rather small/nonexistent(?) percentage of users that a) have http-interop style middleware and b) use `__invoke()` instead of `process()`. Is this more to allow users to include older 3rd-party http-interop middleware?

weierophinney: It's a way to allow using closures as middleware that are duck-type compatible with http-interop/http-middleware. Consider:

```php
$app->route('/ping', function ($request, DelegateInterface $delegate) {
    return new JsonResponse(['ack' => time()]);
}, 'ping');
```

This is nice for prototyping.

You *could*, of course, pass an anonymous class in instead:

```php
$app->get('/ping', new class implements ServerMiddlewareInterface {
    {
        return new JsonResponse(['ack' => time()]);
    }
}, 'ping');
```

But that's far more verbose and kludgy to write.

moderndeveloperllc: Thanks. Definitely a bigger use case than I first thought. I imagine a `migration/to-v2.md` or similar doc is coming soon to help explain all the changes and possibilities.

src/Application.php
```diff
@@ -472,8 +309,14 @@ public function route($path, $middleware = null, array $methods = null, $name =
         $this->checkForDuplicateRoute($path, $methods);
 
         if (! isset($route)) {
-            $methods = (null === $methods) ? Router\Route::HTTP_METHOD_ANY : $methods;
-            $route   = new Router\Route($path, $middleware, $methods, $name);
+            $methods    = (null === $methods) ? Router\Route::HTTP_METHOD_ANY : $methods;
```
webimpress: Useless parentheses around `null === $methods`.
Is that right that in case of empty array we will have an empty array there not `Router\Route::HTTP_METHOD_ANY`?

src/MarshalMiddlewareTrait.php
```diff
@@ -35,52 +38,55 @@
...
-                    ? get_class($middleware) . "[Object]"
-                    : gettype($middleware) . '[Scalar]')
-                )
-            );
-        }
-
-        return $callable;
+        throw new Exception\InvalidMiddlewareException(sprintf(
+            'Unable to resolve middleware "%s" to a callable or %s',
+            (is_object($middleware) ? get_class($middleware) . "[Object]" : gettype($middleware) . '[Scalar]'),
```
webimpress: Useless parentheses around the whole expression, and please change `"` to `'` around `[Object]`.

src/Middleware/ImplicitHeadMiddleware.php
```diff
@@ -1,13 +1,15 @@
 <?php
 /**
  * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
- * @copyright Copyright (c) 2016 Zend Technologies USA Inc. (http://www.zend.com)
+ * @copyright Copyright (c) 2017 Zend Technologies USA Inc. (http://www.zend.com)
```
webimpress: Here should be range `2016-2017` instead.

test/Router/IntegrationTest.php
```diff
@@ -0,0 +1,475 @@
...
+ * @copyright Copyright (c) 2015-2017 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   https://github.com/zendframework/zend-expressive/blob/master/LICENSE.md New BSD License
+ */
+
+namespace ZendTest\Expressive\Router;
+
+use Fig\Http\Message\RequestMethodInterface as RequestMethod;
+use Fig\Http\Message\StatusCodeInterface as StatusCode;
+use Interop\Http\ServerMiddleware\DelegateInterface;
+use PHPUnit_Framework_TestCase as TestCase;
```
webimpress: We can use `PHPUnit\Framework\TestCase` as it is in other classes.

src/MarshalMiddlewareTrait.php
```diff
@@ -1,27 +1,30 @@
...
 
 namespace Zend\Expressive;
 
 use Closure;
 use Interop\Container\ContainerInterface;
-use Interop\Http\Middleware\DelegateInterface;
-use Interop\Http\Middleware\ServerMiddlewareInterface;
-use ReflectionFunction;
-use ReflectionMethod;
+use Interop\Http\ServerMiddleware\DelegateInterface;
```
webimpress: Seems to be unused in this file.

src/MarshalMiddlewareTrait.php
```diff
@@ -1,27 +1,30 @@
...
+use Interop\Http\ServerMiddleware\MiddlewareInterface as ServerMiddlewareInterface;
+use Psr\Http\Message\ResponseInterface;
 use Zend\Stratigility\Delegate\CallableDelegateDecorator;
```
webimpress: It is not used, can be removed now.

src/MarshalMiddlewareTrait.php
```diff
@@ -1,27 +1,30 @@
...
 /**
  * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
- * @copyright Copyright (c) 2015-2016 Zend Technologies USA Inc. (http://www.zend.com)
+ * @copyright Copyright (c) 2015-2017 Zend Technologies USA Inc. (http://www.zend.com)
  * @license   https://github.com/zendframework/zend-expressive/blob/master/LICENSE.md New BSD License
  */
 
 namespace Zend\Expressive;
 
 use Closure;
```
webimpress: Unused, can be removed.

test/ApplicationTest.php
```diff
@@ -9,10 +9,12 @@
 
 use BadMethodCallException;
 use DomainException;
+use Fig\Http\Message\StatusCodeInterface as StatusCode;
 use Interop\Container\ContainerInterface;
+use Interop\Http\ServerMiddleware\DelegateInterface;
 use InvalidArgumentException;
 use Mockery;
-use PHPUnit_Framework_TestCase as TestCase;
+use PHPUnit\Framework\TestCase as TestCase;
```
webimpress: `as TestCase` is redundant here.

test/Container/ApplicationFactoryTest.php
```diff
@@ -178,9 +199,7 @@ public function testNoRoutesAreAddedIfSpecDoesNotProvidePathOrMiddleware()
 
         $app = $this->factory->__invoke($this->container->reveal());
 
-        $r = new ReflectionProperty($app, 'routes');
-        $r->setAccessible(true);
-        $routes = $r->getValue($app);
+        $routes = $app->getRoutes();
         $this->assertEquals(0, count($routes));
```
webimpress: `assertCount` could be used instead.

src/Application.php
```diff
@@ -639,33 +469,9 @@ private function checkForDuplicateRoute($path, $methods = null)
      */
     private function emitMarshalServerRequestException($exception)
     {
-        $response = (new Response())
+        $response = $this->responsePrototype
             ->withStatus(StatusCode::STATUS_BAD_REQUEST);
-        $finalHandler = $this->getFinalHandler();
-        $response = $finalHandler(new ServerRequest(), $response, $exception);
```
webimpress: Hm, now parameter `$exception` is never used in that method. Is it right?

test/Application/MarshalMiddlewareTraitTest.php
```diff
@@ -0,0 +1,298 @@
...
+        $second = 'second-middleware';
+        $third  = 'third-middleware';
+
+        // No container is passed to the method, so this will not matter
+        $this->container->has('second-middleware')->willReturn(true);
+        $this->container->has('third-middleware')->willReturn(true);
+
+        $base = [$first, $second, $third];
+
+        $this->setExpectedException(InvalidMiddlewareException::class, 'second-middleware');
```
webimpress: As I can see we are using PHPUnit 5.7, so here we should use:
`expectException` and `expectExceptionMessage` instead of deprecated method `setExpectedException`.

test/Application/MarshalMiddlewareTraitTest.php
```diff
@@ -0,0 +1,298 @@
...
+        $this->assertInstanceOf(CallableMiddlewareWrapper::class, $middleware);
+        $this->assertAttributeInstanceOf(TestAsset\CallableMiddleware::class, 'middleware', $middleware);
+    }
+
+    public function testPreparingInvalidInvokableMiddlewareRaisesException()
+    {
+        $base = stdClass::class;
+        $this->container->has(stdClass::class)->willReturn(false);
+
+        $this->setExpectedException(InvalidMiddlewareException::class, 'invalid; neither invokable');
```
webimpress: Please use `expectException` and `expectExceptionMessage` instead.

test/Application/MarshalMiddlewareTraitTest.php
```diff
@@ -0,0 +1,298 @@
...
+            'non-callable-object'     => [new stdClass(), $defaultExpectedMessage],
+        ];
+    }
+
+    /**
+     * @dataProvider invalidMiddlewareTypes
+     */
+    public function testPreparingUnknownMiddlewareTypeRaisesException($invalid, $expectedMessage)
+    {
+        $this->setExpectedException(InvalidMiddlewareException::class, $expectedMessage);
```
webimpress: Please use `expectException` and `expectExceptionMessage` instead.

test/Container/ApplicationFactoryTest.php
```diff
@@ -265,7 +271,7 @@ public function testMiddlewareIsNotAddedIfSpecIsInvalid()
 
         $this->injectServiceInContainer($this->container, 'config', $config);
 
-        $this->setExpectedException(ContainerException\InvalidArgumentException::class, 'pipeline');
+        $this->setExpectedException(ExpressiveException\InvalidArgumentException::class, 'pipeline');
```
webimpress: Please use `expectException` and `expectExceptionMessage` instead.

test/Middleware/LazyLoadingMiddlewareTest.php
```diff
@@ -0,0 +1,120 @@
...
+
+    /**
+     * @dataProvider invalidMiddleware
+     */
+    public function testRaisesExceptionIfMiddlewarePulledFromContainerIsInvalid($middleware)
+    {
+        $this->container->get('middleware')->willReturn($middleware);
+        $lazyLoadingMiddleware = $this->buildLazyLoadingMiddleware('middleware');
+
+        $this->setExpectedException(InvalidMiddlewareException::class);
```
webimpress: Please use `expectException` and `expectExceptionMessage` instead.

src/Application.php
```diff
@@ -19,9 +19,10 @@
 use Zend\Diactoros\Response\SapiEmitter;
 use Zend\Diactoros\ServerRequest;
 use Zend\Diactoros\ServerRequestFactory;
-use Zend\Stratigility\FinalHandler;
+use Zend\Stratigility\Delegate\CallableDelegateDecorator;
 use Zend\Stratigility\Http\Response as StratigilityResponse;
```
webimpress: It seems to be not used (even more, not defined) any more.

src/Middleware/LazyLoadingMiddleware.php
weierophinney: There's already a document, but I'll be updating it further. One thing I want to do as well is create a 1.1 version that lists deprecations, which we can note in that same document.

src/Application.php
weierophinney: Good catch; thanks!

src/Application.php
weierophinney: Yes; empty array means "responds to only implicit methods". It was due to a feature request somewhere along the line.

src/Application.php
weierophinney: Um, correct, but that line is _removed_.

src/Middleware/ImplicitHeadMiddleware.php
weierophinney: First publication in a _release_  is this year.

weierophinney:
@webimpress Addressed all feedback you provided; thanks!

src/Application.php
webimpress: Yes, so maybe I've placed comment on wrong place, please see on signature of method `emitMarshalServerRequestException` - there is parameter `$exception, which now is never used, because above line is removed. We should then remove this parameter in the method signature and all invocations all do something with the exception here.

weierophinney: I went a step further, and updated the method as follows:

- Checks to see if a container is present, and if it contains a `Zend\Expressive\Middleware\ErrorResponseGenerator` service; if not, it uses the previous logic.
- If it does, it pulls the service.
- It then invokes it, using the exception, an empty request, and the response prototype, in order to obtain a response.
- Logic then proceeds as normal for emitting the response.

This will provide a better error page for users. If they are using the whoops-based error generator, it will generate a whoops response with the exception details. If they are using the templated error generator, it will generate a templated response, with exception details if the debug flag is enabled.

Thanks for the heads-up and clarifying what you were flagging!

src/Application.php
```diff
@@ -639,33 +468,15 @@ private function checkForDuplicateRoute($path, $methods = null)
      */
     private function emitMarshalServerRequestException($exception)
     {
-        $response = (new Response())
-            ->withStatus(StatusCode::STATUS_BAD_REQUEST);
-        $finalHandler = $this->getFinalHandler();
-        $response = $finalHandler(new ServerRequest(), $response, $exception);
+        if ($this->container && $this->container->has(Middleware\ErrorResponseGenerator::class)) {
```
moderndeveloperllc: I have to admit I'm not a huge fan of this. It's basically bypassing any `ErrorHandler` class pipeline that is already in the system - specifically the hooks into listeners. For instance, my db and APM error logging is done through an `ErrorLoggerListener` class that is attached to the `ErrorHandler` via a factory.

weierophinney: This is outside that system - this method is specifically invoked within `run()` only if an exception occurs when attempting to create a `ServerRequest` via the `ServerRequestFactory`; the pipeline is never executed in this situation.

moderndeveloperllc: I understand that this is before the stack is formed, I was just hoping to not have two sets of error handling - one for "startup" exceptions, and one for when the stack is run. No listeners means no custom logging, notifications, etc.

What about something like:
```
if ($this->container && $this->container->has(Middleware\ErrorResponseGenerator::class)) {
    //This is factoried already. Can be modified by user to include listeners, etc.
    $errorHandler = $this->container->get(Middleware\ErrorHandler::class);
    $response =  $errorHandler->handleStartupException(
        $exception, 
        new ServerRequest());
} else {
    $response = $this->responsePrototype->withStatus(StatusCode::STATUS_BAD_REQUEST);
}
```

With the new function in `Zend\Stratigility\Middleware\ ErrorHandler` being:
```
public function handleStartupException($e, $request)
{
    set_error_handler($this->createErrorHandler());
    $response = $this->handleThrowable($e, $request);
    restore_error_handler();

    return $response;
}
```

So we are constructing the `ErrorHandler`, but bypassing the pipeline `process()` because we already know there is an Exception. This would allow the generator and listeners to do their things and return a response to the emitter while keeping everything for both startup and process errors using the same code.

weierophinney: I understand where you're coming from, but I still disagree.

`Zend\Stratigility\Middleware\ErrorHandler` is _middleware_, and its public methods are only around:

- attaching listeners
- handling requests

To make it behave as you suggest would change the purpose of the `ErrorHandler` class: it is intended to be _middleware_, allowing developers to replace the service with _other middleware_. If we make the API change you suggest, we'd effectively be indicating that the class must implement _some other interface_, one that defines the `handleStartupException()` method. It's supposed to be _just middleware_.

My point in re-using the `ErrorResponseGenerator` is that it has a defined interface: a callable accepting the exception, a request, and a response, and returns a response. It's an instance called *by* the `ErrorHandler`, and, because it's a service, can be re-used elsewhere, which is what I've done here.

What that means is that you could provide your own ErrorResponseGenerator implementation entirely if you feel the need to handle startup exceptions due to inability to marshal a request in the same way as other application exceptions. In such a case, I'd likely attach my listeners to *that* instance, instead of the `ErrorHandler`, as the `ErrorResponseGenerator` could then notify listeners instead.

I personally feel that this is an edge case; there are very few situations where this should happen, and the only ones I can think of off the top of my head are when using async systems, which have to marshal the request information manually instead of relying on the SAPI. As such, the liklihood that you'd ever need to log such an error is pretty slim.

moderndeveloperllc: OK. I see where you are coming from. I think focusing on the listener logic made me overlook the fact that `ErrorHandler` is in fact just middleware and should not be treated like a service. I agree that this is a very edge case that is much like an `Exception` being thrown because the container has an error, missing config, etc. I can modify my already custom `ErrorResponseGenerator` to replicate the `ErrorHandler` listener logic and attach to there instead of onto the middleware.

Perhaps at a later date it can be discussed if the listener logic should be moved to `ErrorResponseGenerator` in Stratigility to help bring `ErrorHandler` to a more "pure" middleware class without ancillary methods.

composer.json
```diff
@@ -12,24 +12,25 @@
...
         "fig/http-message-util": "^1.1"
     },
     "require-dev": {
         "filp/whoops": "^1.1 || ^2.0",
         "malukenho/docheader": "^0.1.5",
         "mockery/mockery": "^0.9.5",
         "phpunit/phpunit": "^5.7.1",
+        "phpspec/prophecy": "^1.6",
```
webimpress: Maybe better will be just update PHPUnit dependency in the library to the latest version (^5.7.9) and remove `phpspec/prophecy` here?
https://github.com/sebastianbergmann/phpunit/blob/5.7/composer.json#L29

weierophinney: Awesome; wasn't aware what version would pick up the prophecy 1.6 release. Updated and pushing!

src/Container/ApplicationFactory.php
xtreamwayz: I just reviewed the 1.1.0 deprecation changes. I guess that would make this PR the expressive 2.0.0 version. That would mean stuff like this can be removed directly if 1.1.0 deprecates this stuff.

src/Container/TemplatedErrorHandlerFactory.php
```diff
@@ -38,6 +38,15 @@
  * </code>
  *
  * If any of the keys are missing, default values will be used.
+ *
+ * @deprecated Since 2.0, to remove in 3.0. Please migrate your code to use
```
xtreamwayz: Can we deprecate this in version 1.1.0 and remove it in 2.0?

src/Container/WhoopsErrorHandlerFactory.php
```diff
@@ -60,6 +60,15 @@
  * </code>
  *
  * All values are booleans; omission of any implies boolean false.
+ *
+ * @deprecated Since 2.0, to remove in 3.0. Please migrate your code to use
```
xtreamwayz: Same as before: Can we deprecate this in version 1.1.0 and remove it in 2.0?

src/IsCallableInteropMiddlewareTrait.php
```diff
@@ -0,0 +1,56 @@
...
+     */
+    private function reflectMiddleware(callable $middleware)
+    {
+        if (is_array($middleware)) {
+            $class = array_shift($middleware);
+            $method = array_shift($middleware);
+            return new ReflectionMethod($class, $method);
+        }
+
+        if ($middleware instanceof Closure || ! is_object($middleware)) {
```
xtreamwayz: Missing import: `use Closure;`

src/TemplatedErrorHandler.php
```diff
@@ -19,6 +19,15 @@
  * Provides the optional ability to render a template for each of 404 and
  * general error conditions. If no template renderer is provided, returns
  * empty responses with appropriate status codes.
+ *
+ * @deprecated Since 2.0, to remove in 3.0. Please migrate your code to use
```
xtreamwayz: Same as before, deprecate in 1.1.0 and remove in 2.0.0.

src/WhoopsErrorHandler.php
```diff
@@ -20,6 +20,14 @@
  * interface for navigating an exception stack trace.
  *
  * @see http://filp.github.io/whoops/
+ * @deprecated Since 2.0, to remove in 3.0. Please migrate your code to use
```
xtreamwayz: Same as before, deprecate in 1.1.0 and remove in 2.0.0.

src/Container/ApplicationFactory.php
weierophinney: We still need a "final handler" of some sort for cases where the internal queue is exhausted, which is why I marshal the `NoopFinalHandler` here. What needs to happen is documenting the use case; it should likely either return a 404, a 400, or a 500 response, and how it does so will vary from application to application. Making a not of that now.

src/Container/TemplatedErrorHandlerFactory.php
weierophinney: Hmm... This is tricky. As noted in another reply, we still need a final handler of some sort, and this is one way to provide it; the problem is that the implementation is largely irrelevant due to the fact that the `$e`/`$error` argument to it will never be provided, and the `$response` will almost always be the original response or an empty response, due to the way the delegators are propagated. In other words, this factory, and the implementation it provides, is largely useless.

I'm just not sure how to message its deprecation.

Let me think on that some more before making a decision.

weierophinney: Okay, I came up with the following: the default delegate which I proposed in this patch needs to be a `DelegateInterface`, and should likely return a 404 response. As such, I made the following changes:

- Modified the constructor to allow null, callable, or `DelegateInterface` instances for the default delegate, and validate what I receive, decorating callables with `CallableDelegateWrapper`.
- Created `Zend\Expressive\Delegate\NotFoundDelegate`, which provides a 404 response. I also changed `Zend\Expressive\Middleware\NotFoundHandler` to act as a middleware proxy to the delegate, to allow re-using the logic between the two.
- Modified `getDefaultDelegate()` to pull the service `Zend\Expressive\Delegate\DefaultDelegate` from the container, if the container and the service exist; otherwise, it creates an instance of `NotFoundDelegate` using the current response prototype (but no attempt to retrieve a renderer).
- Updated `run()` to simply call `getDefaultDelegate()`, with no additional logic.
- Updated `ApplicationFactory` to pull the `DefaultDelegate` service if present, using a `null` value otherwise, when creating the `$defaultDelegate` argument for the `Application` instance.

Once those were in place, I removed each of these legacy final handlers. I'll update the deprecations PR for 1.1 to reflect that.

Thanks for the feedback, @xtreamwayz !

src/Delegate/NotFoundDelegate.php
```diff
@@ -0,0 +1,103 @@
...
+    ) {
+        $this->responsePrototype = $responsePrototype;
+        $this->renderer          = $renderer;
+        $this->template          = $template;
+    }
+
+    /**
+     * Creates and returns a 404 response.
+     *
+     * @param ServerRequestInterface $request  Ignored.
```
webimpress: Only one space before param description please.
And not sure if the description is clear enough? Is `$request` ignored? I can see it pass thru, so maybe it's not (?)

src/MarshalMiddlewareTrait.php
```diff
@@ -35,52 +35,55 @@
...
-                    ? get_class($middleware) . "[Object]"
-                    : gettype($middleware) . '[Scalar]')
-                )
-            );
-        }
-
-        return $callable;
+        throw new Exception\InvalidMiddlewareException(sprintf(
+            'Unable to resolve middleware "%s" to a callable or %s',
+            is_object($middleware) ? get_class($middleware) . "[Object]" : gettype($middleware) . '[Scalar]',
```
malukenho: Single quotes around `[Object]`

PR #429, opened 2017-01-26
Mark deprecated functionality

weierophinney:
This patch marks a number of methods and classes as deprecated for an upcoming 1.1.0 version, based on features proposed in #428. These include:

- `Application::pipeErrorHandler()`
- `Application::routeMiddleware()`
- `Application::dispatchMiddleware()`
- `Application::getFinalHandler()` (this patch provides `getDefaultDelegate()` as a forwards-compatibility measure)
- `Container\InvalidArgumentException`
- `Container\NotFoundException`
- `Container\TemplatedErrorHandlerFactory`
- `Container\WhoopsErrorHandlerFactory`
- `ErrorMiddlewarePipe`
- `TemplatedErrorHandler`
- `WhoopsErrorHandler`

Do *not* merge until #428 is accepted to develop.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -335,6 +335,13 @@ class Application extends MiddlewarePipe implements Router\RouteResultSubjectInt
      * Once middleware detection and wrapping (if necessary) is complete,
      * proxies to pipe().
      *
+     * @deprecated since 1.1.0; to remove in 2.0.0. Stratigility v1-style
+     *     "error middleware" (middleware with four arguments, the first of which
+     *     being an error) was deprecated in Stratigility 1.3, and support
+     *     removed in Stratigility 2.0. You can start using standard middleware
+     *     for error handling by calling `raiseThrowables()` on the `Application`
+     *     instance, and piping such middleware in an outer layer of your application.
+     *     For an example, see `Zend\Stratigility\Middleware\ErrorHandler`.
      * @param string|callable $path Either a URI path prefix, or middleware.
      * @param null|string|callable $middleware Middleware
      * @return self
@@ -410,6 +417,11 @@ class Application extends MiddlewarePipe implements Router\RouteResultSubjectInt
      * Allow header, and `$next()` is called with its `$error` argument set
      * to the value `405` (invoking the next error middleware).
      *
+     * @deprecated since 1.1.0; to remove in 2.0.0. This method is extracted to
+     *     its own dedicated middleware class starting in 2.0.0. If you are
+     *     manually piping this method into your application or middleware
+     *     pipelines, or overriding the method, you will need to update your
+     *     code when migrating to 2.0.
      * @param  ServerRequestInterface $request
      * @param  ResponseInterface $response
      * @param  callable $next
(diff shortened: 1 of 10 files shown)


src/Application.php
```diff
@@ -335,6 +335,13 @@ public function pipe($path, $middleware = null)
      * Once middleware detection and wrapping (if necessary) is complete,
      * proxies to pipe().
      *
+     * @deprecated since 1.1.0; to remove in 2.0.0. Stratigility v1-style
+     *     "error middleware" (middleware with four arguments, the first of which
+     *     being an error) was deprecated in Stratigility 1.3, and support
+     *     removed in Stratigility 2.0. You can start using standard middleware
+     *     for error handling by calling `raiseThrowables()` on the `Application`
+     *     instance, and piping such middleware in an outer layer of your application.
+     *     For an example, see `Zend\Stratigility\Middleware\ErrorHandler`.
```
malukenho: `{@see Zend\Stratigility\Middleware\ErrorHandler}` ?

src/Application.php
```diff
@@ -619,6 +639,20 @@ public function getFinalHandler(ResponseInterface $response = null)
...
     /**
+     * Return the default delegate to use during `run()` when the stack is exhausted.
+     *
+     * Provided as a forwards compatibility measure with version 2, and proxies
+     * to getFinalHandler(); please note that it does not accept any arguments,
+     * nor pass them on to that method.
+     *
+     * @return callable|null
+     */
+    public function getDefaultDelegate()
```
danizord: Why do we need to expose a getter for final handler after all?

weierophinney: It's used internally to lazy-load it when required and none was injected during instantiation, which makes it a canonical source for locating it if the service is not defined.

danizord: I see, but it does not need to be public, no?

weierophinney: Again, it becomes the canonical source for the service if no service is registered in the container. As such, if you need access to it for any reason (e.g., to re-use it elsewhere), it would need to be public.

danizord: mhmm, sorry but it's not clear to me why would someone want to re-use it outside of `Application` context. Any example? :thinking:

weierophinney: Sure: you could be nesting an application within another, and want to re-use the default delegate from the parent application. As such, you might build the factory like this:

```php
use Zend\Expressive\AppFactory;
use Zend\Expressive\Router\FastRouteRouter;

function ($container)
{
    $app = $container->get(Application::class);
    $nested = new Application(
        new FastRouteRouter(),
        $container,
        $app->getDefaultDelegate()
    );
    
    /* pipe and/or route some middleware */
    $nested->pipe(/* ... */);
    $nested->get('/subpath', /* ... */);
    
    return $nested;
}
```

In normal circumstances, if this were dispatched, if a call to `$next()`/`$delegate->process()` occurs but the pipeline is exhausted, this would return handling to the parent, which might lead to execution later down the stack. By passing the default delegate in, we ensure that this nested application then becomes terminal; once it is dispatched, it _must_ return a response, ensuring no deeper layers are executed.

danizord: Haa, thanks for clarifying! These fallback services would be a lot easier if Expressive could provide its default container config that people could override, but unfortunately it would not work with all container interop implementations. :confused:

PR #432, opened 2017-01-30
Backport forwards-compatibility changes for 1.1.0 release

weierophinney:
This patch backports a number of changes from the current develop branch (targeted at version 2.0.0) for purposes of a 1.1.0 release. The changes include:

- Addition of the ability to configure the final handler instance when creating an `Application` instance (#309)
- Removal of the `RouteResultObserver` features (#406)
- Interception of `ServerRequestFactory` exceptions for purposes of reporting (#373)
- Addition of the `programmatic_pipelines` configuration flag
- Addition of the `raise_throwables` configuration flag

TODO
----

- [x] Document installing `zendframework/zend-expressive-tooling` for purposes of detecting `getOriginal*()` usage and error middleware usage/definition.
- [x] Document the `programmatic_pipelines` flag
- [x] Document the `raise_throwables` flag
- [x] Document how to effectively enable programmatic pipelines in Expressive 1.1, using either Stratigility middleware or custom middleware.


src/Application.php
```diff
@@ -8,12 +8,15 @@
 namespace Zend\Expressive;
 
 use Interop\Container\ContainerInterface;
+use InvalidArgumentException;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
+use UnexpectedValueException;
 use Zend\Diactoros\Request;
```
xtreamwayz: Not being used: Zend\Diactoros\Request

doc/book/reference/migration/to-v1-1.md
```diff
@@ -0,0 +1,688 @@
...
+  take a `$path` argument. (If one argument is present, it is assumed to be
+  middleware; with two arguments, the first argument is the `$path`.) Paths are
+  literal URI path segments. If the incoming request matches that segment, the
+  middleware will execute; otherwise, it will not. These can be used to provide
+  sub-applications with their own routing.
+
+- `pipeRoutingMiddleware()` is used to pipe the internal routing middleware into
+  the pipeline.
+
+- `pipeDispatchMiddleware()` is used to pipe the internal dispatch middleware into
```
danizord: Not internal anymore :)

test/ApplicationMarshalMiddlewarePipeTest.php
```diff
@@ -0,0 +1,109 @@
...
+            $this->createPassthroughMiddleware(),
+            $this->createExceptionMiddleware('exceptional'),
+            $this->createTerminalMiddleware(),
+            $this->createErrorMiddleware(),
+        ]);
+
+        $app->run($request, $response);
+    }
+
+    public function createErrorHandler($expected)
```
webimpress: It doesn't matter to much, but I would use it as `private`/`protected` method and keep only test methods `public`.
The same also about few more helpers below.

test/ApplicationMarshalMiddlewarePipeTest.php
```diff
@@ -0,0 +1,109 @@
...
+    {
+        return function ($request, $response, $next) use ($expected) {
+            try {
+                $response = $next($request, $response);
+                return $response;
+            } catch (\Throwable $e) {
+                // fall-through
+            } catch (\Exception $e) {
```
webimpress: Usually we import everything, don't we need then import `Throwable` and `Exception` here and use without `\` here?

test/Container/ApplicationFactoryTest.php
```diff
@@ -1296,4 +914,141 @@ public function testProperlyRegistersNestedErrorMiddlewareAsLazyErrorMiddleware(
...
+        $noPath = clone $api;
+        $goodbye = clone $api;
+        $pipelineFirst = clone $api;
+        $hello = clone $api;
+        $pipelineLast = clone $api;
+
+        $config = [
+            'middleware_pipeline' => [
+                [ 'path' => '/api', 'middleware' => $api ],
+                [ 'path' => '/dynamic-path', 'middleware' => 'DynamicPath' ],
```
webimpress: Please remove spaces after `[` and before `]`.

test/Container/ApplicationFactoryTest.php
```diff
@@ -1296,4 +914,141 @@ public function testProperlyRegistersNestedErrorMiddlewareAsLazyErrorMiddleware(
         $this->assertTrue($r->isClosure(), 'Configured middleware is not the expected lazy-middleware closure');
         $this->assertEquals(4, $r->getNumberOfParameters(), 'Configured middleware is not error middleware');
     }
+
+    /**
+     * @group programmatic
+     */
+    public function testWillNotInjectConfiguredRoutesOrPipelineIfProgrammaticPipelineFlagEnabled()
+    {
+        // @codingStandardsIgnoreStart
```
webimpress: Do we need it? I know that without it `}` must be in new line, but the same we have in line 1036.

test/RouteMiddlewareTest.php
```diff
@@ -621,100 +621,29 @@ public function testInjectsRouteResultAsAttribute()
...
-
         $app = $this->getApplication();
-        $this->assertAttributeNotContains($routeResultObserver->reveal(), 'routeResultObservers', $app);
+        $app->raiseThrowables();
 
-        $app->detachRouteResultObserver($routeResultObserver->reveal());
-        $this->assertAttributeNotContains($routeResultObserver->reveal(), 'routeResultObservers', $app);
+        $test = $app->routeMiddleware($request, $response, $next);
+        $this->assertInstanceOf(ResponseInterface::class, $test);
+        $this->assertEquals(405, $test->getStatusCode());
```
webimpress: Should we use here constant `Fig\Http\Message\StatusCodeInterface::STATUS_METHOD_NOT_ALLOWED`?

test/RouteMiddlewareTest.php
```diff
@@ -621,100 +621,29 @@ public function testInjectsRouteResultAsAttribute()
...
+        $allow = $test->getHeaderLine('Allow');
+        $this->assertContains('GET', $allow);
+        $this->assertContains('POST', $allow);
```
webimpress: Should we use here constants `Fig\Http\Message\RequestMethodInterface::GET` and `Fig\Http\Message\RequestMethodInterface::POST`?

test/ApplicationTest.php
```diff
@@ -621,4 +625,64 @@ public function testPipingNotInvokableErrorMiddlewareRisesException()
...
+                UnexpectedValueException::class,
+                'Unrecognized protocol version (foo-bar)',
+            ],
+        ];
+    }
+
+    /**
+     * @runInSeparateProcess
+     * @preserveGlobalState disabled
+     * @dataProvider invalidRequestExceptions
```
webimpress: Can we add params into PHPDocs?

```
@param string $expectedException
@param string $message
```

test/ApplicationTest.php
```diff
@@ -621,4 +625,64 @@ public function testPipingNotInvokableErrorMiddlewareRisesException()
...
+                $this->assertEquals(400, $response->getStatusCode());
+                $this->assertInstanceOf($expectedException, $err);
+                $this->assertEquals($message, $err->getMessage());
+                return $response;
+            };
+
+            $emitter = $this->prophesize(EmitterInterface::class);
+            $emitter->emit(Argument::that(function ($response) {
+                $this->assertInstanceOf(ResponseInterface::class, $response, 'Emitter did not receive a response');
+                $this->assertEquals(400, $response->getStatusCode());
```
webimpress: As above: should we use here const `Fig\Http\Message\StatusCodeInterface::STATUS_BAD_REQUEST`?

test/ApplicationTest.php
```diff
@@ -621,4 +625,64 @@ public function testPipingNotInvokableErrorMiddlewareRisesException()
...
+        try {
+            $requestFactory = Mockery::mock('alias:' . ServerRequestFactory::class)
+                ->shouldReceive('fromGlobals')
+                ->withNoArgs()
+                ->andThrow($expectedException, $message)
+                ->once()
+                ->getMock();
+
+            $finalHandler = function ($request, $response, $err = null) use ($expectedException, $message) {
+                $this->assertEquals(400, $response->getStatusCode());
```
webimpress: Should we use here const `Fig\Http\Message\StatusCodeInterface::STATUS_BAD_REQUEST`?

test/RouteMiddlewareTest.php
```diff
@@ -621,100 +621,29 @@ public function testInjectsRouteResultAsAttribute()
...
-            return $response;
-        };
-        $next = function ($request, $response, $err = null) {
-            $this->fail('Should not hit next');
-        };
-
         $request  = new ServerRequest();
         $response = new Response();
-        $result   = RouteResult::fromRouteMatch('resource', $middleware, $matches);
+        $result   = RouteResult::fromRouteFailure(['GET', 'POST']);
```
webimpress: Should we use here constants `Fig\Http\Message\RequestMethodInterface::GET` and `Fig\Http\Message\RequestMethodInterface::POST`?

composer.json
```diff
@@ -22,7 +22,7 @@
         "zendframework/zend-diactoros": "^1.1",
         "zendframework/zend-expressive-router": "^1.1",
         "zendframework/zend-expressive-template": "^1.0.1",
-        "zendframework/zend-stratigility": ">=1.1.0 < 1.3.0 || >=1.3.1 <2.0.0"
+        "zendframework/zend-stratigility": ">=1.2.0 < 1.3.0 || >=1.3.1 <2.0.0"
```
webimpress: Do we want still allow stratigility `1.2.0` here? Why not just from the latest version 1.3 (`1.3.3`)?
If still we want support `1.2` then maybe better starting from `1.2.1` with some fixes in compare to `1.2.0`?

webimpress: Just noticed that [the latest build](https://travis-ci.org/zendframework/zend-expressive/builds/197051503) is failing with lowest dependencies. It is because `raiseThrowables` method has beed added in Stratigility 1.3, so maybe we can just use `^1.3.3` in `composer.json` and it will resolve the issue?

doc/book/reference/migration/to-v1-1.md
weierophinney: In 1.1 (based off of current master) it still is; 2.0 (based off of current develop) makes it a separate class. ;-)

test/ApplicationMarshalMiddlewarePipeTest.php
```diff
@@ -0,0 +1,109 @@
+<?php
+/**
+ * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
+ * @copyright Copyright (c) 2016-2016 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   https://github.com/zendframework/zend-expressive/blob/master/LICENSE.md New BSD License
+ */
+
+namespace ZendTest\Expressive;
+
+use PHPUnit_Framework_TestCase as TestCase;
```
weierophinney: 2.0 release only at this point; I'd rather not duplicate the changes here, as it makes merging more difficult.

test/ApplicationMarshalMiddlewarePipeTest.php
weierophinney: Visibility really doesn't matter here, as these will never be extended. I'll leave as-is.

test/RouteMiddlewareTest.php
weierophinney: I *think* I've updated all of those on the develop branch. There's a LOT of them, and, as noted, I'd rather not duplicate effort here, in particular to prevent merge issues later.

composer.json
weierophinney: Per a conversation with you in IRC, I'll update this to `^1.3.3`, since we're now using `raiseThrowables()` explicitly within some of the updated code.

test/ApplicationTest.php
```diff
@@ -621,4 +625,66 @@ public function testPipingNotInvokableErrorMiddlewareRisesException()
...
+                return true;
+            }))->shouldBeCalled();
+
+            $app = new Application($this->router->reveal(), null, $finalHandler, $emitter->reveal());
+
+            $app->run();
+        } catch (\Throwable $e) {
+            $this->fail($e->getMessage());
+        } catch (\Exception $e) {
```
webimpress: Please import `Throwable` and `Exception`.

src/Application.php
```diff
@@ -694,4 +628,18 @@ private function swallowDeprecationNotices()
 
         $previous = set_error_handler($handler);
     }
+
+    /**
+     * @var \Exception|\Throwable $exception
```
webimpress: `@param` instead of `@var`, and maybe import `Exception` and `Throwable`?

test/ApplicationTest.php
weierophinney: The `Zend\Expressive\Exception` is already imported, so we need to have this one globally qualified. I'll import `Throwable`, though.

src/Application.php
weierophinney: `Exception` cannot be imported, as it would conflict with the `Zend\Expressive\Exception` namespace. I'll import `Throwable`, however.

PR #433, opened 2017-02-09
Package fixes and updates (for 2.0 release)

webimpress:
I've prepared some minor fixes and updates:

- run license header checks also in tests
- added license header checks to `composer check` script (so it will be checked also with every build on travis)
- updated license headers in tests
- allowed PHPUnit 6.0
- updated PHPDocs - added missing params/fixes/typos
- added private properties in tests
- CS fixes (removed redundant spaces)
- use constants (StatusCode, RequestMethod)
- if condition improvements
- removed ignoring cs checks sections + cs fixes
- updated docs
- updated `suggest` section in composer to reflect latest version of packages
- removed unused imports


doc/book/cookbook/using-routed-middleware-class-as-controller.md
```diff
@@ -201,10 +201,10 @@ class AlbumPage extends AbstractPage
 >         $action = $request->getAttribute('action', 'index') . 'Action';
 > 
 >         if (! method_exists($this, $action)) {
->             return $response->withStatus(404);
+>             return new EmptyResponse(404);
```
moderndeveloperllc: Want to go ahead and use 
`use Fig\Http\Message\StatusCodeInterface as StatusCode;`
and 
`StatusCode::STATUS_NOT_FOUND` for the 404? Yes it's just docs, but since the rest of the code uses it...

webimpress:
I've noted also that:

- NOTES.md is outdated (there is note about it on the top, maybe will be good to do something with it...)
- there is one "todo: remove in 2.0" https://github.com/zendframework/zend-expressive/blob/develop/test/IntegrationTest.php#L82-L85 - should it be removed and maybe also some other functionality then?

weierophinney:
@webimpress I'll remove `NOTES.md` and the test you reference during merge. Thanks!

weierophinney:
Thanks, @webimpress\!

Issue #434, opened 2017-02-09
1.1-dev dependencies


webimpress:
I was using dev-develop for a while, as it was prepared to 1.1 release, and yesterday everything get merged - master is now 1.1-dev and develop is now 2.0-dev. So I tried update my application to `dev-master` and I got collision in composer dependencies:

```
  Problem 1
    - Conclusion: don't install zendframework/zend-expressive-router 2.1.0
    - Conclusion: remove zendframework/zend-expressive-router 2.0.0
    - Installation request for zendframework/zend-expressive dev-master as 1.1.0 -> satisfiable by zendframework/zend-expressive[dev-master].
    - Installation request for zendframework/zend-expressive-helpers ^3.0.1 -> satisfiable by zendframework/zend-expressive-helpers[3.0.1].
    - zendframework/zend-expressive-fastroute 2.0.0 requires zendframework/zend-expressive-router ^2.0 -> satisfiable by zendframework/zend-expressive-router[2.0.0, 2.1.0].
    - zendframework/zend-expressive-fastroute 2.0.0 requires zendframework/zend-expressive-router ^2.0 -> satisfiable by zendframework/zend-expressive-router[2.0.0, 2.1.0].
    - Conclusion: don't install zendframework/zend-expressive-router 2.0.0
    - Installation request for zendframework/zend-expressive-fastroute ^2.0 -> satisfiable by zendframework/zend-expressive-fastroute[2.0.0].
```

Here are my `composer.json` dependencies:

```json
    "zendframework/zend-cache": "^2.7.2",
    "zendframework/zend-expressive": "dev-master as 1.1.0",
    "zendframework/zend-expressive-fastroute": "^2.0",
    "zendframework/zend-expressive-helpers": "^3.0.1",
    "zendframework/zend-hydrator": "^2.2.1",
    "zendframework/zend-i18n": "^2.7.3",
    "zendframework/zend-inputfilter": "^2.7.3",
    "zendframework/zend-serializer": "^2.8",
    "zendframework/zend-servicemanager": "^3.2",
    "zendframework/zend-stdlib": "^3.1",
    "zendframework/zend-validator": "^2.8.2"
```

And here is my current `composer show` (before change `dev-develop` to `dev-master`):
```
zendframework/zend-cache                2.7.2             provides a generic way to cache any data
zendframework/zend-code                 2.6.3             provides facilities to generate arbitrary code using an object oriented interface
[...]

PR #435, opened 2017-02-09
Allow expressive-router ^2.0 as well as ^1.1 (Expressive 1.1)

webimpress:
Please see #434.

It is possible to use version 2.0 of libraries:
- zend-expressive-aurarouter
- zend-expressive-fastroute
- zend-expressive-zendrouter


weierophinney:
Thanks, @webimpress\!

Issue #436, opened 2017-02-09
1.1 Implicit Methods Middleware are not available anymore


webimpress:
After yesterday merge now on master (1.1-dev) implicit methods middleware are no longer avaibale.

There is documentation on develop (2.0-dev) which says:
> Starting with version 1.1, Expressive offers middleware for implicitly supporting HEAD and OPTIONS requests. 
https://github.com/zendframework/zend-expressive/blob/develop/doc/book/features/middleware/implicit-methods-middleware.md

Is it possible to backport them also to 1.1 version or documentation of 2.0 will be updated and this feature is gonna be only in version 2?

Please see #413 (base PR with this functionality).

weierophinney:
We _could_ backport them; however, they are currently written to target http-middleware 0.4.1, which means they'd need to be rewritten to work with Expressive 1.1 — which is why I didn't do it in the first place.

I'd rather save the feature for 2.0, and have 1.1 focus on deprecations and preliminary support for pipelines. As such, I'm marking this as a documentation fix, and will flag it for the 2.0.0 milestone.

webimpress:
@weierophinney I've tried it before, and here is PR if you change your mind and want backport this feature https://github.com/zendframework/zend-expressive/pull/439

I've just cherry pick your first commits with this functionality.

xtreamwayz:
Just my 2 cents: Don't backport too much. I think it makes thing much more difficult and it delays a 1.1 release even more. Use 1.1 for the current merged features and deprecations, release it and leave all new features for 2.0.

weierophinney:
Decided not to do this for the 1.1.0 release, for reasons stated in #439 and here.

webimpress:
@weierophinney we need then documentation fix, please see:
https://github.com/zendframework/zend-expressive/blob/develop/doc/book/features/middleware/implicit-methods-middleware.md, this is why you labeled this issue ;)

PR #439, opened 2017-02-09
Hotfix #436 - backport Implicit Methods Middlewares

webimpress:


weierophinney:
I would prefer to keep this feature for 2.0. Having it backported means that if somebody extends the class, they will need to rewrite their extension for 2.0 due to signature changes.

PR #440, opened 2017-02-10
CS fixes (v2.0)

webimpress:
- fixed indents
- parenthesis on instantiating a new class
- fixed PHPDocs
- added type hinting

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -123,7 +123,7 @@ class Application extends MiddlewarePipe
      * @throws Exception\BadMethodCallException if the $method is not in $httpRouteMethods.
      * @throws Exception\BadMethodCallException if receiving more or less than 2 arguments.
      */
-    public function __call($method, $args)
+    public function __call($method, array $args)
     {
         if (! in_array(strtoupper($method), $this->httpRouteMethods, true)) {
             throw new Exception\BadMethodCallException('Unsupported method');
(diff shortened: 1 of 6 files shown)


src/Application.php
```diff
@@ -123,7 +123,7 @@ public function __construct(
      * @throws Exception\BadMethodCallException if the $method is not in $httpRouteMethods.
      * @throws Exception\BadMethodCallException if receiving more or less than 2 arguments.
      */
-    public function __call($method, $args)
+    public function __call($method, array $args)
```
webimpress: Please note this change. If someone override this method need to do the same on his extension, otherwise will get
> Strict Standards: Declaration ...::__call should be compatible with Application::__call($method, array $args)

weierophinney: This is an interesting case, as, technically, `__call()` is defined originally at the engine level, which allows the user to either include _or omit_ the typehint for the `$args` (or, per the documentation, `$arguments`) argument.

That said, once the method is _declared_ by a class, whatever signature it uses _must_ be used by any extending class.

In this particular case, while we _are_ bumping the major version, it's a bit of a WTF for upgraders if they are overriding the class to add other virtual methods. That said, it's easy enough to correct, as PHP provides the necessary information to update your own code. I'll detail it in the migration notes when merging.

weierophinney:
Thanks, @webimpress

PR #442, opened 2017-02-11
Respect JsonResponseHandler flags when false

danizord:
Currently, when you explicitly set the flags to false, it behaves as if it were true.

### DIFF
--- a/src/Container/WhoopsFactory.php
+++ b/src/Container/WhoopsFactory.php
@@ -75,11 +75,11 @@ class WhoopsFactory
 
         $handler = new JsonResponseHandler();
 
-        if (isset($config['json_exceptions']['show_trace'])) {
+        if (isset($config['json_exceptions']['show_trace']) && true === $config['json_exceptions']['show_trace']) {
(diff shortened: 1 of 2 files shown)


src/Container/WhoopsFactory.php
```diff
@@ -75,11 +75,11 @@ private function registerJsonHandler(Whoops $whoops, $config)
 
         $handler = new JsonResponseHandler();
 
-        if (isset($config['json_exceptions']['show_trace'])) {
+        if (isset($config['json_exceptions']['show_trace']) && true === $config['json_exceptions']['show_trace']) {
```
webimpress: Maybe we can have here `! empty`? What do you think?

danizord: Not sure, I think stricter is better :)

weierophinney: `! empty` is functionally equivalent here, more succinct, and more in line with other similar checks against configuration flags we're doing elsewhere.

weierophinney:
Thanks, @danizord\!

PR #445, opened 2017-02-14
update documentation on modular configurations

michaerum:
as configuration required by the ApplicationFactory is now strictly array in 1.1.x


weierophinney:
1.1.0 had a bug, as we were still supporting Aura.Di versions less than 3.1.0, which do _not_ allow array values for services. As such, a fix in #447 ensures we continue to support `ArrayObject` for `config` service values.

Thanks for taking the time to make this patch, though!

PR #450, opened 2017-02-15
Add PSR-11 support

xtreamwayz:
Replaces: 
- `Interop\Container\ContainerInterface` => `Psr\Container\ContainerInterface`
- `Interop\Container\Exception\ContainerException` => `Psr\Container\ContainerExceptionInterface`
- `Interop\Container\Exception\NotFoundException` => `Psr\Container\NotFoundExceptionInterface`

Updates all container-interop references in the docs.


composer.json
```diff
@@ -17,8 +17,8 @@
     },
     "require": {
         "php": "^5.6 || ^7.0",
-        "container-interop/container-interop": "^1.1",
         "psr/http-message": "^1.0",
+        "psr/container": "^1.0",
```
webimpress: Can we have here alphabetical order, so `psr/container` before `psr/http-message`, please?

webimpress:
I can see that tests fails also with lowest dependencies, maybe we should add:

```json
    "conflict": {
        "container-interop/container-interop": "<1.2"
    },
```
into `composer.json`?

xtreamwayz:
> Isn't this PR a BC Break? We have now typehints for different
classes/interfaces, so if somebody extends one of our classes needs to
update his source as well. Am I wrong?

I was thinking the same after I created the PR. I'll leave it as is for the
discussion. If needed I'll revise on develop.

> I can see that tests fails also with lowest dependencies, maybe we should
add...

That might fix the issue.

Op wo 15 feb. 2017 09:41 schreef Michał Bundyra <notifications@github.com>:

composer.json
weierophinney: We can make this fully BC by requiring:

- `"container-interop/container-interop": "^1.2"`
- `"psr/container": "^1.0"`

That way, if folks are still type-hinting on container-interop interfaces, their code will continue to work. We can message then that we'll be removing support for container-interop in the next major version (v3).

One reason we can't go full-on with psr/container quite yet is because none of the supported containers currently support it. Until there are versions that do, we'd actually make Expressive useless!

webimpress: Hm... we can't change namespaces in expressive 1.X to `Psr\Container ... ` it will be BC break.

I think we can do it with v2 (as it is not yet released), because container-interop 1.2 extends psr-container, so for example ServiceManager with container-interop 1.2 is actually PSR-11 compatible.

weierophinney: Alternately, we could have a `conflict` section:

```json
"require": {
    "psr/container": "^1.0"
},
"conflict: {
    "container-interop/container-interop": "<1.2.0"
}
```

Since the various container implementations we support all require container-interop currently, and version 1.2.0 of container-interop extends the various psr/container interfaces, then an update would grab that version, and the various typehints we use would work.

weierophinney: @xtreamwayz Can you re-target this PR at the develop branch, and add the `conflict` statement I've suggested above? If so, I'll give it a spin, and see how my various applications work with it.

webimpress: @weierophinney I was suggesting adding "conflict" in my comment below ;-)
https://github.com/zendframework/zend-expressive/pull/450#issuecomment-279949836

xtreamwayz: @webimpress Don't worry, I have seen it :)

webimpress: @xtreamwayz Please sort packages in alphabetical order as I requested in first comment. Thx!

xtreamwayz:
@weierophinney Rebased on develop. The docs need to be checked again.

weierophinney:
Labeled as BC break, as anybody extending the various classes where signatures have typehints against container-interop would break — this is, of course, also why it's put in the 2.0.0 milestone.

weierophinney:
Thanks, @xtreamwayz\!

PR #454, opened 2017-02-22
Fixes #453

### LINKED ISSUE #453: [Develop] Call to a member function process() on string
Earlier we can add a route as 

```
$router->addRoute(new Route('/welcome', 'Hkt\Psr7AssetExample\Action\WelcomeAction', ['GET'], 'hkt/psr7-asset-example:welcome'));
```

Please pay special attention to `WelcomeAction`. The only requirement was it should be got from the container. But after trying to upgrade I noticed this no longer is working. But we need to get the object from the container itself.

```
$router->addRoute(new Route('/welcome', $di->get('Hkt\Psr7AssetExample\Action\WelcomeAction'), ['GET'], 'hkt/psr7-asset-example:welcome'));
```

The error occurs in 

```php
class DispatchMiddleware implements ServerMiddlewareInterface
{   
    {
        $routeResult = $request->getAttribute(RouteResult::class, false);
        if (! $routeResult) {
            return $delegate->process($request);
        }
 
        $middleware = $routeResult->getMatchedMiddleware();
 
        // $middleware obtained from getMatchedMiddleware is a string .

        return $middleware->process($request, $delegate);
    }
}
```

harikt:
This will fix #453 , and I believe this will fix the one reported by @danizord  https://github.com/zendframework/zend-expressive/issues/453#issuecomment-281470012 .

I need to write some unit test, if someone have the time / energy please.

Else when I get sometime I will do the same.

Thank you

### DIFF
--- a/src/MarshalMiddlewareTrait.php
+++ b/src/MarshalMiddlewareTrait.php
@@ -52,7 +52,7 @@ trait MarshalMiddlewareTrait
         }
 
         if ($middleware === Application::DISPATCH_MIDDLEWARE) {
-            return new Middleware\DispatchMiddleware();
+            return new Middleware\DispatchMiddleware($container);
         }
 
         if ($middleware instanceof ServerMiddlewareInterface) {
--- a/src/Middleware/DispatchMiddleware.php
+++ b/src/Middleware/DispatchMiddleware.php
@@ -9,6 +9,7 @@ namespace Zend\Expressive\Middleware;
 
 use Interop\Http\ServerMiddleware\DelegateInterface;
 use Interop\Http\ServerMiddleware\MiddlewareInterface as ServerMiddlewareInterface;
+use Psr\Container\ContainerInterface;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use Zend\Expressive\Router\RouteResult;
@@ -27,6 +28,16 @@ use Zend\Expressive\Router\RouteResult;
 class DispatchMiddleware implements ServerMiddlewareInterface
 {
     /**
+     * @var null|ContainerInterface
+     */
+    private $container;
+
+    public function __construct(ContainerInterface $container = null)
+    {
+        $this->container = $container;
+    }
+
+    /**
      * @param ServerRequestInterface $request
      * @param DelegateInterface $delegate
      * @return ResponseInterface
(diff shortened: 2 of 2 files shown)


harikt:
I have not considered returning an array in this case. https://github.com/zendframework/zend-expressive/issues/453#issuecomment-281470012 .  I also wonder whether we really need the trait  `Zend\Expressive\MarshalMiddlewareTrait` which is not able to handle the same. The most of the things written probably need to be also in `Zend\Expressive\Middleware\DispatchMiddleware` .

xtreamwayz:
>  I also wonder whether we really need the trait Zend\Expressive\MarshalMiddlewareTrait

Exactly this. If you do this as you proposed, that trait could be removed and all it's logic could move into `DispatchMiddleware`.

It would fix both issues mentioned in #453. However I doubt this is the right way. Basically all dispatch logic that is split into the middleware and the trait in 2.0 is moved into the `DispatchMiddleware`.

I still have my doubts that using zend-expressive-router stand alone would be considered causing a bug since it's just a bridge specifically for expressive. But that's for @weierophinney to decide what direction to take.

weierophinney:
I would do this very differently, to be honest; I'd recommend using the `LazyLoadingMiddleware` instead, as it contains more and better logic for ensuring that dispatchable middleware is created.

As such, I'd argue for the following implementation:

```php
use Interop\Http\ServerMiddleware\DelegateInterface;
use Interop\Http\ServerMiddleware\MiddlewareInterface as ServerMiddlewareInterface;
use Zend\Expressive\Router\RouteResult;

/**
 * Default dispatch middleware.
 *
 * Uses the composed container to marshal the middleware matched during
 * routing, and then dispatches it.
 *
 * If no route result is present in the request, delegates to the next
 * middleware.
 *
 * @internal
 */
class DispatchMiddleware implements ServerMiddlewareInterface
{
    /**
     * @var ContainerInterface
     */
    private $container;

    /**
     * @var ResponseInterface
     */
    private $responsePrototype;

    public function __construct(ContainerInterface $container, ResponseInterface $responsePrototype)
    {
        $this->container = $container;
        $this->responsePrototype = $responsePrototype;
    }

    /**
     * @param ServerRequestInterface $request
     * @param DelegateInterface $delegate
     * @return ResponseInterface
     */
    {
        $routeResult = $request->getAttribute(RouteResult::class, false);
        if (! $routeResult) {
            return $delegate->process($request);
        }

        $middleware = $routeResult->getMatchedMiddleware();

        if (! $middleware instanceof ServerMiddlewareInterface) {
            $middleware = new LazyLoadingMiddleware(
                $this->container,
                $this->responsePrototype,
                $middleware
            );
        }

        return $middleware->process($request, $delegate);
    }
}
```

weierophinney:
Hmm... nevermind, the above won't work. As such, the best approach is to use the `MarshalMiddlewareTrait`. Let me see if I can get a competing patch ready with that in place.

weierophinney:
Superceded by #455

PR #455, opened 2017-02-22
Allow DispatchMiddleware to handle callables and service names

### LINKED ISSUE #453: [Develop] Call to a member function process() on string

weierophinney:
Because `Zend\Expressive\Router\Route` is lenient in the middleware it composes, we cannot assume that the middleware composed is valid http-interop middleware in the dispatcher. As such, this patch does the following:

- Updates `DispatchMiddleware` to compose a router, response prototype, and container, as all are necessary for preparing interop middleware shims.
- Updates `DispatchMiddleware::process` to check if the middleware received is http-interop compatible; if not, it delegates to `MarshalMiddlewareTrait::prepareMiddleware()` to cast it to interop middleware.

This patch supercedes the approach from #454, and fixes #453.

### DIFF
--- a/src/Middleware/DispatchMiddleware.php
+++ b/src/Middleware/DispatchMiddleware.php
@@ -9,23 +9,61 @@ namespace Zend\Expressive\Middleware;
 
 use Interop\Http\ServerMiddleware\DelegateInterface;
 use Interop\Http\ServerMiddleware\MiddlewareInterface as ServerMiddlewareInterface;
+use Psr\Container\ContainerInterface;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
+use Zend\Expressive\MarshalMiddlewareTrait;
 use Zend\Expressive\Router\RouteResult;
+use Zend\Expressive\Router\RouterInterface;
 
 /**
  * Default dispatch middleware.
  *
- * Uses the composed container to marshal the middleware matched during
- * routing, and then dispatches it.
+ * Checks for a composed route result in the request. If none is provided,
+ * delegates to the next middleware.
  *
- * If no route result is present in the request, delegates to the next
- * middleware.
+ * Otherwise, it pulls the middleware from the route result. If the middleware
+ * is not http-interop middleware, it uses the composed router, response
+ * prototype, and container to prepare it, via the
+ * `MarshalMiddlewareTrait::prepareMiddleware()` method. In each case, it then
+ * processes the middleware.
  *
  * @internal
  */
 class DispatchMiddleware implements ServerMiddlewareInterface
 {
+    use MarshalMiddlewareTrait;
+
+    /**
+     * @var ContainerInterface|null
+     */
+    private $container;
+
+    /**
+     * @var ResponseInterface
+     */
+    private $responsePrototype;
+
+    /**
+     * @var RouterInterface
+     */
+    private $router;
+
+    /**
+     * @param RouterInterface $router
+     * @param ResponseInterface $responsePrototype
+     * @param ContainerInterface|null $container
+     */
+    public function __construct(
+        RouterInterface $router,
+        ResponseInterface $responsePrototype,
+        ContainerInterface $container = null
+    ) {
+        $this->router = $router;
+        $this->responsePrototype = $responsePrototype;
+        $this->container = $container;
+    }
+
     /**
      * @param ServerRequestInterface $request
      * @param DelegateInterface $delegate
@@ -39,6 +77,15 @@ class DispatchMiddleware implements ServerMiddlewareInterface
         }
 
         $middleware = $routeResult->getMatchedMiddleware();
+
+        if (! $middleware instanceof ServerMiddlewareInterface) {
+            $middleware = $this->prepareMiddleware(
+                $middleware,
+                $this->router,
+                $this->responsePrototype,
+                $this->container
+            );
+        }
 
         return $middleware->process($request, $delegate);
     }
--- a/src/MarshalMiddlewareTrait.php
+++ b/src/MarshalMiddlewareTrait.php
@@ -52,7 +52,7 @@ trait MarshalMiddlewareTrait
         }
 
         if ($middleware === Application::DISPATCH_MIDDLEWARE) {
-            return new Middleware\DispatchMiddleware();
+            return new Middleware\DispatchMiddleware($router, $responsePrototype, $container);
         }
 
         if ($middleware instanceof ServerMiddlewareInterface) {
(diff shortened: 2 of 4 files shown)


weierophinney:
Ping @harikt and @danizord , as each of you indicated you'd observed the behavior...

harikt:
Hi @weierophinney ,

I have tested your PR over : 
https://github.com/harikt/psr7-asset-example-zendexpressive/pull/5/files#diff-866788e2c87aa76a70397dfc39ce9980R42 . 

If you pass a string this works. If you pass a callable, in my case it is an instance of `Aura\Di\Injection\LazyGet` , then it cannot handle it and returns 

```
Zend \ Stratigility \ Exception \ MissingResponseException
Last middleware executed did not return a response. Method: GET Path: / .Handler: Zend\Expressive\Middleware\DispatchMiddleware
```

So the question does a callable also should work ?

harikt:
The problem is over here 

```
        if ($this->isCallableInteropMiddleware($middleware)) {
            return new CallableInteropMiddlewareWrapper($middleware);
        }

        if (is_callable($middleware)) {
            return new CallableMiddlewareWrapper($middleware, $responsePrototype);
        }
```

If the order is flipped this should work. We probably need more test cases for MarshalTrait .

harikt:
Hm... Flipping the order also will not work. I understand why now.

If it is a callable it expects to pass the `$request, $this->responsePrototype, $delegate` or `$request` , but on the other hand what I have given is an instance of callable itself which don't need those.  I want to look if I can pass a different `DispatchMiddleware` to expressive.

Thank you.

src/Middleware/DispatchMiddleware.php
```diff
@@ -40,6 +78,15 @@ public function process(ServerRequestInterface $request, DelegateInterface $dele
 
         $middleware = $routeResult->getMatchedMiddleware();
 
+        if (! $middleware instanceof ServerMiddlewareInterface) {
+            $middleware = $this->prepareMiddleware(
+                $middleware,
+                $this->router,
+                $this->responsePrototype,
+                $this->container
+            );
```
xtreamwayz: By adding the `prepareMiddleware` method here, I guess it can be removed now from [Application::route](https://github.com/weierophinney/zend-expressive/blob/a14315c04713d5463920c9e0853857f5f46de829/src/Application.php#L314-L319)?

danizord:
@weierophinney it feels wrong to me that now we have this logic to support various types of middlewares spread across various places in the framework.

Now that we have a solid and interoperable `MiddlewareInterface`, I think we should typehint against it everywhere and handle the marshaling only in `Application` level, before reaching the router, so that the router, `DispatchMiddleware` and others can always expect a `MiddlewareInterface` instance. Here are some advantages of this approach:

* Fails early when routing or piping invalid middlewares (e.g. with invalid signature). Currently they are only validated at dispatch time, so the developers won't even see the error unless they run a request against that route.
* Makes the router component more simple and interoperable, so that it can be used by other http-interop middleware dispatchers.
* Simplifies maintenance and makes it more extensible by having only a single place to handle all the various supported signatures.

It would require another major version bump of router stuff, but I think it's worth it :)

harikt:
@danizord so you mean there is no need of lazy loading. Am I correct understanding you ?

danizord:
@harikt nope, we can still have lazy loading by wrapping middlewares with `LazyLoadingMiddleware`, the point is that this wrapping should be done by `Application` before adding the route, so that the Router component and any `RouteResult` consumers can always expect `MiddlewareInterface` instances.

src/Middleware/DispatchMiddleware.php
weierophinney: No. This call only gets executed if we do not have an http-interop/http-middleware instance. If `route()` was called (either directly, or indirectly via `__call()`), then this logic will be skipped.

Essentially, this is a fallback for if non-interop middleware was injected in a `Route` instance, which will only happen if you're creating those instances manually.

weierophinney:
@danizord I see your point, and it's my preference as well. However, I want to wait to do that until PSR-15 is formally accepted; until then, zend-expressive-router should continue to work with both single-pass and double-pass middleware, as the specification may still change. (This is less and less likely at this point, but not impossible.) This allows re-use of zend-expressive-router outside of Expressive itself for those who want to use the router interface in other stacks.

weierophinney:
@harikt 

> If you pass a callable, in my case it is an instance of Aura\Di\Injection\LazyGet or Aura\Di\Injection\LazyNew ( both are callables ) then it cannot handle it

This is to be expected. The callable composed in the `Route` instance is expected to be _middleware_, and not a _factory_ (which is what `LazyGet` and `LazyNew` provide). As such, I would not expect these to work, and the fact that they ever did (did they?) within Expressive is a fluke.

weierophinney:
> So the question is, Is there a way to pass DispatchMiddleware to expressive application ?

Absolutely!

So, first, a caveat: since this PR was around behavior on the develop branch, which is the future 2.0 release, I'll be discussing that specifically; what I detail may not be entirely true for the 1.X series.

By default, if you call `pipeDispatchMiddleware()`, internally the `MarshalMiddlewareTrait` creates an instance of `Zend\Expressive\Middleware\DispatchMiddleware` and pipes it into the pipeline. If you want to use your own dispatch middleware, you can simply choose not to call that method, and instead call `pipe()` with your own middleware. So, this:

```php
$app->pipeDispatchMiddleware();`
```

becomes:

```php
$app->pipe(My\Custom\DispatchMiddleware::class); // or a concrete instance, or a callable, whatever
```

Just make sure you do it after the routing middleware (which would be done via `pipeRoutingMiddleware()` typically, but could _also_ be custom middleware if you want).

You will typically want something like the following to begin with:

```php
$routeResult = $request->getAttribute(RouteResult::class, false);
if (! $routeResult) {
    return $delegate->process($request);
}

$middleware = $routeResult->getMatchedMiddleware();
```

What you do from there is up to you. :smile:

Good luck, @harikt !

PR #457, opened 2017-03-02
Cookbook entry: passing data between middleware

xtreamwayz:


doc/book/cookbook/passing-data-between-middleware.md
```diff
@@ -0,0 +1,110 @@
...
+        
+        // Step 3: Do something (with the response) before returning the response
+        
+        // Step 4: Return the response
+        return $response;
+    }
+}
+```
+
+In stead of passing data to other middleware you can also use the data in
```
edigu: In stead => Instead

doc/book/cookbook/passing-data-between-middleware.md
```diff
@@ -0,0 +1,110 @@
...
+        // Step 4: Return the response
+        return $response;
+    }
+}
+```
+
+Instead of passing data to other middleware you can also use the data in 
+actions. It's the same concept since an action class is middleware itself. In
+the `ExampleAction` the data is subtracted from the request and passed to the
+template renderer to create a HtmlResponse.
```
weierophinney: I'm not even sure we need to make a distinction between "actions" and "middleware"; as you note, they are the same thing. Perhaps we could just make this two separate examples, showing two middleware that use the passed data?

weierophinney:
Thanks, @xtreamwayz\!

doc/book/cookbook/passing-data-between-middleware.md
xtreamwayz: I'm not sure if beginners know they are the same. Showing both the Middleware and an Action might prevent questions.

weierophinney: I showed both, but did not draw attention to the "action" vs "middleware" verbiage.

PR #469, opened 2017-03-14
Adds Info for usage of expressive-pipeline-from-config

manuakasam:
I ran into the trouble that using the `expressive-pipeline-from-config` tool my routes haven't been found. That's because at some points the expressive-skeleton application didn't have the configuration for the routing- and dispatch-middlewares in their configs, as they were programmatically added.

This patch provides additional information that it will most likely be required to manually add them to your `middleware_pipeline` configuration.

### DIFF
--- a/doc/book/reference/migration/to-v2.md
+++ b/doc/book/reference/migration/to-v2.md
@@ -679,6 +679,12 @@ time, to allow you to test and verify your application first; however, due to
 the configuration in `config/autoload/programmatic-pipeline.global.php`, these
 are now ignored.
 
+Be aware, the default expressive-skeleton installation does not provide configuration
+for the dispatch- and routing-middleware. Make sure that you have registered the 
+`Zend\Expressive\Application::ROUTING_MIDDLEWARE` as well as 
+`Zend\Expressive\Application::DISPATCH_MIDDLEWARE` inside your `middleware_pipeline` 
+configuration.
+
 If you wish to use Whoops in your development environment, you may add the
 following to a local configuration file (e.g., `config/autoload/local.php`):
 


manuakasam:
I've just added this as a paragraph, however, it might be worth to add a full config example as well. Thoughts?

weierophinney:
@manuakasam I'm not sure I understand the problem you are running into.

At that point in the guide, you will have run `expressive-pipeline-from-config`, which creates `config/pipeline.php` for you, and this _does_ make the calls to `pipeRoutingMiddleware()` and `pipeDispatchMiddleware()` for you, _so long as you defined them in your configuration to begin with_ (see the [Generator class](https://github.com/zendframework/zend-expressive-tooling/blob/aedfba7e502dceed153d4094c65d972218bd6de6/src/GenerateProgrammaticPipelineFromConfig/Generator.php#L216-L236) for details).

If you did _not_ have these in your configuration, it does not add them. The assumption at that point is that you have your own custom router and/or dispatcher, and you're not using those provided by Expressive.

As such, I'm unclear what you're trying to clarify here.

manuakasam:
@weierophinney the issue I ran into is that the skeleton does not provide this configuration (in array form inside configs). When you run the config generator on a naked skeleton, the routing- and dispatch-middlewares are lost, as they aren't configured anywhere by default.

This is what I want to raise awareness for.

weierophinney:
@manuakasam What version of the skeleton are you running the tool on?

On v2, the pipeline is already programmatic.

On v1, we *did* configure the routing and dispatch middleware by default: https://github.com/zendframework/zend-expressive-skeleton/blob/887baee15c6e019f11427c6d517b095fb5be18ee/config/autoload/middleware-pipeline.global.php#L47-L59

Is it possible you removed that configuration in your application?

manuakasam:
@weierophinney Yes, the pipeline is programmatic and the dispatch and routing middlewares are within the default pipeline.php

However, if you then run the tool again (because you added more modules routes n stuff via modules configs), the initial pipeline will be replaced. Because the pipelines for route and dispatch are NOT configured anywhere, they are then lost.

Simple reproduction should be: simply do a new install of a skeletton. Instantly run the pipeline and route generator script. The pipeline.php will then no longer have the middlewares for routing and dispatching piped.

Am I making myself any clearer? If NOT, I suppose this could then be considered a local problem and I created a non-existing case on my end :P

weierophinney:
> However, if you then run the tool again

Honestly, this was never a use case we envisioned; the tool was meant for an initial migration, not for repeated use.

What we _could_ do is add some detection in the tool for existing `config/pipeline.php` and/or `config/routes.php` files, and have it raise an error early in those cases. If you think that would be useful, could you please raise an issue on the [zend-expressive-tooling repo](https://github.com/zendframework/zend-expressive-tooling/issues/new)?

PR #476, opened 2017-03-27
Add response status code for the Whoops error handler

### LINKED ISSUE #475: Whoops json error response uses status code 200
Hi,

I noticed that the Whoops json error response uses status code 200 when returning error.
Shouldn't it be 500 or something like that instead?

xtreamwayz:
fixes #475

### DIFF
--- a/test/Middleware/WhoopsErrorResponseGeneratorTest.php
+++ b/test/Middleware/WhoopsErrorResponseGeneratorTest.php
@@ -142,11 +142,32 @@ class WhoopsErrorResponseGeneratorTest extends TestCase
         $this->request->getParsedBody()->willReturn([]);
 
         $this->response->withHeader('Content-Type', 'application/json')->will([$this->response, 'reveal']);
-        $this->response->withStatus(StatusCode::STATUS_INTERNAL_SERVER_ERROR)->will([$this->response, 'reveal']);
-        $this->response->getStatusCode()->willReturn(StatusCode::STATUS_INTERNAL_SERVER_ERROR);
+        $this->response->withStatus(StatusCode::STATUS_IM_A_TEAPOT)->will([$this->response, 'reveal']);
+        $this->response->getStatusCode()->willReturn(StatusCode::STATUS_IM_A_TEAPOT);
         $this->response->getBody()->will([$this->stream, 'reveal']);
 
         $this->stream->write('error')->shouldBeCalled();
+
+        $generator = new WhoopsErrorResponseGenerator($this->whoops->reveal());
+
+        $this->assertSame(
+            $this->response->reveal(),
+            $generator($error, $this->request->reveal(), $this->response->reveal())
+        );
+    }
+
+    public function testNonErrorStatusCodeIsSetTo500()
+    {
+        $error = new RuntimeException();
+
+        $this->whoops->getHandlers()->willReturn([]);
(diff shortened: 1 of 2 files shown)


test/Middleware/WhoopsErrorResponseGeneratorTest.php
```diff
@@ -155,4 +155,25 @@ public function testJsonContentTypeResponseWithJsonResponseHandler()
...
+    public function testNonErrorStatusCodeIsSetTo500()
+    {
+        $error = new RuntimeException();
+
+        $this->whoops->getHandlers()->willReturn([]);
+        $this->whoops->handleException($error)->willReturn('WHOOPS');
+
+        $this->response->withStatus(StatusCode::STATUS_INTERNAL_SERVER_ERROR)->will([$this->response, 'reveal']);
+        $this->response->getBody()->will([$this->stream, 'reveal']);
+        $this->response->getStatusCode()->willReturn(StatusCode::STATUS_MOVED_PERMANENTLY);
```
weierophinney: wait... shouldn't this be `StatusCode::STATUS_INTERNAL_SERVER_ERROR` here?

test/Middleware/WhoopsErrorResponseGeneratorTest.php
```diff
@@ -142,8 +142,8 @@ public function testJsonContentTypeResponseWithJsonResponseHandler()
         $this->request->getParsedBody()->willReturn([]);
 
         $this->response->withHeader('Content-Type', 'application/json')->will([$this->response, 'reveal']);
-        $this->response->withStatus(StatusCode::STATUS_INTERNAL_SERVER_ERROR)->will([$this->response, 'reveal']);
-        $this->response->getStatusCode()->willReturn(StatusCode::STATUS_INTERNAL_SERVER_ERROR);
+        $this->response->withStatus(StatusCode::STATUS_IM_A_TEAPOT)->will([$this->response, 'reveal']);
+        $this->response->getStatusCode()->willReturn(StatusCode::STATUS_IM_A_TEAPOT);
```
weierophinney: Not sure I understand this change. `$error` here is a simple `RuntimeException`, so I'd expect an internal server error. Or are you redefining the `$error` somehow?

test/Middleware/WhoopsErrorResponseGeneratorTest.php
xtreamwayz: No, it's deliberately an error code below 400 to make sure `Utils::getStatusCode` does its job and returns the default `500` code. However I don't know if it should be tested here and just rely on Stratigility tests.

test/Middleware/WhoopsErrorResponseGeneratorTest.php
xtreamwayz: > Looks good, in general. I think we need an additional test for when $e has an exception code that falls elsewhere in the 400-599 range, to validate that the response status is set accordingly.

I might have misunderstood that... or you wanted to see changes to the default `RunTimeException` error code.

test/Middleware/WhoopsErrorResponseGeneratorTest.php
xtreamwayz: Uhm, it doesn't make sense since the default error code is 0. So it's already converted and tested in the previous tests.

PR #479, opened 2017-04-16
use Run|RunInterface @var for WhoopsErrorResponseGenerator::$whoops property

samsonasik:


samsonasik:
actually, whoops doesn't has `RunInterface`, so, I applied `Run|RunInterace` into @var

xtreamwayz:
Thanx @samsonasik

PR #480, opened 2017-04-26
Forwarded method flag in head request

snapshotpl:
`ImplicitHeadMiddleware` clears body, but we now we are not able to detect that is a HEAD request in final middleware. `forwarded_http_method` can be use to abandon prepare response body

### DIFF
--- a/src/Middleware/ImplicitHeadMiddleware.php
+++ b/src/Middleware/ImplicitHeadMiddleware.php
@@ -43,6 +43,8 @@ use Zend\Expressive\Router\RouteResult;
  */
 class ImplicitHeadMiddleware implements ServerMiddlewareInterface
 {
+    const FORWARDED_HTTP_METHOD_ATTRIBUTE = 'forwarded_http_method';
+
     /**
      * @var null|ResponseInterface
      */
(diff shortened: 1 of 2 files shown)


src/Middleware/ImplicitHeadMiddleware.php
```diff
@@ -89,7 +89,9 @@ public function process(ServerRequestInterface $request, DelegateInterface $dele
         }
 
         $response = $delegate->process(
-            $request->withMethod(RequestMethod::METHOD_GET)
+            $request
+                ->withMethod(RequestMethod::METHOD_GET)
+                ->withAttribute('forwarded_http_method', RequestMethod::METHOD_HEAD)
```
weierophinney: Create a constant for this attribute name, please, and then update the code to use that constant.

weierophinney: And, on second thought: shouldn't this be accomplished with either an `X-Forwarded-Proto` or `Forwarded` header (see [the MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto))?

snapshotpl: This headers should be using in specified cases but not to forward http methods.

weierophinney:
Thanks, @snapshotpl

weierophinney:
Merged to develop for release with 2.1.0.

PR #486, opened 2017-05-21
Update docs to reflect current code and recommendations

devgkz:
Actualize Application namespace to App
Replace string classnames to fully qualified names with `::class`


weierophinney:
Thanks, @devgkz

PR #489, opened 2017-05-26
fix(WhoopsError): ensure the SCRIPT_NAME is set before attempting to use it

gsdevme:
PSR states
```
     * Retrieves data related to the incoming request environment,
     * typically derived from PHP's $_SERVER superglobal. The data IS NOT
     * REQUIRED to originate from $_SERVER.
```
Nothing suggests that `SCRIPT_NAME` will always exist. Therefore I've added a `isset()` guard and left the default as an empty string.

### DIFF
--- a/src/Middleware/WhoopsErrorResponseGenerator.php
+++ b/src/Middleware/WhoopsErrorResponseGenerator.php
@@ -91,11 +91,18 @@ class WhoopsErrorResponseGenerator
     {
         $uri = $request->getAttribute('originalUri', false) ?: $request->getUri();
         $request = $request->getAttribute('originalRequest', false) ?: $request;
+        
+        $serverParams = $request->getServerParams();
+        $scriptName = '';
+        
+        if (isset($serverParams['SCRIPT_NAME'])) {
+            $scriptName = $serverParams['SCRIPT_NAME'];
+        }
 
         $handler->addDataTable('Expressive Application Request', [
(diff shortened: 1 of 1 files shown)


src/Middleware/WhoopsErrorResponseGenerator.php
```diff
@@ -91,11 +91,18 @@ private function prepareWhoopsHandler(ServerRequestInterface $request, PrettyPag
     {
         $uri = $request->getAttribute('originalUri', false) ?: $request->getUri();
         $request = $request->getAttribute('originalRequest', false) ?: $request;
+        
+        $serverParams = $request->getServerParams();
+        $scriptName = '';
+        
+        if (isset($serverParams['SCRIPT_NAME'])) {
+            $scriptName = $serverParams['SCRIPT_NAME'];
+        }
```
moderndeveloperllc: This can be cleaned up to:
```
$serverParams = $request->getServerParams();
$scriptName = (isset($serverParams['SCRIPT_NAME'])) ? $serverParams['SCRIPT_NAME'] : '';
```
It'll be nice when ZE is pinned to PHP 7.1 and we can use null coalesce operators...

gsdevme: Always found it bizarre that people think this is "cleaner". I didn't realise I had written really unwieldily code here. I guess the above is also using a ternary operator so it does at least make it consistent.

PR #493, opened 2017-06-08
Feature - extended CS sniffs

webimpress:
Please see: https://github.com/zendframework/zend-coding-standard/pull/1

### DIFF
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -5,4 +5,7 @@
     <!-- Paths to check -->
     <file>src</file>
     <file>test</file>
+
+    <exclude-pattern>test/TestAsset/</exclude-pattern>
+    <exclude-pattern>test/**/TestAsset/</exclude-pattern>
 </ruleset>
--- a/test/ApplicationTest.php
+++ b/test/ApplicationTest.php
@@ -9,9 +9,11 @@ declare(strict_types=1);
 
 namespace ZendTest\Expressive;
 
+use Generator;
 use PHPUnit\Framework\Assert;
 use PHPUnit\Framework\TestCase;
 use Prophecy\Argument;
+use Prophecy\Prophecy\ObjectProphecy;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Server\MiddlewareInterface;
@@ -25,9 +27,28 @@ use Zend\Stratigility\Middleware\PathMiddlewareDecorator;
 use Zend\Stratigility\MiddlewarePipe;
 use Zend\Stratigility\MiddlewarePipeInterface;
 
+use function array_unshift;
+use function sprintf;
+use function strtoupper;
+
 class ApplicationTest extends TestCase
 {
-    public function setUp()
+    /** @var MiddlewareFactory|ObjectProphecy */
+    private $factory;
+
+    /** @var MiddlewarePipeInterface|ObjectProphecy */
+    private $pipeline;
+
+    /** @var RouteMiddleware|ObjectProphecy */
+    private $routes;
+
+    /** @var RequestHandlerRunner|ObjectProphecy */
+    private $runner;
+
+    /** @var Application */
+    private $app;
+
+    protected function setUp()
     {
         $this->factory = $this->prophesize(MiddlewareFactory::class);
         $this->pipeline = $this->prophesize(MiddlewarePipeInterface::class);
@@ -42,7 +63,7 @@ class ApplicationTest extends TestCase
         );
     }
 
-    public function createMockMiddleware()
+    private function createMockMiddleware() : MiddlewareInterface
     {
         return $this->prophesize(MiddlewareInterface::class)->reveal();
     }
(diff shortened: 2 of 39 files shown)


phpcs.xml
```diff
@@ -2,6 +2,10 @@
 <ruleset name="Zend Framework Coding Standard">
     <rule ref="./vendor/zendframework/zend-coding-standard/ruleset.xml"/>
 
+    <rule ref="ZendCodingStandard.Commenting.FunctionComment.TypeHintMissing">
+        <severity>0</severity>
+    </rule>
```
webimpress: We disable this rule for the whole library right now. The issue is only in [`Emitter/EmitterStack.php`](https://github.com/zendframework/zend-expressive/blob/master/src/Emitter/EmitterStack.php) lines [`58`](https://github.com/zendframework/zend-expressive/blob/master/src/Emitter/EmitterStack.php#L54-L58), [`71`](https://github.com/zendframework/zend-expressive/blob/master/src/Emitter/EmitterStack.php#L67-L71), [`84`](https://github.com/zendframework/zend-expressive/blob/master/src/Emitter/EmitterStack.php#L80-L84). In PHPDocs we have said parameter `$emitter` should have type `EmitterInterface` but there is missing type hint.

Hm... Now I think maybe better will be just exclude only this file (`EmitterStack`) from this rule instead of disabling the rule for the whole library?

composer.json
```diff
@@ -30,7 +30,7 @@
         "malukenho/docheader": "^0.1.5",
         "mockery/mockery": "^0.9.5",
         "phpunit/phpunit": "^6.0.8 || ^5.7.15",
-        "zendframework/zend-coding-standard": "~1.0.0",
+        "zendframework/zend-coding-standard": "dev-feature/sniffs",
```
xtreamwayz: I don't think it's a good idea to rely on development branches in stable packages.

froschdesign: But a good way to push the topic! 😉

froschdesign:
We will wait for zendframework/zend-coding-standard#1 (I added the milestone 2.0.0)

webimpress:
The purpose of this PR was to demonstrate how many issues can detect using zend-coding-standard and the power of that tool. Many of these issues could be fixed automatically. Definitely I'd like to push forward first zend-coding-standard and than update all repositories. I've created this PR to test the sniffs and ofc it shouldn't be merged before zend-coding-standard 2.0 release.

composer.json
danizord: It's a development dependency so I don't see any issue :)

src/Container/ResponseFactory.php
```diff
@@ -21,16 +21,19 @@
...
-                . ' implementation instance. By default, we assume usage of'
-                . ' zend-diactoros for PSR-7, but it does not appear to be'
-                . ' present on your system. Please install zendframework/zend-diactoros'
-                . ' or provide an alternate factory for the %s service that'
-                . ' can produce an appropriate %s instance.',
+                    . ' implementation instance. By default, we assume usage of'
+                    . ' zend-diactoros for PSR-7, but it does not appear to be'
+                    . ' present on your system. Please install zendframework/zend-diactoros'
+                    . ' or provide an alternate factory for the %s service that'
+                    . ' can produce an appropriate %s instance.',
```
xtreamwayz: Shouldn't these lines start at the same point?

test/ApplicationTest.php
```diff
@@ -27,7 +29,22 @@
...
+    private $pipeline;
+
+    /** @var RouteMiddleware|ObjectProphecy */
+    private $routes;
+
+    /** @var RequestHandlerRunner|ObjectProphecy */
+    private $runner;
+
+    /** @var Application */
+    private $app;
```
xtreamwayz: Are private properties needed in Test classes?

test/Container/ApplicationConfigInjectionDelegatorTest.php
```diff
@@ -83,7 +84,7 @@ public function getQueueFromApplicationPipeline(Application $app)
         return $r->getValue($pipeline);
     }
 
-    public static function assertRoute($spec, array $routes)
+    public static function assertRoute(array $spec, array $routes)
     {
         Assert::assertThat(
             array_reduce($routes, function ($found, $route) use ($spec) {
```
xtreamwayz: I would take this to the next level and import functions:

`use function array_reduce;`

test/MiddlewareFactoryTest.php
```diff
@@ -24,33 +26,39 @@
...
+    private $factory;
+
+    protected function setUp()
     {
         $this->container = $this->prophesize(MiddlewareContainer::class);
         $this->factory = new MiddlewareFactory($this->container->reveal());
     }
 
-    public function assertLazyLoadingMiddleware(string $expectedMiddlewareName, MiddlewareInterface $middleware)
+    private function assertLazyLoadingMiddleware(string $expectedMiddlewareName, MiddlewareInterface $middleware)
```
xtreamwayz: Not sure about this. Although these assert functions are only used in this class, I would keep them in line with the phpunit style and make them at least static as well.

test/Router/IntegrationTest.php
```diff
@@ -458,16 +429,17 @@ public function testAllowedMethodsWhenNoHttpMethodsSet($adapter, $method)
...
 
-            return true;
-        }), Argument::any())->will(function (array $args) {
-            return $args[1]->handle($args[0]);
-        });
+                       return true;
+                   }), Argument::any())
+                   ->will(function (array $args) {
+                       return $args[1]->handle($args[0]);
+                   });
```
xtreamwayz: The spacing is wrong here. The original is good. there is no need to use so many spaces.

test/ApplicationTest.php
webimpress: It doesn't matter, it was not done automatically, I've added them manually where I've noted there are missing. These could be `protected` or even `public`, I just think that it's nice to have only `public test*` method and all others `private`. We don't need to have even these properties and we can define them dinalically but you know - for IDE it's better to have them in place :)

test/Container/ApplicationConfigInjectionDelegatorTest.php
webimpress: We could do, it's not a problem, sniffs for import internal functions and constants are there, but by default these are disabled, because per discussion in other repository it looks like we don't want to have it everywhere. If now we want it's not a problem to enable these sniffs :)

test/MiddlewareFactoryTest.php
webimpress: In PHPUnit style these should be `static`, and it can't be because this requires access to non static property of the class... So definitely the function shouldn't be used outside the class. When we change it to be static then we can make it public again. It's not important, as these are tests only...

test/Router/IntegrationTest.php
webimpress: The original is not clear, because we have `->will` starting in the middle of line.
When we have multiline chain call each part should be in new line, and now we have two options:
- align with `->` operator
- always just 4 spaces.

The sniff I wrote has this configuration and we can change it. Currently we are very inconsistent, because sometimes we have:
1. first call in the same line and every next in new line with one indent (4 spaces):
```php
$variable->call(...)
    ->method(...);
```

2. every call in new line with one indent (4 spaces):
```php
$variable
    ->call(....)
    ->method(....);
```

3. first call in the same line and next indented with `->` operator:
```php
$variable->call(...)
         ->method(...);
```

How the sniff is configured currently is to allow just 2 and 3 (case 1 is converted to 3).
We can reconfigure it to use 1 and 2 or make changes to have always 2.

Why I used 2 and 3? Because usually for short call 3 looks good and is readable, when we have something more complex then I would prefer to use 2.
Why I don't like 1? Because it's not aligned 🙄

src/Container/ResponseFactory.php
webimpress: And here is another big concern (at least for me). I was thinking about it a lot and I decided to write the sniff that way. I know that PHPStorm is formatting it differently but I think PHPStorm is formatting a lot of things differently/incorrectly (maybe this is configurable), and this is why I didn't take PHPStorm formatting as the correct one, and always have a though how it should be.

So the whole sniff which checks indents is really complicated. I had a lot issues with PHP_CodeSniffer one, so I've decided to write another one which will support everything.

Why I'm adding one more indent here?
For me it's a continuation of the expression - it is a continuation - concatenation in that case.

Usually when we do something like that we also have one more indent:
```php
$var = 'something'
    . ' continue here';
```

or:

```php
return 'hello'
    . ' world';
```

or:

```php
echo $veryLongVariable
    . $anotherVariable;
```

So, basically I think it looks clearer and you easly can see where the parameter is finished and where the next is starting. It's consistent with PSR-2 (I can't see any issues reported by PSR-2 checks), for example:
```php
sprintf(
    'This is'
        . ' very long messsage'
        . ' and it has one more indent on concatenation',
    'param1',
    'param2'
);
```

Let's now think about other operators, like `+`, how it should look like?
```php
method(
    $myVariable // please not this is just example, here could be even more complicated operation
        + ($anotherVariable * $multiplier),
    $secondParam
);
```

What about logical operators, let say:
```php
$this->callSomething(
    $this->checkIfYouCanGoToThePartyTonight($param1, $param2, $param3)
        && $this->havanaClubIsOpenTillMidnight($currentDate),
    $heyItsMe
);
```
Isn't it look clearer than:
```php
$this->callSomething(
    $this->checkIfYouCanGoToThePartyTonight($param1, $param2, $param3)
    && $this->havanaClubIsOpenTillMidnight($currentDate),
    $heyItsMe
);
```

Just another note - I always require to have operator at the beginning of the line, never at the end, so we can't have situation like that - what would be even more confusing:
```php
$this->callSomething(
    $this->checkIfYouCanGoToThePartyTonight($param1, $param2, $param3) && 
    $this->havanaClubIsOpenTillMidnight($currentDate),
    $heyItsMe
);
```

weierophinney: Agreed that the operator should not be on the end of the line, but at the start of the next.

I'm torn about these examples:

```php
$this->callSomething(
    $this->checkIfYouCanGoToThePartyTonight($param1, $param2, $param3)
        && $this->havanaClubIsOpenTillMidnight($currentDate),
    $heyItsMe
);
```

and

```php
$this->callSomething(
    $this->checkIfYouCanGoToThePartyTonight($param1, $param2, $param3)
    && $this->havanaClubIsOpenTillMidnight($currentDate),
    $heyItsMe
);
```

The latter is how I've been handling them. You can see a ton of examples particularly with `sprintf()` line continuations:

```php
throw new SomeDomainException(sprintf(
    'This is a really, really, really long line for which I want'
    . ' a continuation to occur. %s and all that',
    $string
));
```

My take has always been that seeing the operator on the subsequent line helps me know that this is a continuation, and thus no additional indentation is necessary. In fact, I tend to think of this as a way to create a visual block, as extra indentation generally implies a _new_ block (that may be due to having played a bit with python in the past, though).

I definitely prefer same level of indentation, but I can also see an argument for the extra indentation, particularly when it comes to argument lists, as it makes it more visually clear when a single argument ends.

I think this is one where we need to poll folks to find out if there's a clear preference.

test/ApplicationTest.php
weierophinney: This is one area where I'm definitely on the fence. Test classes do not have the same context as the source code for the package; I will never, ever use phpDoc on them, and appreciate the ability to create new properties on an ad-hoc basis.

That said, I totally understand folks who use IDEs and want to have the typehints, so they know what they're allowed to call, and with what arguments.

Again, I think we need a poll on this to see if there's consensus with a broader audience.

test/Container/ApplicationConfigInjectionDelegatorTest.php
weierophinney: I think if we enable it, we need to document how to disable the feature per-identifier and/or per file. (I already know how to do this, but we need to ensure there's good documentation for even new contributors.)

I'd be okay with enabling this by default; it definitely forces you to consider things like file-system operations and `require`/`include` within your code!

test/MiddlewareFactoryTest.php
weierophinney: Again, I'm on the fence here. Test classes are different than package classes. Nobody is going to re-use them, but if they do (and I've extended test cases a number of times), having `private` as the default is a real pain. Testing needs to be as easy as possible. Public visibility is perfectly fine in test classes.

test/Router/IntegrationTest.php
weierophinney: Spacing of the sample above looks fine to me; what did you object to, exactly, @xtreamwayz?

xtreamwayz: The file is gone! No more evidence :)

I think the original had 4 space indentation and the new had 12 or even more spaces.

Issue #504, opened 2017-08-30
Replace Application's __call magic method


jansvoboda11:
The argument reflowing in `__call` magic method on `Zend\Expressive\Application` is very confusing an unnecessarily complex.
I think that adding real `get`, `post`, `put`, `patch` and `delete` methods would be simpler.
The public API of the class wouldn't change, so this is not a breaking change.
Currently, the `Application` class provides signature of the routing methods via PHPDoc, but IDE's like PhpStorm get confused when extending and overriding those methods.

samsonasik:
should be close-able in favor of #505

PR #505, opened 2017-08-30
Replace magic __call method with real routing methods

jansvoboda11:
Solves #504 and simplifies tests.


weierophinney:
Thanks, @jansvoboda11; merged for an upcoming 2.1.0 release.

PR #507, opened 2017-09-06
ImplicitOptionsMiddleware: Add Access-Control-Allow-Methods if needed

moderndeveloperllc:
Add Access-Control-Allow-Methods to response if the request sends Access-Control-Request-Method. Used for cross-domain calls.

### DIFF
--- a/test/Middleware/ImplicitOptionsMiddlewareTest.php
+++ b/test/Middleware/ImplicitOptionsMiddlewareTest.php
@@ -169,7 +169,10 @@ class ImplicitOptionsMiddlewareTest extends TestCase
         $delegate->process($request->reveal())->shouldNotBeCalled();
 
         $expected = $this->prophesize(ResponseInterface::class);
-        $expected->withHeader('Access-Control-Allow-Methods', implode(',', $allowedMethods))->will([$expected, 'reveal']);
+        $expected->withHeader(
+            'Access-Control-Allow-Methods',
+            implode(',', $allowedMethods)
+        )->will([$expected, 'reveal']);
         $expected->withHeader('Allow', implode(',', $allowedMethods))->will([$expected, 'reveal']);
 
         $middleware = new ImplicitOptionsMiddleware($expected->reveal());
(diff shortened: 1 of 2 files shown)


test/Middleware/ImplicitOptionsMiddlewareTest.php
```diff
@@ -106,6 +106,7 @@ public function testWhenNoResponseProvidedToConstructorImplicitOptionsRequestCre
 
         $request = $this->prophesize(ServerRequestInterface::class);
         $request->getMethod()->willReturn(RequestMethod::METHOD_OPTIONS);
+        $request->hasHeader("Access-Control-Request-Method")->willReturn(false);
```
webimpress: Please use single quotes.

test/Middleware/ImplicitOptionsMiddlewareTest.php
```diff
@@ -134,6 +135,7 @@ public function testInjectsAllowHeaderInResponseProvidedToConstructorDuringOptio
...
```
webimpress: Single quotes here as well.

test/Middleware/ImplicitOptionsMiddlewareTest.php
```diff
@@ -146,4 +148,35 @@ public function testInjectsAllowHeaderInResponseProvidedToConstructorDuringOptio
...
+        $route = $this->prophesize(Route::class);
+        $route->implicitOptions()->willReturn(true);
+        $route->getAllowedMethods()->willReturn($allowedMethods);
+
+        $result = $this->prophesize(RouteResult::class);
+        $result->getMatchedRoute()->will([$route, 'reveal']);
+
+        $request = $this->prophesize(ServerRequestInterface::class);
+        $request->getMethod()->willReturn(RequestMethod::METHOD_OPTIONS);
+        $request->hasHeader("Access-Control-Request-Method")->willReturn(true);
```
webimpress: Single quotes, please.

weierophinney:
I agree with @webimpress here — this is beyond the scope of the existing ImplicitOptionsMiddleware and should be implemented as a separate middleware, covering the full spectrum of Access-Allow headers.

Please open a new pull request targeting the develop branch for that feature.

moderndeveloperllc:
@weierophinney @webimpress The reason I only did `Access-Control-Request-Method`/`Access-Control-Allow-Methods` is that the only other request header is `Access-Control-Request-Headers`. That's really too variable to do anything implicitly as that is an explicit control over what can be sent. To implicitly allow any header that the client asks for could be seen as a security risk.

The other `Access-Control-Allow-*` response headers are also all too security related to treat implicitly - their values should be set at the server level, not variably per request. `Access-Control-Request-Method`/`Access-Control-Allow-Methods` is unique amongst the headers as it can be securely set based solely on the application routing configuration.

Issue #360, opened 2016-06-23
How to change layout of error pages with Zend View


pdrosos:
Hi,

how can I change the layout of the Expressive error pages? For example 401 (Unauthorized), 404 (Not found), 405 (Method not allowed) etc. I am using Zend View.

Currently the error pages use the default layout, which is set in the templates configuration. But as I use the default layout on many other places too, I would like to set a different lightweight layout especially for the error pages without being forces to change the default layout and specifically set the layout everywhere else on rendering.

sniphpet:
Hello @pdrosos ,

I apologize in advance for the long answer. 

When you request an instance of ```TemplateRendererInterface``` the ```ZendViewRendererFactory``` is used. This factory return an instance of ```ZendViewRenderer```. Unfortunately  ```ZendViewRenderer``` allow you to set the template only during the instantiation.

Expressive error pages are handled by ``` Zend\Expressive\FinalHandler```. By default the FinalHandler resolve to ```TemplatedErrorHandlerFactory```.
If you are using Whoops the FinalHandler resolve to ```WhoopsErrorHandlerFactory```

These factories request an instance of  ```TemplateRendererInterface``` this is how expressive is able to use the layout when displaying error I guess.

To solve your issue I found 2 solutions
# Using reflection :

Inside ```templates.global.php ``` under the map array i added the following 
`'layout/error'   => 'path/to/my/error/template',`
Then I created my own ```TemplatedErrorHandlerFactory``` 

```php
namespace App\Factory;

use Interop\Container\ContainerInterface;
use Zend\Expressive\Template\TemplateRendererInterface;
use Zend\Expressive\TemplatedErrorHandler;
use Zend\ServiceManager\Factory\FactoryInterface;
use Zend\View\Resolver;
use Zend\View\Model\ViewModel;

class TemplatedErrorHandlerFactory implements FactoryInterface
{
    public function __invoke(ContainerInterface $container, $requestedName, array $options = null)
    {
        $template = $container->has(TemplateRendererInterface::class)
            ? $container->get(TemplateRendererInterface::class)
            : null;

        $r = new \ReflectionClass($template);
        $view = new ViewModel();
        $view->setTemplate('layout/error');
        $prop = $r->getProperty('layout');
        $prop->setAccessible(true);
        $prop->setValue($template, $view);

        return new TemplatedErrorHandler($template);
    }
}
```
Don't forget to map the final handler to this factory Inside ```templates.global.php``` : 
 ```php
[...]

PR #510, opened 2017-09-15
Added custom layout support in NotFoundDelegate

mano87:
This is a better implementation for chaning the layout of error pages. See also the issue: https://github.com/zendframework/zend-expressive/issues/360

I hope my pull request will be accepted :-)


throny:
Looks pretty good to me, probably a nice way to implement different templates for errorpages.

Note: I did NOT used it yet.

mano87:
Sorry, i now pulled in develop

webimpress:
@mano87 For future reference, you can change target branch, you don't need to close PR and open another one 😄

PR #511, opened 2017-09-15
Added custom layout support in NotFoundDelegate

mano87:
This is a better implementation for chaning the layout of error pages. See also the issue: #360 

I hope my pull request will be accepted :-)

### DIFF
--- a/src/Delegate/NotFoundDelegate.php
+++ b/src/Delegate/NotFoundDelegate.php
@@ -16,6 +16,7 @@ use Zend\Expressive\Template\TemplateRendererInterface;
 class NotFoundDelegate implements DelegateInterface
 {
     const TEMPLATE_DEFAULT = 'error::404';
+    const LAYOUT_DEFAULT = 'layout::error';
 
     /**
      * @var TemplateRendererInterface
@@ -36,18 +37,26 @@ class NotFoundDelegate implements DelegateInterface
     private $template;
 
     /**
+     * @var string
+     */
+    private $layout;
+
+    /**
      * @param ResponseInterface $responsePrototype
      * @param TemplateRendererInterface $renderer
      * @param string $template
+     * @param string $layout
      */
     public function __construct(
         ResponseInterface $responsePrototype,
         TemplateRendererInterface $renderer = null,
-        $template = self::TEMPLATE_DEFAULT
+        $template = self::TEMPLATE_DEFAULT,
+        $layout = self::LAYOUT_DEFAULT
     ) {
         $this->responsePrototype = $responsePrototype;
         $this->renderer          = $renderer;
         $this->template          = $template;
+        $this->layout            = $layout;
     }
 
     /**
@@ -93,9 +102,14 @@ class NotFoundDelegate implements DelegateInterface
      */
     private function generateTemplatedResponse(ServerRequestInterface $request)
     {
+        $templateData = [
+            'request'  => $request,
+            'layout'  => $this->layout,
+        ];
+
         $response = $this->responsePrototype->withStatus(StatusCodeInterface::STATUS_NOT_FOUND);
         $response->getBody()->write(
-            $this->renderer->render($this->template, ['request' => $request])
+            $this->renderer->render($this->template, $templateData)
         );
 
         return $response;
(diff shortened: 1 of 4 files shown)


src/Delegate/NotFoundDelegate.php
```diff
@@ -93,9 +102,14 @@ private function generatePlainTextResponse(ServerRequestInterface $request)
...
     {
+        $templateData = [
+            'request'  => $request,
+            'layout'  => $this->layout,
+        ];
+
         $response = $this->responsePrototype->withStatus(StatusCodeInterface::STATUS_NOT_FOUND);
         $response->getBody()->write(
-            $this->renderer->render($this->template, ['request' => $request])
+            $this->renderer->render($this->template, $templateData)
```
xtreamwayz: Why add a new variable?

```php
$this->renderer->render($this->template, [
    'request'  => $request,
    'layout'  => $this->layout,
]);
```

mano87: I will rework this point to make the parameter optional - or do you have a better idea? Another option would be that in the factory other optional parameters can be passed to the renderer no matter what...

src/Delegate/NotFoundDelegate.php
```diff
@@ -58,7 +67,7 @@ public function __construct(
      */
     public function process(ServerRequestInterface $request)
     {
-        if (! $this->renderer) {
+        if (!$this->renderer) {
```
webimpress: Revert this change please. CS is now failing:
```
There must be a single space after a NOT operator; 0 found
```

src/Delegate/NotFoundDelegate.php
```diff
@@ -78,7 +87,7 @@ private function generatePlainTextResponse(ServerRequestInterface $request)
             ->write(sprintf(
                 'Cannot %s %s',
                 $request->getMethod(),
-                (string) $request->getUri()
+                (string)$request->getUri()
```
webimpress: Revert this change please. We don;t have currently sniff to check it, so all seems to be fine, but for consistency we should have here single space.

test/Delegate/NotFoundDelegateTest.php
```diff
@@ -68,8 +70,10 @@ public function testUsesRendererToGenerateResponseContentsWhenPresent()
         $request = $this->prophesize(ServerRequestInterface::class)->reveal();
 
         $renderer = $this->prophesize(TemplateRendererInterface::class);
-        $renderer->render(NotFoundDelegate::TEMPLATE_DEFAULT, ['request' => $request])
-            ->willReturn('CONTENT');
+        $renderer->render(NotFoundDelegate::TEMPLATE_DEFAULT, [
+            'request' => $request,
+            'layout' => NotFoundDelegate::LAYOUT_DEFAULT
+            ])->willReturn('CONTENT');
```
webimpress: The last line should have one less indent:
```php
        $renderer->render(NotFoundDelegate::TEMPLATE_DEFAULT, [
            'request' => $request,
            'layout' => NotFoundDelegate::LAYOUT_DEFAULT
        ])->willReturn('CONTENT');
```

test/Delegate/NotFoundDelegateTest.php
```diff
@@ -68,8 +70,10 @@ public function testUsesRendererToGenerateResponseContentsWhenPresent()
         $request = $this->prophesize(ServerRequestInterface::class)->reveal();
 
         $renderer = $this->prophesize(TemplateRendererInterface::class);
-        $renderer->render(NotFoundDelegate::TEMPLATE_DEFAULT, ['request' => $request])
-            ->willReturn('CONTENT');
+        $renderer->render(NotFoundDelegate::TEMPLATE_DEFAULT, [
+            'request' => $request,
+            'layout' => NotFoundDelegate::LAYOUT_DEFAULT
```
webimpress: Please add trailing comma after last element in the array

src/Delegate/NotFoundDelegate.php
```diff
@@ -16,6 +16,7 @@
 class NotFoundDelegate implements DelegateInterface
 {
     const TEMPLATE_DEFAULT = 'error::404';
+    const LAYOUT_DEFAULT = 'layout::error';
```
weierophinney: The main problem I have with this is that we do not at this time define a `layout::error` template in the skeleton application, only a `layout::default`. As such, I think that should likely be the default in this situation, to allow for backwards compatibility with existing applications.

weierophinney:
Thanks, @mano87

PR #517, opened 2017-10-06
Added support for http-interop/http-middleware 0.5.0 (no BC Break)

webimpress:
We accomplish it by using `webimpress/http-middleware-compatibility` library
which allows us to use any version of http-interop middleware.
Consumer of the library can decide what version want to use.

In composer.lock we keep `http-interop/http-middleware` in version 0.4.1
so tests run with the following version of middleware:
- lowest: 0.1.1
- locked: 0.4.1
- latest: 0.5.0

The question is if we want to support http-interop/http-middleware in version 0.1.1. This requires method `next` in `NotFoundHandler` (as a proxy to handle method). Method `process` is also now a proxy to `handle` method.

## Requires:

- https://github.com/zendframework/zend-expressive-router/pull/36
- https://github.com/zendframework/zend-stratigility/pull/112

to be merged and released before.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,15 +24,26 @@
             "dev-develop": "2.1-dev"
         }
     },
+    "repositories": [
+        {
+            "type": "git",
+            "url": "https://github.com/webimpress/zend-expressive-router.git"
+        },
+        {
+            "type": "git",
+            "url": "https://github.com/webimpress/zend-stratigility.git"
+        }
+    ],
     "require": {
         "php": "^5.6 || ^7.0",
         "fig/http-message-util": "^1.1.2",
         "psr/container": "^1.0",
         "psr/http-message": "^1.0.1",
+        "webimpress/http-middleware-compatibility": "^0.1.1",
         "zendframework/zend-diactoros": "^1.3.10",
-        "zendframework/zend-expressive-router": "^2.1",
+        "zendframework/zend-expressive-router": "dev-feature/http-middleware-0.5 as 2.2",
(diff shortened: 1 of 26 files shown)


weierophinney:
Thanks, @webimpress!

weierophinney:
After seeing #519 opened, I started thinking about this, and realized we have a bit of a problem: with zend-stratigility 2.1.0, that package no longer directly depends on http-middleware, which means a fresh install or an update can leave an application without the http-middleware interfaces. Further, it means that a `composer update` can potentially upgrade to http-middleware 0.5.0... making all their existing middleware break.

I've raised this with @webimpress, and we feel that the way to fix this is to have a composer plugin package that checks for `http-interop/http-middleware` in the root `composer.json`; if it's not present, it would prompt the user to install it, suggesting either 0.4.1 or 0.5.0 as constraints (and defaulting to 0.4.1 in non-interactive installs). The compat package would then depend on this plugin.

weierophinney:
I'm going to unmerge this for the 2.1.0 release. In the 2.0 series, we already pin to http-middleware 0.4.1, which prevents triggering the prompt from Stratigility. While users will get the message that http-middleware is abandoned, this is true even with the v0.5.0 release that this particular patch provides support for.

I think at this point, it would be less invasive on the code to remove this patch, and do an immediate 2.1.0 release with the other features slated for it; we can then apply #529 against the release-3.0.0 branch to provide the final upgrade path to PSR-15.

PR #520, opened 2017-10-09
Fix typo

jezemery:
Fix typo from `rotue` to `route`


webimpress:
@jezemery Duplicate of #515

jezemery:
Ah. Thanks @webimpress

PR #523, opened 2017-10-11
Updated webimpress/http-middleware-compatibility to 0.1.4

webimpress:
This version install also composer plugin which prompt user to install `http-interop/http-middleware` as explicit dependency in their `composer.json`.
It prompts only when `http-interop/http-middleware` is not installed, otherwise it use version currently installed to update `composer.json` and message with information how to update is shown in console.

See also https://github.com/zendframework/zend-expressive/pull/517#issuecomment-335302764

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ cache:
 
 env:
   global:
-    - COMPOSER_ARGS="--no-interaction"
+    - COMPOSER_ARGS="--no-interaction --no-plugins"
     - COVERAGE_DEPS="satooshi/php-coveralls"
     - LEGACY_DEPS="phpunit/phpunit"
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ cache:
 
 env:
   global:
-    - COMPOSER_ARGS="--no-interaction --no-plugins"
+    - COMPOSER_ARGS="--no-interaction"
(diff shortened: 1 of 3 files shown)


vlakarados:
Can this be merged please? This is required for working with the latest stratigility and not working with deprecated http-interop/http-middleware

webimpress:
Closed in favour of #529

vlakarados:
@webimpress sorry, but why would we do this together with 7.1 requirement?

webimpress:
@vlakarados  PSR-15 is going to support only PHP 7.0+.  Currently http-interop/http-server-middleware also supports only PHP 7.0+.

This is decision we have made in ZF, for more information please see:
https://discourse.zendframework.com/t/psr-15-compatibility-issues-and-proposal/378/10

vlakarados:
@webimpress yes, but why it's 7.1 then? :)

webimpress:
@vlakarados See post on the forum.
It's said there:
> Update the minimum PHP version to 7.1 ([per our next major release milestone announcement](https://framework.zend.com/blog/2017-06-06-zf-php-7-1.html))

It has been announced in June 2017.

This PR is not gonna go live. Instead we are going to have new MAJOR release with PSR-15 (~ early 2018) and alpha 3.0.0 probably this week.

vlakarados:
@webimpress I see, too bad, but another reason and more motivation to upgrade to 7.1 :) Thanks!

PR #525, opened 2017-11-01
Test coverage for nested pipelines

danizord:
@weierophinney here's the test coverage for nested pipelines. It works in almost all cases, except when piping an `[$object, 'string']`. But I've opened for #527 that.

### DIFF
--- a/test/Application/MarshalMiddlewareTraitTest.php
+++ b/test/Application/MarshalMiddlewareTraitTest.php
@@ -13,6 +13,7 @@ use PHPUnit\Framework\TestCase;
 use Prophecy\Prophecy\ObjectProphecy;
 use Psr\Container\ContainerInterface;
 use Psr\Http\Message\ResponseInterface;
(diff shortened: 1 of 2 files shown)


danizord:
Here's the failing test:

```php
public function testPreparingArrayWithPairOfObjectAndStringMiddlewaresShouldNotBeTreatedAsCallable() {
    $first  = $this->prophesize(ServerMiddlewareInterface::class)->reveal();
    $second = TestAsset\CallableInteropMiddleware::class;

    $base = [$first, $second];

    $middleware = $this->prepareMiddleware($base);
    $this->assertInstanceOf(MiddlewarePipe::class, $middleware);

    $r = new ReflectionProperty($middleware, 'pipeline');
    $r->setAccessible(true);

    $this->assertCount(2, $r->getValue($middleware));
}
```

I haven't included it as part of this PR because I'm not sure if it is possible to solve. Any idea?

weierophinney:
Thanks, @danizord. I will work on #527 shortly.

PR #526, opened 2017-11-06
QA Tools + Docs update

webimpress:
Documentation moved into `docs` directory and all supporting files moved there.
Updated Travis CI configuration (removed allow failures on PHP 7.2 and updated legacy dependencies).
Updated composer skeleton (all support links, the same order of sections... consistency fixes).

__It is based on `develop` branch, as it is different from `master` and probably will go live soon.__


weierophinney:
Thanks, @webimpress

PR #529, opened 2017-11-20
PSR-15 support (http-interop/http-server-middleware) + PHP 7.1

webimpress:
- dropped PHP 5.6 and 7.0
- updated Travis CI configuration

Requires:
- [x] https://github.com/zendframework/zend-expressive-router/pull/39
- [x] https://github.com/zendframework/zend-stratigility/pull/122

We need update the docs, and remove probably some functionality like double-pass middlewares.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,28 +27,37 @@
     "config": {
         "sort-packages": true
     },
+    "repositories": [
+        {
+            "type": "git",
+            "url": "https://github.com/webimpress/zend-expressive-router.git"
+        },
+        {
+            "type": "git",
+            "url": "https://github.com/webimpress/zend-stratigility.git"
+        }
+    ],
     "require": {
-        "php": "^5.6 || ^7.0",
+        "php": "^7.1",
         "fig/http-message-util": "^1.1.2",
-        "http-interop/http-middleware": "^0.4.1",
+        "http-interop/http-server-middleware": "^1.0.1",
         "psr/container": "^1.0",
         "psr/http-message": "^1.0.1",
-        "webimpress/http-middleware-compatibility": "^0.1.1",
         "zendframework/zend-diactoros": "^1.3.10",
-        "zendframework/zend-expressive-router": "^2.2",
+        "zendframework/zend-expressive-router": "dev-feature/psr-15 as 2.2",
         "zendframework/zend-expressive-template": "^1.0.4",
-        "zendframework/zend-stratigility": "^2.1"
+        "zendframework/zend-stratigility": "dev-feature/psr-15 as 2.2"
     },
     "require-dev": {
-        "filp/whoops": "^2.1.6 || ^1.1.10",
-        "malukenho/docheader": "^0.1.5",
+        "filp/whoops": "^1.1.10 || ^2.1.13",
+        "malukenho/docheader": "^0.1.6",
         "mockery/mockery": "^1.0",
-        "phpunit/phpunit": "^5.7.22 || ^6.4.1",
+        "phpunit/phpunit": "^6.4.4",
         "zendframework/zend-coding-standard": "~1.0.0",
         "zendframework/zend-expressive-aurarouter": "^2.0",
         "zendframework/zend-expressive-fastroute": "^2.0",
         "zendframework/zend-expressive-zendrouter": "^2.0.1",
-        "zendframework/zend-servicemanager": "^3.3 || ^2.7.8"
+        "zendframework/zend-servicemanager": "^2.7.8 || ^3.3"
     },
     "conflict": {
         "container-interop/container-interop": "<1.2.0"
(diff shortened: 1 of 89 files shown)


asgrim:
Probably worth linking to the discussion on this for anyone researching: https://discourse.zendframework.com/t/psr-15-compatibility-issues-and-proposal/378

webimpress:
The last build is failing because of CS issue, it's fixed in #530

composer.json
```diff
@@ -27,28 +27,33 @@
...
+        "php": "^7.1",
         "fig/http-message-util": "^1.1.2",
-        "http-interop/http-middleware": "^0.4.1",
+        "http-interop/http-server-middleware": "^1.0.1",
         "psr/container": "^1.0",
         "psr/http-message": "^1.0.1",
-        "webimpress/http-middleware-compatibility": "^0.1.1",
         "zendframework/zend-diactoros": "^1.3.10",
-        "zendframework/zend-expressive-router": "^2.2",
+        "zendframework/zend-expressive-router": "dev-feature/psr-15 as 2.2",
```
weierophinney: You can now pin this to 3.0.x-dev or dev-release-3.0.0.

composer.json
```diff
@@ -27,28 +27,33 @@
     "config": {
         "sort-packages": true
     },
+    "repositories": [
+        {
+            "type": "git",
+            "url": "https://github.com/webimpress/zend-expressive-router.git"
+        }
+    ],
```
weierophinney: You can remove this now.

composer.json
```diff
@@ -27,28 +27,33 @@
...
+        "http-interop/http-server-middleware": "^1.0.1",
         "psr/container": "^1.0",
         "psr/http-message": "^1.0.1",
-        "webimpress/http-middleware-compatibility": "^0.1.1",
         "zendframework/zend-diactoros": "^1.3.10",
-        "zendframework/zend-expressive-router": "^2.2",
+        "zendframework/zend-expressive-router": "dev-feature/psr-15 as 2.2",
         "zendframework/zend-expressive-template": "^1.0.4",
-        "zendframework/zend-stratigility": "^2.1"
+        "zendframework/zend-stratigility": "dev-release-3.0.0"
```
weierophinney: Maybe try one of the following constraints:

- `^3.0.0@dev`
- `^3.0.0@dev as 3.0.0alpha1`
- `^3.0.0-dev`
- `^3.0.0-dev as 3.0.0alpha1`

My understanding is that we should use semver constraints instead of the branch name if an alias exists; I just can't recall off the top of my head what that should look like. Experiment! :grin:

composer.json
webimpress: Removed.

composer.json
webimpress: Unfortunately for now we need to have `3.0.x-dev as 2.3` as expressive-*router libraries does not support v3 yet. I'm going to work on it soon.

composer.json
webimpress: Looks like `^3.0.0@dev` and `^3.0.0-dev` are the same, when I try with alias I got error:
> Problem 1
    - The requested package zendframework/zend-stratigility ^3.0.0@dev as 3.0.0alpha1 exists as zendframework/zend-stratigility[1.0.0, 1.0.1, 1.0.2, 1.1.0, 1.1.1, 1.1.2, 1.1.3, 1.2.0, 1.2.1, 1.3.0, 1.3.1, 1.3.2, 1.3.3, 2.0.0, 2.0.1, 2.1.0, 2.1.1, 2.1.2, dev-develop, 2.2.x-dev, dev-master, 2.1.x-dev, dev-release-3.0.0, 3.0.x-dev] but these are rejected by your constraint.

webimpress:
@weierophinney because we decided to typehint on `MiddlewareInterface` in `Route` we have to load all middlewares (also routed) on application start. Before routed middlewares were loaded only when the path was hit, and it was right, IMO. Do you have any suggestion what we can do or we should just revert change in zend-expressive-router?

Other that than we have some issues in test caused by bug in prophecy, please see:
https://github.com/phpspec/prophecy/issues/368
I've already provided fix there (https://github.com/phpspec/prophecy/pull/377) but not sure when we have it released.

weierophinney:
> @weierophinney because we decided to typehint on MiddlewareInterface in Route we have to load all middlewares (also routed) on application start. Before routed middlewares were loaded only when the path was hit, and it was right, IMO. Do you have any suggestion what we can do or we should just revert change in zend-expressive-router?

This is not true. In the v2 series, `route()` will create a `LazyLoadingMiddleware` instance when a middleware service name is provided, and inject that into the `Route` instance created.

Currently, the only way the middleware passed to a `Route` instance will NOT be a `MiddlewareInterface` is if the `Route` is created manually by the user.

weierophinney:
@webimpress Do not worry about rebasing; I've almost completed the merge locally, and will push to release-3.0.0 shortly.

weierophinney:
Thanks, @webimpress!

PR #536, opened 2017-12-12
Fix zend-expressive-template dev-dependency

xtreamwayz:
The zendframework/zend-expressive-template dependency was not yet updated.


weierophinney:
Thanks, @xtreamwayz

PR #538, opened 2017-12-14
Add missing types

snapshotpl:
As we support PHP 7.1 as minimum we can use new language features - strict types.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -114,7 +114,7 @@ class Application extends MiddlewarePipe
      * @param null|string $name The name of the route.
      * @return Router\Route
      */
-    public function get($path, $middleware, string $name = null): Router\Route
+    public function get($path, $middleware, string $name = null) : Router\Route
     {
         return $this->route($path, $middleware, ['GET'], $name);
     }
@@ -125,7 +125,7 @@ class Application extends MiddlewarePipe
      * @param null|string $name The name of the route.
      * @return Router\Route
      */
-    public function post($path, $middleware, $name = null): Router\Route
+    public function post($path, $middleware, $name = null) : Router\Route
     {
         return $this->route($path, $middleware, ['POST'], $name);
     }
@@ -136,7 +136,7 @@ class Application extends MiddlewarePipe
      * @param null|string $name The name of the route.
      * @return Router\Route
      */
-    public function put($path, $middleware, string $name = null): Router\Route
+    public function put($path, $middleware, string $name = null) : Router\Route
     {
         return $this->route($path, $middleware, ['PUT'], $name);
     }
@@ -147,7 +147,7 @@ class Application extends MiddlewarePipe
      * @param null|string $name The name of the route.
      * @return Router\Route
      */
-    public function patch($path, $middleware, string $name = null): Router\Route
+    public function patch($path, $middleware, string $name = null) : Router\Route
     {
         return $this->route($path, $middleware, ['PATCH'], $name);
     }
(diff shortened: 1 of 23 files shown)


snapshotpl:
OK. I think more repos from new dev expressive need same changes. PR in progress ;)

BTW: after my changes I have found that many phpdocs dupicates with types defined directly in code. What are you think to avoid it when they are 1:1 same?

weierophinney:
> OK. I think more repos from new dev expressive need same changes. PR in progress ;)

Just curious: which ones? PHP 7.1 typehint patches were some of the first submitted and merged, so I'm curious what we missed.

> BTW: after my changes I have found that many phpdocs dupicates with types defined directly in code. What are you think to avoid it when they are 1:1 same?

The plan is to remove any annotations that duplicate what is in the signature already. If they add information, however, we keep them. (E.g., if a typehint is `iterable` or `array`, we may detail the structure, or indicate something like `MoreSpecificType[]`; in those cases, we'd keep the annoation.)

Most of the patches we've merged against repos receiving updates do this already.

snapshotpl:
You're right - any other repository looks fine - I don't check it before I write it.

Should I delete duplicates?

weierophinney:
> Should I delete duplicates?

Yes; you'll see from the patches to the other repos that this is what we've done.

Thanks!

src/ApplicationConfigInjectionTrait.php
```diff
@@ -240,8 +237,6 @@ private function createCollectionMapper()
      *
      * The function is useful to reduce an array of pipeline middleware to a
      * priority queue.
-     *
-     * @return callable
```
webimpress: Please then add return type `callable` on the method.

src/Application.php
```diff
@@ -109,67 +106,55 @@ public function __construct(
     }
 
     /**
-     * @param string|Router\Route $path
      * @param callable|string $middleware Middleware (or middleware service name) to associate with route.
      * @param null|string $name The name of the route.
-     * @return Router\Route
      */
-    public function get($path, $middleware, $name = null)
+    public function get($path, $middleware, string $name = null) : Router\Route
```
webimpress: typehint `string` should be added on `$path`, see method `route`.

src/Application.php
```diff
@@ -109,67 +106,55 @@ public function __construct(
     }
 
     /**
-     * @param string|Router\Route $path
      * @param callable|string $middleware Middleware (or middleware service name) to associate with route.
```
webimpress: types should be updated, `callable|string|MiddlewareInterface`

src/Application.php
```diff
@@ -109,67 +106,55 @@ public function __construct(
     }
 
     /**
-     * @param string|Router\Route $path
```
webimpress: Please be careful with removing comments like that, you've removed the comment but haven't update the typehint in the function declaration. In that case it's fine because as I've mentioned the `$path` is possible to be string only, so the comment will be removed anyway.

src/Application.php
```diff
@@ -206,7 +191,6 @@ public function any($path, $middleware, $name = null)
      *
      * @param string|array|callable $path Either a URI path prefix, or middleware.
      * @param null|string|array|callable $middleware Middleware
-     * @return self
```
webimpress: It should not be removed, I think we should have here `$this` instead.
See PSR-5 draft (the very bottom of the page):
https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md
> `$this`, the element to which this type applies is the same exact instance as the current class in the given context. As such this type is a stricter version of static as, in addition, the returned instance must not only be of the same class but also the same instance.

src/Application.php
```diff
@@ -317,7 +297,7 @@ public function route(string $path, $middleware, array $methods = null, string $
      *
      * @return Router\Route[]
      */
-    public function getRoutes()
+    public function getRoutes() : iterable
```
webimpress: I think we can use just `array` here.

PR #539, opened 2017-12-21
Update setting-locale-without-routing-parameter.md

jbelien:
Documentation update : [How can I setup the locale without routing parameters?](https://docs.zendframework.com/zend-expressive/cookbook/setting-locale-without-routing-parameter/)

Fix the removing the locale segment so it works with both formats : `de` **and** `de_DE`


weierophinney:
Thanks, @jbelien

PR #542, opened 2018-01-20
Updated documentation - Aura.Di and Pimple containers configuration

webimpress:
Now we have separate repositories which provide configuration for these dependency injection containers and these should be used with expressive applications.


weierophinney:
Thanks, @webimpress!

PR #543, opened 2018-01-25
Refactor: extract middleware marshaling, routing, and app running

weierophinney:
This patch implements the changes as outlined in the [RFC: Expressive 3 Design Changes](https://discourse.zendframework.com/t/rfc-expressive-3-design-changes/438), as well as those as outlined in [this comment](https://github.com/zendframework/zend-expressive/pull/543#issuecomment-361386657) and [this comment](https://github.com/zendframework/zend-expressive/pull/543#issuecomment-361725207), below.

In particular:

- Updates to consume and implement PSR-15.

- The following changes were made to middleware marshaling:
  - A new class, `MiddlewareContainer`, implementing PSR-11's `ContainerInterface` and decorating a `ContainerInterface`. This class ensures that middleware pulled from the container implements the PSR-15 `MiddlewareInterface`. It also allows retrieving middleware by class name if the class does not require constructor arguments.

  - A new class, `MiddlewareFactory`, accepts a PSR-11 `ContainerInterface` to its constructor, which it decorates internally as a `MiddlewareContainer`. It then exposes methods for marshaling middleware:

    - `callable(callable $middleware) : CallableMiddlewareDecorator`
    - `lazy(string $middleware) : LazyLoadingMiddleware`
    - `pipeline(...$middleware) : MiddlewarePipeline`
    - `path(string $path, $middleware) : PathMiddlewareDecorator`
    - `prepare($middleware) : MiddlewareInterface`; this one proxies to the other methods.

    Double-pass middleware is not supported.

- A new class, `ApplicationRunner`, composes:
  - A `RequestHandlerInterface` instance (typically, a `MiddlewarePipeInterface`)
  - An `EmitterInterface` instance (typically, an `EmitterStack` composing a `SapiEmitter`)
  - A callable factory for generating a `ServerRequestInterface` instance
  - A callable response generator, used when the server request factory raises a throwable or exception.

  When `run()` is called, the class marshals a request, potentially generates an error response, and otherwise passes handling of the request to the handler; it then passes the returned response to the emitter.

  **This class can potentially be extracted to its own library.**

- The `RouteMiddleware` has been updated to add the methods `route`, `get`, `post`, etc. from `Application`. Unlike `Application`, they only accept `MiddlewareInterface` instances.

- New factories (in the `Zend\Expressive\Container` namespace):
  - `RouteMiddlewareFactory`
  - `DispatchMiddlewareFactory`
  - `EmitterFactory` should be mapped to `Zend\Diactros\Response\EmitterInterface`.
  - `ServerRequestFactoryFactory` should be mapped to `Zend\Expressive\ServerRequestFactory`
  - `ServerRequestErrorResponseGeneratorFactory` should be mapped to `Zend\Expressive\ServerRequestErrorResponseGenerator`.

- `Application` has been completely rewritten. It now composes:
  - A `MiddlewareFactory` instance
  - A `MiddlewarePipeInterface` instance, generally a `MiddlewarePipe`
  - A `RouteMiddleware` instance
  - An `ApplicationRunner` instance

  It implements `RequestHandlerInterface` and `MiddlewareInterface`, and composes the `ApplicationConfigInjectionTrait`. It defines the same public API as previously, minus the various getters (as none of them are applicable anymore). Each method now proxies to the appropriate collaborator, pre-processing arguments as necessary:

  - `handle()` delegates to the `MiddlewarePipeInterface`
  - `process()` delegates to the `MiddlewarePipeInterface`
  - `run()` delegates to the `ApplicationRunner`
  - `pipe()` delegates to the `MiddlewarePipeInterface`, after first passing its arguments to the `MiddlewareFactory`, and, in the case of two arguments, the `Zend\Stratigility\path()` utility function.
  - `route()` and related methods delegate to the `RouteMiddleware`, after first passing the `$middleware` argument to the `MiddlewareFactory`
  - `pipeRoutingMiddleware` and `pipeDispatchMiddleware` are removed; users will now pipe these middleware just like any other middleware.

- `Zend\Expressive\Delegate\NotFoundDelegate` was renamed to `Zend\Expressive\Handler\NotFoundHandler`.

- `Zend\Expressive\Middleware\NotFoundHandler` was renamed to `Zend\Expressive\Middleware\NotFoundMiddleware`; its related factory in the `Zend\Expressive\Container` namespace was similarly renamed.

- Tests were updated to reflect the changes, as necessary.

### Updates

- 2018-01-30: Updated to reflect refactors of routing and app running.


src/ApplicationConfigInjectionTrait.php
```diff
@@ -9,6 +9,7 @@
 
 namespace Zend\Expressive;
 
+use Psr\Http\Server\MiddlewareInterface;
```
xtreamwayz: MiddlewareInterface is not used.

src/Container/ApplicationFactory.php
```diff
@@ -64,15 +61,19 @@ public function __invoke(ContainerInterface $container) : Application
...
+            ? $container->get('Zend\Expressive\Handler\DefaultHandler')
             : null;
 
         $emitter = $container->has(EmitterInterface::class)
             ? $container->get(EmitterInterface::class)
             : null;
 
-        $app = new Application($router, $container, $delegate, $emitter);
+        $response = $container->has(ResponseInterface::class)
+            ? $container->get(ResponseInterface::class)
```
xtreamwayz: ResponseInterface needs import.

src/Middleware/DispatchMiddleware.php
```diff
@@ -9,11 +9,11 @@
 
 namespace Zend\Expressive\Middleware;
 
-use Interop\Http\Server\MiddlewareInterface;
-use Interop\Http\Server\RequestHandlerInterface;
 use Psr\Container\ContainerInterface;
```
xtreamwayz: ContainerInterface not used.

src/Middleware/DispatchMiddleware.php
```diff
@@ -9,11 +9,11 @@
...
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
 use Zend\Expressive\MarshalMiddlewareTrait;
```
xtreamwayz: MarshalMiddlewareTrait doesn't exist.

src/Middleware/DispatchMiddleware.php
```diff
@@ -9,11 +9,11 @@
...
 use Zend\Expressive\Router\RouteResult;
 use Zend\Expressive\Router\RouterInterface;
```
xtreamwayz: RouterInterface isn't used.

composer.json
```diff
@@ -24,13 +24,13 @@
     "require": {
         "php": "^7.1",
         "fig/http-message-util": "^1.1.2",
-        "http-interop/http-server-middleware": "^1.0.1",
         "psr/container": "^1.0",
         "psr/http-message": "^1.0.1",
+        "psr/http-server-middleware": "^1.0.0",
```
webimpress: Usually we use just `^1.0`.

src/Application.php
```diff
@@ -85,212 +89,165 @@ class Application extends MiddlewarePipe
...
+     *
+     * Once it has processed itself, it emits the returned response using the
+     * composed emitter.
      */
-    public function post(string $path, $middleware, $name = null) : Router\Route
+    public function run(ServerRequestInterface $request = null, ResponseInterface $response = null) : void
     {
-        return $this->route($path, $middleware, ['POST'], $name);
+        try {
+            $request  = $request ?: ServerRequestFactory::fromGlobals();
```
webimpress: One space before `=` please

src/Application.php
```diff
@@ -85,212 +89,165 @@ class Application extends MiddlewarePipe
...
+        );
+
+        $emitter = $this->getEmitter();
+        $emitter->emit($response);
     }
 
     /**
-     * @param callable|string $middleware Middleware (or middleware service name) to associate with route.
-     * @param null|string $name The name of the route.
+     * @param string|array|callable|MiddlewareInterface $middleware Middleware
```
webimpress: As I've suggested in another place order of types - `null|(simple types in alphabetical order)|(complex types in alphabetical order)` - so here we should have `array|callable|string|MiddlewareInterface`.

src/Container/DispatchMiddlewareFactory.php
```diff
@@ -0,0 +1,21 @@
+<?php
+/**
+ * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
+ * @copyright Copyright (c) 2018 Zend Technologies USA Inc. (https://www.zend.com)
+ * @license   https://github.com/zendframework/zend-expressive/blob/master/LICENSE.md New BSD License
+ */
+
```
webimpress: No new line before strict_type declaration please.

src/Container/DispatchMiddlewareFactory.php
```diff
@@ -0,0 +1,21 @@
...
+namespace Zend\Expressive\Container;
+
+use Psr\Container\ContainerInterface;
+use Zend\Expressive\Middleware\DispatchMiddleware;
+
+class DispatchMiddlewareFactory
+{
+    public function __invoke(ContainerInterface $container) : DispatchMiddleware
+    {
+        return new DispatchMiddleware();
```
webimpress: Do we need that factory? Isn't it `invokable`?

src/Container/NotFoundHandlerFactory.php
```diff
@@ -1,7 +1,7 @@
 <?php
 /**
  * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
- * @copyright Copyright (c) 2016-2017 Zend Technologies USA Inc. (https://www.zend.com)
+ * @copyright Copyright (c) 2017 Zend Technologies USA Inc. (https://www.zend.com)
```
webimpress: Why 2016 is removed? It could be updated to 2018.

src/Container/NotFoundMiddlewareFactory.php
```diff
@@ -0,0 +1,22 @@
+<?php
+/**
+ * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
+ * @copyright Copyright (c) 2016-2017 Zend Technologies USA Inc. (https://www.zend.com)
```
webimpress: New file, 2018 only please.

src/Container/NotFoundMiddlewareFactory.php
```diff
@@ -0,0 +1,22 @@
...
+ * @license   https://github.com/zendframework/zend-expressive/blob/master/LICENSE.md New BSD License
+ */
+
```
webimpress: Remove empty line before strict_type declaration.

src/Container/RouteMiddlewareFactory.php
```diff
@@ -0,0 +1,26 @@
+<?php
+/**
+ * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
+ * @copyright Copyright (c) 2018 Zend Technologies USA Inc. (https://www.zend.com)
+ * @license   https://github.com/zendframework/zend-expressive/blob/master/LICENSE.md New BSD License
+ */
+
```
webimpress: Remove empty line before strict_type declaration

src/Exception/InvalidMiddlewareException.php
```diff
@@ -9,8 +9,35 @@
 
 namespace Zend\Expressive\Exception;
 
+use Psr\Http\Server\MiddlewareInterface;
 use RuntimeException;
 
 class InvalidMiddlewareException extends RuntimeException implements ExceptionInterface
 {
+    /**
+     * @var mixed $middleware The middleware that does not fulfill the
```
webimpress: `@param` instead of `@var` please

src/Exception/InvalidMiddlewareException.php
```diff
@@ -9,8 +9,35 @@
...
+        return new self(sprintf(
+            'Middleware "%s" is neither a string service name, a PHP callable,'
+            . ' a %s instance, or an array of such arguments',
+            is_object($middleware) ? get_class($middleware) : gettype($middleware),
+            MiddlewareInterface::class
+        ));
+    }
+
+    /**
+     * @var mixed $service The actual service created by the container.
```
webimpress: `@param` instead of `@var` please

src/MiddlewareContainer.php
```diff
@@ -0,0 +1,64 @@
+<?php
+/**
+ * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
+ * @copyright Copyright (c) 2018 Zend Technologies USA Inc. (https://www.zend.com)
+ * @license   https://github.com/zendframework/zend-expressive/blob/master/LICENSE.md New BSD License
+ */
+
```
webimpress: Remove empty line before strict_type declaration.

src/MiddlewareFactory.php
```diff
@@ -0,0 +1,147 @@
+<?php
+/**
+ * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
+ * @copyright Copyright (c) 2018 Zend Technologies USA Inc. (https://www.zend.com)
+ * @license   https://github.com/zendframework/zend-expressive/blob/master/LICENSE.md New BSD License
+ */
+
```
webimpress: Remove empty line before strict_type declaration

src/MiddlewareFactory.php
```diff
@@ -0,0 +1,147 @@
...
+        if (is_callable($middleware)) {
+            return $this->callable($middleware);
+        }
+
+        if (is_array($middleware)) {
+            return $this->pipeline(...$middleware);
+        }
+
+        if ((! is_string($middleware) || empty($middleware))) {
```
webimpress: remove one pair of parentheses and the second part can be just `! $middleware`

src/MiddlewareFactory.php
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    /**
+     * Create lazy loading middleware based on a service name.
+     */
+    public function lazy(string $middleware) : Middleware\LazyLoadingMiddleware
+    {
+        if (! $this->container) {
+            throw new Exception\ContainerNotRegisteredException(sprintf(
+                'Cannot marshal middleware by service name "%s"; no container registered',
```
webimpress: Maybe it should be moved to the static method in the exception?

src/MiddlewareFactory.php
```diff
@@ -0,0 +1,147 @@
...
+class MiddlewareFactory
+{
+    /**
+     * @var null|MiddlewareContainer
+     */
+    private $container;
+
+    public function __construct(ContainerInterface $container = null)
+    {
+        $this->container = $container ? new MiddlewareContainer($container) : null;
```
webimpress: For better testability we should pass `MiddlewareContainer` as param instead of creating one in the constructor.

test/Container/NotFoundMiddlewareFactoryTest.php
```diff
@@ -0,0 +1,32 @@
+<?php
+/**
+ * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
+ * @copyright Copyright (c) 2016-2017 Zend Technologies USA Inc. (https://www.zend.com)
```
webimpress: 2018 only, please

test/Container/NotFoundMiddlewareFactoryTest.php
```diff
@@ -0,0 +1,32 @@
...
+ * @license   https://github.com/zendframework/zend-expressive/blob/master/LICENSE.md New BSD License
+ */
+
```
webimpress: Remove empty line before strict_type declaration

test/MiddlewareContainerTest.php
```diff
@@ -0,0 +1,88 @@
+<?php
+/**
+ * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
+ * @copyright Copyright (c) 2018 Zend Technologies USA Inc. (https://www.zend.com)
+ * @license   https://github.com/zendframework/zend-expressive/blob/master/LICENSE.md New BSD License
+ */
+
```
webimpress: Remove empty line before strict_type declaration

test/MiddlewareFactoryTest.php
```diff
@@ -0,0 +1,269 @@
+<?php
+/**
+ * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
+ * @copyright Copyright (c) 2018 Zend Technologies USA Inc. (https://www.zend.com)
+ * @license   https://github.com/zendframework/zend-expressive/blob/master/LICENSE.md New BSD License
+ */
+
```
webimpress: Remove empty line before strict_type declaration please

src/Container/DispatchMiddlewareFactory.php
Ocramius: @webimpress let's try reducing noise over CS by moving these discussions to the CS checks repo instead 👍

froschdesign: I'm also little bit unhappy with this type of comments. Because the first comments are always related to the coding standard. After the _noise_, there are often no other comments. CS is important, but code formatting should not have the first place in this and other discussions. (imo)

froschdesign: _Coding Standard Police_ 😉

webimpress: This is why we need proper CS for zf and then we are not going to have these comments at all. Sorry, I'm adding comments about everything I can see at the beginning because otherwise (it happened already couple times) it could be merged without fixing...

src/Container/NotFoundHandlerFactory.php
froschdesign: Why we do not have a script for the entire header? It is tiring to read, comment and fix always the same.

webimpress:
@weierophinney 
> A new class, MiddlewareContainer, implementing PSR-11's ContainerInterface and decorating a ContainerInterface. This class ensures that middleware pulled from the container implements the PSR-15 MiddlewareInterface. It also allows retrieving middleware by class name if the class does not require constructor arguments.

Do we need it? Maybe we should just register these type of middlewares as invokable in container?

src/Container/NotFoundHandlerFactory.php
webimpress: Because we still don't have zend-coding-standard in place?
And as you can see different developers usually use different style here, different years, even they know how it should be.... Even in this PR we have different range year for new files, year changed here, not updated to the current year...

src/Container/DispatchMiddlewareFactory.php
froschdesign: @webimpress 
Sorry, my comment addressed not only to you. It was a hint to: zendframework/zend-coding-standard/pull/1

src/Application.php
weierophinney: I would argue for the following order:

- `null`
- all other scalar types, in alphabetical order
- complex built-in types (`array`, `callable`, `resource`) in alphabetical order
- instance types, in alphabetical order

So, clearly we have differing opinions, and we'll need to make a decision. In the meantime, this is not the main point of this PR at this time; I need feedback on _architecture_ and _design_.

src/Container/DispatchMiddlewareFactory.php
weierophinney: It _is_ invokable, but it may not always be. Additionally, having it here simplifies how we refer to it when writing configuration for different container types.

src/Container/NotFoundHandlerFactory.php
weierophinney: I'm not sure why 2017 was used here; it should have been 2018; likely me still getting used to the new year.

This class is technically new, as it's now creating a `Zend\Expressive\Handler\NotFoundHandler` vs the old `Zend\Expressive\Middleware\NotFoundHandler`. The class below, `NotFoundMiddlewareFactory`, is essentially a rename of the latter factory.

src/Container/NotFoundMiddlewareFactory.php
weierophinney: This is a _renamed_ file; it used to be `NotFoundHandlerFactory`, but the name changed due to correctly identifying `Zend\Expressive\Middleware\NotFoundHandler` as _middleware_.

src/Container/NotFoundMiddlewareFactory.php
weierophinney: This is not relevant at this time _as all other files in this branch already have an empty line between the file-level docblock and the declaration_; as such, this is just internally consistent.

We need to have a separate discussion in the zend-coding-standard repo about what our standard will be, as we now have two repos using two different approaches.

src/MiddlewareFactory.php
weierophinney: It makes usability harder, however, as developers who are creating an instance on the fly now need to do the following:

```php
$factory = new MiddlewareFactory(new MiddlewareContainer($app->getContainer()));
```

Additionally, the above example I present would fail with an `E_FATAL` if `$app->getContainer()` returns `null` (which is possible currently), which means it technically needs to become:

```php
$factory = new MiddlewareFactory(
    $app->getContainer()
    ? new MiddlewareContainer($app->getContainer())
    : null
);
```

which is not user friendly at all. (FTR, if the middleware in a specification or a pipeline resolves to an constructor-less class, a callable, or a `MiddlewareInterface` instance, the factory can still do its work; it's only if a service name is used that the container becomes necessary. At that point, it raises an exception when a container is missing.)

test/Container/NotFoundMiddlewareFactoryTest.php
weierophinney: This file was renamed from `NotFoundHandler`; date range is correct.

src/MiddlewareFactory.php
```diff
@@ -0,0 +1,147 @@
...
+
+    /**
+     * Segregate one or more middleware by path.
+     *
+     * Creates and returns a PathMiddlewareDecorator instance after first
+     * passing the $middleware argument to prepare().
+     *
+     * @param string|array|MiddlewareInterface $middleware
+     */
+    public function path(string $path, $middleware) : PathMiddlewareDecorator
```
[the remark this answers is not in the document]
weierophinney: I'm considering removing the `path()` method, as the same can be accomplished with the following:

```php
use function Zend\Stratigility\path;

$app->pipe(path('/api', $factory->prepare($middleware));
```

If we keep the methods down to only those needed for preparing middleware of different types (`prepare()`, `callable()`, `lazy()`) and pipelines (`pipeline()`), we can use these with any of the various decorator classes and/or their related utility functions. That means we can add new decorators and utility functions in Stratigility or Expressive, and they're automatically supported.

src/Container/ApplicationFactory.php
weierophinney: Done.

weierophinney:
> > A new class, MiddlewareContainer, implementing PSR-11's ContainerInterface and decorating a ContainerInterface. This class ensures that middleware pulled from the container implements the PSR-15 MiddlewareInterface. It also allows retrieving middleware by class name if the class does not require constructor arguments.
> 
> Do we need it? Maybe we should just register these type of middlewares as invokable in container?

Ideally, that's what we'd do. However, that poses a couple of problems:

- For users relying on that functionality currently, their code would break after upgrading, until they add these `invokables` declarations.
- From a usability perspective, if you're declaring a one-off class that has no dependencies, requiring adding configuration for it is more work that is prone to error (forgetting to add the factory).
- From a context perspective: if we are pulling middleware services from the container, we want to be assured that we don't accidentally pull a non-middleware service due to misconfiguration. Having a check on the type fetched is useful in that regard. (And analogous to the scoped `ControllerManager` in zend-mvc.)

We can largely solve the second use case via the zend-expressive-tooling project; in fact, [there is an open issue requesting creation of a factory when middleware:create is called](https://github.com/zendframework/zend-expressive-tooling/issues/42). The first is harder, however; unless we can write a tool to solve that migration problem, it's simpler to have the dedicated container.

The third case, however, is probably the more winning argument here, as it ensures that we _only_ fetch middleware in the context of _marshaling_ middleware for the pipeline and/or routing.

weierophinney:
@Ocramius , @froschdesign , @webimpress, @ezimuel, @xtreamwayz — any feedback on design/implementation?

Thoughts I've had since implementing (some of which are due to discussions with @ezimuel and @xtreamwayz):

- Creating factories for both the `MiddlewareContainer` and `MiddlewareFactory` classes. This would allow us to re-use each (though the former would not likely get used outside the latter), and allow us to provide stronger type-hinting on the `MiddlewareFactory`, resolving the discussion with @webimpress, above.

- Pushing all routing-related methods, the router itself, and a `MiddlewareFactory` instance, into the `RouteMiddleware`. Users would register route-based middleware with that instance:

  ```php
  $routes->get('/', HomeMiddleware::class, 'home');
  $routes->post('/api/books', [
      CorsMiddleware::class,
      AuthenticationMiddleware::class,
      CreateBookMiddleware::class
  ]);
  ```

  My inclination is also to have the `$middleware` argument be strongly typed against `MiddlewareInterface` (more on that below).

  This has the benefit of simplifying the `Application` API, and moving the routing functionality closer to the middleware that uses it.

- Have `Application` implement `MiddlewarePipeInterface` directly, and only accept middleware instances. This means the `Application` instance does not need to compose a container any longer, much less the `MiddlewareFactory`. More on this later.

- Splitting the functionality of piping/processing/handling from _running_ an application.

  What do I mean by that?

  Right now, the remaining complexity of the `Application` class, once routing concerns are removed, is around the `run()` method, as that method requires the following:

  - A way to generate a `ServerRequestInterface` instance if none is provided.  This is currently via the static method `ServerRequestFactory::fromGlobals()`, but we should likely make this configurable in order to accommodate PSR-17 factories in the future.
  - A way to generate an error response in the case that the server request factory raises an exception. This is done currently in two ways, depending on collaborators present:
    - If the instance has a container, and the container has an `ErrorResponseGenerator`, that instance is used to generate the response, by passing an empty `ServerRequest` instance (created on the fly) and a composed response prototype.
    - Otherwise, the response is generated by manipulating the composed response prototype.
  - An emitter, for emitting the response.

  If we were to split responsibilities, the `Application` class, as it is currently named would compose only:

  - a default handler, for times when `handle()` is called.

  We could then create a new class, perhaps an `ApplicationRunner`, that could compose the following:

  - An emitter. The container would define an `EmitterInterface` service that resolves to an `EmitterStack` with the `SapiEmitter` composed by default.
  - A `ServerRequestErrorResponseGenerator` service; this would be used when unable to marshal the server request, in order to generate the response.  It should be capable of receiving only an exception. (A factory could close over the `ErrorResponseGenerator` in order to provide a default `ServerRequest` and response prototype, so as to allow re-use of existing functionality.)

- Create a proxy class that composes the `RouteMiddleware`, `Application`, and `ApplicationRunner`, along with a `MiddlewareFactory`. This class would _mimic the current v2 API_ with regards to piping and routing. Its primary use case would be for applications _upgrading from v2_.

Essentially, for users _upgrading_ from v2, we'd update their `Application` service to resolve to the proxy class. This would allow them to gradually upgrade to the v3 APIs. In v3, we'd alter the `public/index.php` self-executing function: we'd import the application, routing middleware, and middleware factory as variables before calling the `require` statements:

```php
$app = $container->get(Application::class);
$factory = $container->get(MiddlewareFactory::class);
$routes = $container->get(RouteMiddleware::class);

require 'config/pipeline.php';
require 'config/routes.php';
```

These files would then look like the following:

```php
// config/pipeline.php:
use function Zend\Stratigility\path;

$app->pipe(ErrorMiddleware::class);
$app->pipe(path('/api', $factory->pipeline(
    CorsMiddleware::class,
    AuthenticationMiddleware::class,
    AuthorizationMiddleware::class
));

// Since we already have it instantiated, we can pipe the routing middleware as
// an instance:
$app->pipe($routes);

$app->pipe(ImplicitHeadMiddleware::class);
$app->pipe(ImplicitOptionsMiddleware::class);

$app->pipe(DispatchMiddleware::class);
$app->pipe(NotFoundMiddleware::class);

// config/routes.php:
$routes->get('/', HomeMiddleware::class);
$routes->post('/api/books', $factory->pipeline(
    BookValidationMiddleware::class,
    CreateBookMiddleware::class
));
```

Essentially, separating these concerns further means we can have more type-safety, less overall code, and more flexibility for users.

The proxy class provides both a migration tool, but also potentially a _usability layer_ for developers who do not want to always use the factory to prepare middleware for piping or routing, but instead just provide the arguments.

xtreamwayz:
About everything you write sounds good. I'm not sure about the `ApplicationRunner` though. I mean right now I need to grab the application from the container and execute `$app->run()`. That's pretty easy. How would this work with the `ApplicationRunner`? On the other hand, if the complexity of the Application class itself is reduced that might be worth it.

The proxy class for backward compatibility sounds good as well.

weierophinney:
> I mean right now I need to grab the application from the container and execute $app->run().

The `ApplicationRunner` would compose the `Application` instance as well, and expose `run()`:

```php
$runner = $container->get(ApplicationRunner::class);
$runner->run();
```

Internally, `run()` would marshal the request and response (if not provided), call on the composed application's `handle()` method (passing it the request), and then pass the response to the emitter.

Xerkus:
May i suggest a following change:

```php
(require 'config/pipeline.php')($app, $factory, $router);
(require 'config/routes.php')($app, $factory, $router);
```
```php

use function Zend\Stratigility\path;

return function (Application $app, MiddlewareFactory $factory, RouteMiddleware $routes) {
    $app->pipe(ErrorMiddleware::class);
    $app->pipe(path('/api', $factory->pipeline(
        CorsMiddleware::class,
        AuthenticationMiddleware::class,
        AuthorizationMiddleware::class
    ));
    
    // Since we already have it instantiated, we can pipe the routing middleware as
    // an instance:
    $app->pipe($routes);
    
    $app->pipe(ImplicitHeadMiddleware::class);
    $app->pipe(ImplicitOptionsMiddleware::class);
    
    $app->pipe(DispatchMiddleware::class);
    $app->pipe(NotFoundMiddleware::class);
}
```

weierophinney:
> May i suggest a following change:

I really like that idea!

Xerkus:
> We could then create a new class, perhaps an ApplicationRunner, that could compose the following:
> 
>    - An emitter. The container would define an EmitterInterface service that resolves to an EmitterStack with the SapiEmitter composed by default.
>    - A ServerRequestErrorResponseGenerator service; this would be used when unable to marshal the server request, in order to generate the response. It should be capable of receiving only an exception. (A factory could close over the ErrorResponseGenerator in order to provide a default ServerRequest and response prototype, so as to allow re-use of existing functionality.)

This is a good functionality to be split into separate package to be used with any request handler.
In zend-mvc refactoring for PSR I made Application a request handler and could reuse ApplicationRunner and emitter stack.

src/Container/NotFoundMiddlewareFactory.php
webimpress: @weierophinney actually I was wrong, sorry, please see my comment there:
https://github.com/zendframework/zend-expressive-aurarouter/pull/30#discussion_r164587191

src/Container/NotFoundMiddlewareFactory.php
webimpress: So as name is changed maybe we should also update year range then?

src/Container/DispatchMiddlewareFactory.php
webimpress: Sorry guys, actually I was wrong... Probably we should keep the line. See my comments there:
https://github.com/zendframework/zend-expressive-aurarouter/pull/30#discussion_r164587191

test/Container/NotFoundMiddlewareFactoryTest.php
webimpress: Did you use `git mv` to rename the file? We are loosing the connection between new and old file, it looks like new file...
Shouldn't we update year range as the file is updated (renamed)?

Xerkus: best way is to move file 100% unchanged, and then edit it in another commit as git does not really have "move" type of change. Simple diff might not detect it anyway as it uses similarity threshold. It is more for a 3-way merge benefit.

xtreamwayz:
> May i suggest a following change:

You have shown me this before. I really like it.

test/Container/NotFoundMiddlewareFactoryTest.php
xtreamwayz: ... or use PhpStorm :D 

It keeps track of moved files (most of the time) so you can move and edit it in the same commit. Not sure how to do that in git. But as PhpStorm can do it, it must be doable with git as well.

weierophinney:
> You have shown me this before. I really like it.

The pattern used to be very common in node, before the `import` construct was added for ES6. As such, it's quite familiar to me; just not in a _PHP_ context!

test/Container/NotFoundMiddlewareFactoryTest.php
weierophinney: I did a `git mv`, but then also added the new file in the same commit, which is why this happened.

Note to self: be more granular in commits.

ezimuel:
@weierophinney I really like the idea of having routes specified on a `RouteMiddleware` instance (`$routes`). In this way the routing configuration is related to a route object and not the application itself.

I see the need for a `path()` function that returns a middleware but I would like to simplify the $factory usage for the middleware pipeline. IMHO, the syntax that you proposed it's too verbose:

```php
$app->pipe(path('/api', $factory->pipeline(
    CorsMiddleware::class,
    AuthenticationMiddleware::class,
    AuthorizationMiddleware::class
));
```

I think we should continue to support **array** as a way to specify a pipeline. I think this is a very intuitive idea to think about an **ordered sequence of middleware** (a pipeline). I would like to propose this change:

```php
$app->pipe(path('/api, [
    CorsMiddleware::class,
    AuthenticationMiddleware::class,
    AuthorizationMiddleware::class
]);
```

We can accomplish this managing the array of middleware in the `path()` function itself, without changing the `Application::pipe` function. The same idea can be used for the various `get`, `post`, `put`, ... functions of routes. Again, the usage of array is a very intuitive way to define a pipeline, and people using Expressive are already familiar with that, we should not change this API.

This routing is very intuitive, even for people not familiar with Expressive:

```php
$routes->post('/api/books', [
    BookValidationMiddleware::class,
    CreateBookMiddleware::class
]);
```
Regarding the other proposals, I'm in favor of it. It will be awesome to see a complete skeleton application to analyze the changes from a user API point of view.

weierophinney:
I'm just about done handling the changes [as suggested in my previous large design comment](https://github.com/zendframework/zend-expressive/pull/543#issuecomment-361386657), and in doing so, discovered something interesting: separating the routing concerns, removing the application running concerns, and making the `pipe()` method strongly typed means that the `Application` class is essentially just a `MiddlewarePipe` instance.

As such, I'm re-creating the `Application` class to compose the following:

- a `MiddlewareFactory` instance
- a `MiddlewarePipeInterface` implementation
- a `RouteMiddleware` instance
- an `ApplicationRunner` instance

The class then implements both `RequestHandlerInterface` and `MiddlewareInterface`, and keeps most public methods from the current v2 implementation (minus `getContainer()`, `getEmitter()`, and `getDefaultHandler()` methods). The methods it exposes all proxy to one or more of the above classes.

What's interesting about this approach is that it keeps the existing signatures and API _entirely_.

This then moves the discussion to: how do we want to write the skeleton?

It could literally stay exactly the same as it is currently. Alternately, we could do like @Xerkus suggested above, and have the routes and pipeline files return callbacks that accept specific arguments (likely the middleware factory, and then one of either the middleware pipeline or route middleware).

What's interesting to me is that this refactoring exercise leaves us with far more flexibility; simpler, more maintainable code; stronger typing; and choice in how we have users consume the project.

Once the changes are complete, I'll push here.

src/MiddlewareFactory.php
weierophinney: I've since changed this to accept only a `MiddlewareContainer` instance.

weierophinney:
All major refactors are now complete. The only bits I'm mulling over at this time are:

- Moving the functionality of `ApplicationConfigInjectionTrait` into another class using static methods: `ApplicationConfigInjector::injectPipelineFromConfig(Application $application, array $config)`, etc.
- Removing the `MiddlewareFactory::path()` method.

Thoughts on these proposals and/or the shape of the refactor at this time?

src/MiddlewareFactory.php
```diff
@@ -0,0 +1,140 @@
...
+        return $pipeline;
+    }
+
+    /**
+     * Segregate one or more middleware by path.
+     *
+     * Creates and returns a PathMiddlewareDecorator instance after first
+     * passing the $middleware argument to prepare().
+     *
+     * @param string|array|MiddlewareInterface $middleware
```
Xerkus: missing @param for first parameter

src/Container/ApplicationRunnerFactory.php
```diff
@@ -0,0 +1,30 @@
...
+use Zend\Expressive\ServerRequestErrorResponseGenerator;
+
+class ApplicationRunnerFactory
+{
+    public function __invoke(ContainerInterface $container) : ApplicationRunner
+    {
+        return new ApplicationRunner(
+            $container->get(Application::class),
+            $container->get(EmitterInterface::class),
+            $container->get(ServerRequestFactory::class),
```
Xerkus: ServerRequestFactory and ServerRequestErrorResponseGenerator are not real classes, shouldn't they be used as string literals?

src/MiddlewareFactory.php
weierophinney: This is standard across major IDEs, at this point, and phpdocumentor infers types from typehints as well. We've already decided for zend-coding-standard to require annotations only if they're adding information, or the type is unspecified.

src/Container/ApplicationRunnerFactory.php
weierophinney: I go back and forth on this, particularly when the service namespace doesn't match the current one.

We can address later.

src/Middleware/RouteMiddleware.php
```diff
@@ -72,4 +81,127 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
 
         return $handler->handle($request);
     }
+
+    /**
+     * Add a route for the route middleware to match.
+     *
+     * Accepts either a Route instance, or a combination of a path and
+     * middleware, and optionally the HTTP methods allowed.
```
Xerkus: This description is wrong, Route instance is not accepted

danizord:
@weierophinney 
>Removing the MiddlewareFactory::path() method.

I'm 👍 for this. IMHO the `MiddlewareFactory` should focus only on creating middleware from `string|callable|array`, otherwise we would end up adding various other methods for each decorator like `host()` `doublePass()` and such.

For the shape of the refactor, I need to review it with more time tomorrow, if you can wait one more day before merging. I missed the notification of this PR. 😄 

Just a few comments from what I've read so far:

* I love the idea of extracting all the `Application` behaviors into separate participants and making them usable outside of `Application`. That makes Expressive even more flexible and composable. Now I can switch to another middleware pipeline implementation, use a 3rd party router/dispatch middleware, or even build my own framework on top of Expressive components, which is great :)
* What about renaming `ApplicationRunner` to `HttpServer` or something like that? Also, if you could extract it to a separate package, it would be really nice :)
* Please rename `MiddlewareFactory:: prepare()` to `__invoke()` so that we can use it like a function: `$middleware(function ($req, $handler) {...})`.
* Not sure, but I think route configuration methods (`post()`, `get()`, ...) could be moved to `Zend\Expressive\Router\RouterInterface` directly. That would simplify `RouterMiddleware` implementation and allow using a custom router middleware that consumes the Expressive router but behaves diferently. For instance, I'd like to have a router that returns 404 response directly instead of calling the next middleware.

btw, congrats for the great work done in PSR-15 and Expressive 3 :)

weierophinney:
>     What about renaming ApplicationRunner to HttpServer or something like that? Also, if you could extract it to a separate package, it would be really nice :)

@Xerkus has indicated he could use this for the zend-mvc refactor as well, so we will definitely extract it. My thought is to also extract the emitter interface and emitters from zend-stratigility either into their own package, or into the same package as the runner (which will consume them).

>     Please rename MiddlewareFactory:: prepare() to __invoke() so that we can use it like a function: $middleware(function ($req, $handler) {...}).

This makes usage more difficult _within_ the `Application` class and `ApplicationConfigInjectionTrait`, however, as we either need to wrap a property  (`($this->factory)($middleware)`) or call it directly (`$this->factory->__invoke($middleware)`).

I'll think on it.

>     Not sure, but I think route configuration methods (post(), get(), ...) could be moved to Zend\Expressive\Router\RouterInterface directly. That would simplify RouterMiddleware implementation and allow using a custom router middleware that consumes the Expressive router but behaves diferently. For instance, I'd like to have a router that returns 404 response directly instead of calling the next middleware.

The `RouterInterface` is purposely small, to make it simpler to use. Because it's an interface, we cannot require features such as route duplication detection, which is currently something the `Application` class handles (and, in this patch, the `RouteMiddeware`. Additionally, the routing methods we expose are around _path-based routing_ specifically; routing _could_ happen around such things as the scheme, host/authority, and even query parameters, as routes are passed a PSR-7 request itself.

My thought is that we could bring the `RouteMiddleware` from this patch into the zend-expressive-router package, and maybe call it `PathBasedRoutingMiddleware` to indicate that it's specific to that purpose. It could even expose a method to add `Route` instances directly, for those cases when you want to add routes that are not examining the path.

If we were to go that route, I'd suggest:

- A new minor release of zend-expressive-router that brings over the current v2 `RouteMiddleware` as-is, but renaming it as suggested above, and providing a factory for it.
- A new minor release of zend-expressive that has `RouteMiddleware` extend the zend-expressive-router version.
- In zend-expressive-router v3, we add the routing methods as we have them in this patch.
- In zend-expressive v3, we remove `RouteMiddleware` entirely, and consume the middleware from zend-expressive-router.

Thoughts on that approach, @danizord?

src/Application.php
```diff
@@ -1,112 +1,133 @@
...
+     * @var ApplicationRunner
      */
-    private $emitter;
+    private $runner;
+
+    public function __construct(
+        MiddlewareFactory $factory,
+        MiddlewarePipeInterface $pipeline,
+        Middleware\RouteMiddleware $routes,
+        ApplicationRunner $runner
```
Xerkus: This makes Application and ApplicationRunner a circular dependency

weierophinney: No, it doesn't; the `MiddlewarePipe` is what is passed to the runner in this case.

weierophinney: The circular dependency is due to how the factories are constructed; the `ApplicationRunner` pulls the `Application` service, but the `Application` service pulls the `ApplicationRunner`.

I'll figure out a solution to this shortly and push the changes.

weierophinney: Fixed with 38c6ff5

src/Container/ApplicationFactory.php
```diff
@@ -1,92 +1,43 @@
 <?php
 /**
  * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
- * @copyright Copyright (c) 2015-2017 Zend Technologies USA Inc. (https://www.zend.com)
+ * @copyright Copyright (c) 2018 Zend Technologies USA Inc. (https://www.zend.com)
```
webimpress: WE should keep year range here and just update to 2018.

src/Container/ApplicationPipelineFactory.php
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Zend\Expressive\Container;
+
+use Psr\Container\ContainerInterface;
+use Zend\Stratigility\MiddlewarePipe;
+use Zend\Stratigility\MiddlewarePipeInterface;
+
+class ApplicationPipelineFactory
+{
+    public function __invoke(ContainerInterface $container) : MiddlewarePipeInterface
```
webimpress: Hm... We can use here just `MiddlewarePipe` unless we'd like to have this class extensible and return different implementation of that interface?

src/Container/ApplicationRunnerFactory.php
```diff
@@ -0,0 +1,50 @@
...
+ *   instance.
+ * - Zend\Expressive\ServerRequestErrorResponseGenerator, which should resolve
+ *   to a PHP callable that accepts a Throwable argument, and which will return
+ *   a Psr\Http\Message\ResponseInterface instance.
+ *
+ * It also consumes the service Zend\Diactoros\Response\EmitterInterface.
+ */
+class ApplicationRunnerFactory
+{
+
```
webimpress: Remove empty line please.

src/Container/EmitterFactory.php
```diff
@@ -0,0 +1,25 @@
...
+namespace Zend\Expressive\Container;
+
+use Psr\Container\ContainerInterface;
+use Zend\Diactoros\Response\EmitterInterface;
+use Zend\Diactoros\Response\SapiEmitter;
+use Zend\Expressive\Emitter\EmitterStack;
+
+class EmitterFactory
+{
+    public function __invoke(ContainerInterface $container) : EmitterInterface
```
webimpress: I would go with `EmitterStack` as return type (unless we'd like to have this class extensible and return another implementation of `EmitterInterface`).

Just thinking - maybe we should have `final` factories?

src/Container/NotFoundHandlerFactory.php
```diff
@@ -1,7 +1,7 @@
 <?php
 /**
  * @see       https://github.com/zendframework/zend-expressive for the canonical source repository
- * @copyright Copyright (c) 2016-2017 Zend Technologies USA Inc. (https://www.zend.com)
+ * @copyright Copyright (c) 2018 Zend Technologies USA Inc. (https://www.zend.com)
```
webimpress: Should we keep year range here?

src/Container/ServerRequestFactoryFactory.php
```diff
@@ -0,0 +1,21 @@
...
+namespace Zend\Expressive\Container;
+
+use Psr\Container\ContainerInterface;
+use Zend\Diactoros\ServerRequestFactory;
+
+class ServerRequestFactoryFactory
+{
+    public function __invoke(ContainerInterface $container) : callable
+    {
+        return [ServerRequestFactory::class, 'fromGlobals'];
```
webimpress: Yes, this is a callback, but this notation is very confusing for IDEs. Not sure if PHPStorm has anything to find usages also in that notation. Maybe we should use callback as:
```php
return function () {
    return ServerRequestFactory::fromGlobals();
};
```

src/MiddlewareContainer.php
```diff
@@ -0,0 +1,64 @@
...
+    {
+        if (! $this->has($service)) {
+            throw Exception\MissingDependencyException::forMiddlewareService($service);
+        }
+
+        $middleware = $this->container->has($service)
+            ? $this->container->get($service)
+            : new $service();
+
+        if (! $middleware instanceof MiddlewareInterface) {
```
webimpress: As we use PHP 7.1 and we have defined return type in method signature maybe we don't need to have this condition here?

test/Application/ConfigInjectionTest.php
```diff
@@ -99,6 +134,44 @@ public static function assertPipelineContainsInstanceOf($class, $pipeline, $mess
...
+        if ($middleware instanceof Middleware\RouteMiddleware) {
+            Assert::that($middleware instanceof Middleware\RouteMiddleware);
+            return;
+        }
+
+        if (! $middleware instanceof Middleware\LazyLoadingMiddleware) {
+            Assert::fail('Middleware is not an instance of RouteMiddleware');
+        }
+
+        Assert::assertAttributeSame(
```
webimpress: Why `Assert::` not `self::` or `$this->` ?

test/Application/ConfigInjectionTest.php
```diff
@@ -99,6 +134,44 @@ public static function assertPipelineContainsInstanceOf($class, $pipeline, $mess
         Assert::assertThat($found, Assert::isTrue(), $message);
     }
 
+    public function assertRouteMiddleware(MiddlewareInterface $middleware)
+    {
+        if ($middleware instanceof Middleware\RouteMiddleware) {
+            Assert::that($middleware instanceof Middleware\RouteMiddleware);
```
webimpress:
```
Method 'that' not found in \PHPUnit\Framework\Assert 
Referenced method is not found in subject class.
```

Do you mean `self::assertInstanceOf`?

test/Application/ConfigInjectionTest.php
```diff
@@ -99,6 +134,44 @@ public static function assertPipelineContainsInstanceOf($class, $pipeline, $mess
         Assert::assertThat($found, Assert::isTrue(), $message);
     }
 
+    public function assertRouteMiddleware(MiddlewareInterface $middleware)
```
webimpress: Other custom assertion in that file are `static`.

test/Application/ConfigInjectionTest.php
```diff
@@ -99,6 +134,44 @@ public static function assertPipelineContainsInstanceOf($class, $pipeline, $mess
...
+
+        Assert::assertAttributeSame(
+            Middleware\RouteMiddleware::class,
+            'middlewareName',
+            $middleware,
+            'Middleware is not an instance of RouteMiddleware'
+        );
+    }
+
+    public function assertDispatchMiddleware(MiddlewareInterface $middleware)
```
webimpress: The same comments for this assertion as for above one.

danizord:
>My thought is that we could bring the RouteMiddleware from this patch into the zend-expressive-router package, and maybe call it PathBasedRoutingMiddleware to indicate that it's specific to that purpose.

I like that approach. In this case I'd suggest moving the `DispatchMiddleware` as well, since both middlewares are really coupled and meant to work together.

Also, maybe that's a subject for another discussion, but it would be great if zend-expressive-router v3 accepted `RequestHandlerInterface` as route handler. I can't imagine any valid use case for a route handler to delegate the processing to another middleware. Also, it would make testing our "actions" much easier by removing the need to mock the handler and assert that it is never called.

weierophinney:
@webimpress —

> I'd like to have also arrays for pipeline, like suggested by @ezimuel and me before on the forum:
> 
> $routes->post('/api/books', [
>     BookValidationMiddleware::class,
>     CreateBookMiddleware::class
> ]);

This is accomplished already via the `Application` class. The reason for _not_ doing it in the `RouteMiddleware` is that it requires an additional collaborator, the `MiddlewareFactory`, to accomplish, and this adds quite a lot more complexity. Additionally, if we export the `RouteMiddleware` to zend-expressive-router, this would mean we'd need to also extract the `MiddlewareFactory` to its own package, and then things get incredibly complex in terms of package relationships (as if they aren't already!).

Honestly, I think this is the cleanest approach we have which can also support the existing syntax.

> Currently I can see the same 'routingmethods inRouteMiddlewareandApplication(I meanroute/get/post/patch/...). Do we still need them inApplication? As I've seen on suggested examples you'd like to bind routed middleware directly to$routes(RouteMiddleware) so why we still need them inApplication`?

Yes, still needed, in order to allow the various syntaxes currently allowed in v2, and to keep the complexity of `RouteMiddleware` at a manageable level. These methods use the `MiddlewareFactory` to ensure the middleware passed to the `RouteMiddleware` methods are valid PSR-15 middleware, per my above comment.

src/Container/ApplicationPipelineFactory.php
weierophinney: Bingo! This is to allow developers to extend the factory if desired; as long as what they return is a `MiddlewarePipeInterface`, the extension remains valid.

src/Container/EmitterFactory.php
weierophinney: Again, same reasons as for `ApplicationPipelineFactory`. While an `EmitterStack` is the default, the idea is to note that any emitter is valid for this purpose.

In terms of `final` factories: sure, we could go that way. If we do, I'll update the typehints to the more specific classes returned.

src/Container/NotFoundHandlerFactory.php
weierophinney: No, because in this case, it's referencing a brand new class, so this factory is technically new.

src/Container/ServerRequestFactoryFactory.php
weierophinney: That ends up being two function invocations, however. And with PHP 5.6 and up,  you can call any callable using `$callable()` notation.

src/MiddlewareContainer.php
weierophinney: The point is to give a more specific error and actionable information the user can consume in order to fix the information. A `TypeError` will not likely indicate easily that the value fetched from a container was incorrect.

test/Application/ConfigInjectionTest.php
weierophinney: Good catch; fixing.

test/Application/ConfigInjectionTest.php
weierophinney: Missed that; fixing now.

test/Application/ConfigInjectionTest.php
weierophinney: We cannot use `$this->` in static context (thanks for pointing out that change!).

Otherwise, `Assert::` is consistent with the existing custom assertion methods, and what the PHPUnit manual suggests for custom assertions.

danizord:
@weierophinney I see. At least we could accept `MiddlewareInterface|RequestHandlerInterface` :)

weierophinney:
> At least we could accept MiddlewareInterface|RequestHandlerInterface :)

I'll consider that approach as well, though it has impact on the `DispatchMiddleware`, too. :smile:

src/Container/NotFoundHandlerFactory.php
```diff
@@ -10,13 +10,23 @@
...
+        $config   = $container->has('config') ? $container->get('config') : [];
+        $renderer = $container->has(TemplateRendererInterface::class)
+            ? $container->get(TemplateRendererInterface::class)
+            : null;
+        $template = $config['zend-expressive']['error_handler']['template_404']
+            ?? NotFoundHandler::TEMPLATE_DEFAULT;
+        $layout = $config['zend-expressive']['error_handler']['layout']
+            ?? NotFoundHandler::LAYOUT_DEFAULT;
+
+        return new NotFoundHandler(new Response(), $renderer, $template, $layout);
```
danizord: Should pull `$container->get(ResponseInterface::class)` here?

test/IntegrationTest.php
```diff
@@ -1,118 +0,0 @@
...
-use Psr\Http\Message\ResponseInterface;
-use Psr\Http\Message\ServerRequestInterface;
-use Zend\Diactoros\Response;
-use Zend\Diactoros\Response\EmitterInterface;
-use Zend\Diactoros\ServerRequest;
-use Zend\Expressive\Application;
-use Zend\Expressive\Delegate\NotFoundDelegate;
-use Zend\Expressive\Router\FastRouteRouter;
-
-class IntegrationTest extends TestCase
```
danizord: Can be reverted?

src/ApplicationConfigInjectionTrait.php
```diff
@@ -93,7 +88,11 @@ public function injectPipelineFromConfig(array $config = null) : void
 
         foreach ($queue as $spec) {
             $path = $spec['path'] ?? '/';
-            $this->pipe($path, $spec['middleware']);
+            $this->pipe(
+                $path === '/'
+                    ? $spec['middleware']
+                    : path($path, $this->factory->prepare($spec['middleware']))
```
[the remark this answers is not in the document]
weierophinney: Good point; thanks!

src/Container/NotFoundHandlerFactory.php
weierophinney: Yes - thanks!

test/IntegrationTest.php
weierophinney: This particular test case is no longer relevant following the refactor.

weierophinney:
Okay, last few things to do before I think this is ready:

- Extract the `Zend\Stratigility\Response\Emitter*` classes and the `ApplicationRunner` to their own package. Once that is done, I can update this package to depend on that one, and define a factory for the runner class it defines.
- Extract the `RouteMiddleware` and `DispatchMiddleware` to the zend-expressive-router package. At that point, I will need to tag that package with a new alpha release, update this one to consume the package, and update the factories for each accordingly. The `RouteMidldeware` will be renamed to `PathBasedRouteMiddleware` to make it clear it is a specific approach to routing (i.e., doesn't do non-path-related matching).
- Determine if we can and/or should allow request handlers for routes either as a substitute for or in addition to middleware, and implementing it if it is feasible.

Once these tasks are done, and if everyone agrees on this patch, I will do the following:

- Merge this patch.
- Create a patch against master that splits the documentation into multiple versions.
- Create v3 documentation:
  - Updating it to reflect the new API
  - Create a migration document
- Create a patch against develop:
  - Deprecating features removed by this patch
  - Backporting features that can help in migration
  - Documenting the deprecations and steps developers can take to make their applications forwards-proof.

From there, we can tag a v2.2 version of each of zend-stratigility and this repository, and then start preparing for the v3 release.

src/Container/ApplicationConfigInjectionDelegator.php
```diff
@@ -1,19 +1,49 @@
...
 {
+    /**
+     * Decorate an Application instance by injecting routes and/or middleware
+     * from configuration.
+     */
+    public function __invoke(ContainerInterface $container, string $serviceName, callable $callback) : Application
+    {
+        $application = $callback();
+        if (! $application instanceof Application) {
+            return $application;
```
webimpress: It looks like we will have here TypeError, because method declaration has `Application` return type and we return here `$application` which is not instance of `Application`.

src/ConfigProvider.php
```diff
@@ -0,0 +1,63 @@
...
+                MiddlewareContainer::class                 => Container\MiddlewareContainerFactory::class,
+                MiddlewareFactory::class                   => Container\MiddlewareFactoryFactory::class,
+                Middleware\DispatchMiddleware::class       => Container\DispatchMiddlewareFactory::class,
+                Middleware\NotFoundMiddleware::class       => Container\NotFoundMiddlewareFactory::class,
+                Middleware\RouteMiddleware::class          => Container\RouteMiddlewareFactory::class,
+                ResponseInterface::class                   => Container\ResponseFactory::class,
+                ServerRequestErrorResponseGenerator::class => Container\ServerRequestErrorResponseGeneratorFactory::class,
+                ServerRequestFactory::class                => Container\ServerRequestFactoryFactory::class,
+            ],
+            'shared' => [
```
webimpress: Is it going to work with different dependency injection containers (pimple/aura.di)?

weierophinney: I'm not sure; I need to review how each of those work. I know most of them have a way to always return a new instance for specified services, but I'm not sure if we have addressed that in our wrappers.

src/Container/ApplicationConfigInjectionDelegator.php
weierophinney: Good point; I'll remove the typehint from the method.

weierophinney: Done!

PR #544, opened 2018-02-06
Expressive 3.0 fixes

webimpress:
Fixes:

- added constants visibility (to `public`)
- removed unnecessary type casting
- removed default property value
- optimized imports


weierophinney:
Thanks, @webimpress!

PR #546, opened 2018-02-06
Removes concept of "default handler"

### LINKED ISSUE #441: NotFoundDelegate and NotFoundHandler
While upgrading to Expressive 2 I found these new `NotFoundDelegate` and `NotFoundHandler` a bit confusing. :confused: 

The 2.0 docs suggests using `NotFoundDelegate` as default delegate and piping `NotFoundHandler` as the ~outermost~ (EDIT: inner) middleware, which makes no sense for me since `NotFoundHandler` uses `NotFoundDelegate` internally and `NotFoundDelegate` would be called anyway if `NotFoundHandler` was not piped.

As such, to avoid confusion I'd suggest removing `NotFoundHandler` completely and recommend usage of `NotFoundDelegate` only.

Thoughts?

weierophinney:
Because the `MiddlewarePipe` does not compose a "default handler", and because the `MiddlewarePipe` is what the request handler runner dispatches, there is no longer a concept of a "default delegate" (v2) or "default handler" (v3); there is only middleware.

As such, this patch merges the `NotFoundHandler` into the `NotFoundMiddleware`.

A key is kept for the `DefaultDelegate` service, but it will now resolve to the `NotFoundMiddleware`.

Fixes #441.


test/Container/NotFoundMiddlewareFactoryTest.php
```diff
@@ -10,23 +10,74 @@
...
+    private $container;
+
+    protected function setUp()
     {
-        $handler   = $this->prophesize(NotFoundHandler::class)->reveal();
-        $container = $this->prophesize(ContainerInterface::class);
-        $container->get(NotFoundHandler::class)->willReturn($handler);
-        $factory = new NotFoundMiddlewareFactory();
+        $this->container = $this->prophesize(ContainerInterface::class);
+        $this->response = $this->prophesize(ResponseInterface::class)->reveal();
```
webimpress: I'll move this line before `$this->container` to have container configuration together.
And I would add `private` property in the class.

CHANGELOG.md
```diff
@@ -97,18 +97,19 @@ All notable changes to this project will be documented in this file, in reverse
 
   and removes the dependency http-interop/http-server-middleware.
 
-- [#543](https://github.com/zendframework/zend-expressive/pull/543) renames
+- [#543](https://github.com/zendframework/zend-expressive/pull/543) and
+  [#546](https://github.com/zendframework/zend-expressive/pull/546) renames the
```
webimpress: This change is in `3.0.0alpha2` changelog but milestone for this PR is set to `3.0.0alpha3`.

weierophinney: Crap, you're right. I'll update.

weierophinney: Essentially, I'll be merging all the alpha logs to a single 3.0.0 log later, but for now, needs to be slightly different.

PR #547, opened 2018-02-07
Remove disabling shared ResponseInterface

xtreamwayz:
The disable shared service feature is not in the PSR-11 spec so we can't rely on it.
This PR provides another solution by requiring different ResponseInterfaces where needed.

See: https://github.com/zendframework/zend-expressive/pull/543#discussion_r165378017

This PR would make the following PR's not needed:
https://github.com/zendframework/zend-servicemanager/pull/248
https://github.com/zendframework/zend-pimple-config/pull/6


xtreamwayz:
So if someone uses MyAwesomeContainer, and that container shares by default and you can't disable it, it means that that container would be incompatible with Expressive 3.0 without this PR. With this PR all PSR-11 containers are compatible and the application should work as expected.

> As we deliver zend-pimple-config library it would be nice to have there the same functionality as we have in servicemanager, so configure it should configure pimple using servicemanager configuration that the final container behaves the same as servicemanager with the same configuration.

I agree. But since we can add functionality to zend-pimple-config as we own that package. But we can't do that for MyAwesomeContainer as we don't have any influence on it or even should try to force it.

webimpress:
@xtreamwayz 
> So if someone uses MyAwesomeContainer, and that container shares by default and you can't disable it, it means that that container would be incompatible with Expressive 3.0 without this PR. With this PR all PSR-11 containers are compatible and the application should work as expected.

Have you tested expressive 3.0-dev with container which has DISABLED sharing services? Please note that servicemanager and current pimple config shares services by default, but it's not behaviour specified by PSR-11.

> I agree. But since we can add functionality to zend-pimple-config as we own that package. But we can't do that for MyAwesomeContainer as we don't have any influence on it or even should try to force it.

This is why I'd like to have tests in separate repository and require this behaviour to be compatible with expressive. The same we do with routers/template renderers, we require some behaviour to be compatible with expressive.

src/ConfigProvider.php
```diff
@@ -43,22 +43,19 @@ public function getDependencies() : array
...
                 MiddlewareFactory::class                   => Container\MiddlewareFactoryFactory::class,
+                // Change the following in development to the WhoopsErrorResponseGeneratorFactory:
+                Middleware\ErrorResponseGenerator::class   => Container\ErrorResponseGeneratorFactory::class,
                 Middleware\NotFoundMiddleware::class       => Container\NotFoundMiddlewareFactory::class,
+                NotFoundResponseInterface::class           => Container\ResponseFactory::class,
                 RequestHandlerRunner::class                => Container\RequestHandlerRunnerFactory::class,
-                ResponseInterface::class                   => Container\ResponseFactory::class,
                 Router\DispatchMiddleware::class           => Container\DispatchMiddlewareFactory::class,
                 Router\PathBasedRoutingMiddleware::class   => Container\RouteMiddlewareFactory::class,
+                RouterResponseInterface::class             => Container\ResponseFactory::class,
```
weierophinney: I wonder if we should make all of these virtual responses match the service they correspond to:

- `NotFoundMiddlewareResponseInterface`
- `RouteMiddlewareResponseInterface`

src/Container/NotFoundMiddlewareFactory.php
```diff
@@ -10,8 +10,8 @@
 namespace Zend\Expressive\Container;
 
 use Psr\Container\ContainerInterface;
-use Psr\Http\Message\ResponseInterface;
 use Zend\Expressive\Middleware\NotFoundMiddleware;
+use Zend\Expressive\NotFoundResponseInterface;
```
weierophinney: Should be `Zend\Expressive\Response\NotFoundResponseInterface` (or the name I proposed; main problem is namespace).

src/Container/RouteMiddlewareFactory.php
```diff
@@ -10,17 +10,17 @@
 namespace Zend\Expressive\Container;
 
 use Psr\Container\ContainerInterface;
-use Psr\Http\Message\ResponseInterface;
 use Zend\Expressive\Router\PathBasedRoutingMiddleware;
 use Zend\Expressive\Router\RouterInterface;
+use Zend\Expressive\RouterResponseInterface;
```
weierophinney: Should be `Zend\Expressive\Response` namespace, with appropriate name.

test/ConfigProviderTest.php
```diff
@@ -10,22 +10,22 @@
...
 use PHPUnit\Framework\TestCase;
-use Psr\Http\Message\ResponseInterface;
 use Zend\Expressive\Application;
 use Zend\Expressive\ApplicationPipeline;
 use Zend\Expressive\ConfigProvider;
 use Zend\Expressive\Delegate\DefaultDelegate;
 use Zend\Expressive\Middleware;
 use Zend\Expressive\MiddlewareContainer;
 use Zend\Expressive\MiddlewareFactory;
+use Zend\Expressive\NotFoundResponseInterface;
```
weierophinney: Should be under `Zend\Expressive\Response` namespace.

test/ConfigProviderTest.php
```diff
@@ -10,22 +10,22 @@
...
 use Zend\Expressive\Router\DispatchMiddleware;
 use Zend\Expressive\Router\PathBasedRoutingMiddleware;
+use Zend\Expressive\RouterResponseInterface;
```
weierophinney: Should be under `Zend\Expressive\Response` namespace.

test/Container/NotFoundMiddlewareFactoryTest.php
```diff
@@ -15,6 +15,7 @@
 use Psr\Http\Message\ResponseInterface;
 use Zend\Expressive\Container\NotFoundMiddlewareFactory;
 use Zend\Expressive\Middleware\NotFoundMiddleware;
+use Zend\Expressive\NotFoundResponseInterface;
```
weierophinney: Should be under `Zend\Expressive\Response` namespace.

test/Container/RouteMiddlewareFactoryTest.php
```diff
@@ -15,6 +15,7 @@
 use Zend\Expressive\Container\RouteMiddlewareFactory;
 use Zend\Expressive\Router\PathBasedRoutingMiddleware;
 use Zend\Expressive\Router\RouterInterface;
+use Zend\Expressive\RouterResponseInterface;
```
weierophinney: Should be under `Zend\Expressive\Response` namespace.

PR #549, opened 2018-02-07
Make the ServerRequestFactory service compatible with all containers

weierophinney:
We cannot return just `ServerRequestFactory::fromGlobals` or `[ServerRequestFactory::class, 'fromGlobals']` as not all containers allow vanilla PHP callable services. Instead, we wrap it in an anonymous function here, which is allowed by all containers tested at this time.


webimpress:
@weierophinney which container does not support callbacks as array? Maybe it's a problem with the configuration, not with the container itself?

weierophinney:
@webimpress Auryn specifically. And I tried a variety of mechanisms; wrapping it in a closure works across all implementations.

PR #551, opened 2018-02-13
Updates to zend-expressive-router 3.0.0alpha2

weierophinney:
zend-expressive-router 3.0.0alpha2 renames middleware it previously defined, defines new middleware, and creates a number of new package-level constants for virtual services.

This patch:

- **Removes:**
  - the `RouteMiddlewareFactory` and `DispatchMiddlewareFactory`, as these are now provided by the zend-expressive-router package, and registered to the container by that package's config provider.
  - the `Implicit*Middleware` classes and related factories; these are now provided by the zend-expressive-router package.
- **Renames:**
  - the `Middleware\NotFoundMiddleware` to `Handler\NotFoundHandler`, and renames its factory accordingly. The class was acting as a handler, and can be used in more contexts as a handler than as middleware.
- **Adds:**
  - `Zend\Expressive\Container\StreamFactory`, for producing empty, writable stream instances.
  - Package-level constants by which virtual services may be referenced. In all cases, these refer to the FQCN variants previously used to ensure the old values continue to resolve; however, users are encouraged to use the new constants instead, as they will resolve, and because their docblocks will provide context for usage.
- **Changes:**
  - The `ConfigAggregator` was updated to use the new package-level constants, re-map the various aliases to the new services they refer to, and to add entries for various zend-expressive-router virtual services.
  - The `ApplicationConfigInjectionDelegator` now raises an exception if the callback provided to it does not produce an `Application` instance.
  - The `ApplicationConfigInjectionDelegator` changes on exception message to remove details that are not relevant in the 3.0 branch.

These should likely be the last breaking changes for this package before the 3.0.0 release.


src/Container/ApplicationConfigInjectionDelegator.php
```diff
@@ -23,15 +24,20 @@ class ApplicationConfigInjectionDelegator
...
+     *     something other than an `Application` instance, as the delegator cannot
+     *     proceed with its operations.
      */
-    public function __invoke(ContainerInterface $container, string $serviceName, callable $callback)
+    public function __invoke(ContainerInterface $container, string $serviceName, callable $callback) : Application
     {
         $application = $callback();
         if (! $application instanceof Application) {
-            return $application;
+            throw new Exception\InvalidServiceException(sprintf(
```
webimpress: Why not static method in exception?

src/Container/ApplicationConfigInjectionDelegator.php
```diff
@@ -65,8 +71,9 @@ public function __invoke(ContainerInterface $container, string $serviceName, cal
      *     'middleware_pipeline' => [
      *         // An array of middleware to register with the pipeline.
      *         // entries to register prior to routing/dispatching...
-     *         // - entry for \Zend\Expressive\Router\PathBasedRoutingMiddleware::class
-     *         // - entry \Zend\Expressive\Router\DispatchMiddleware::class
+     *         // - entry for \Zend\Expressive\Router\Middleware\PathBasedRoutingMiddleware::class
+     *         // - entry for \Zend\Expressive\Router\Middleware\MethodNotAllowedMiddleware::class
+     *         // - entry \Zend\Expressive\Router\Middleware\DispatchMiddleware::class
```
webimpress: For consistency should we have `entry for...` ?

src/constants.php
```diff
@@ -0,0 +1,73 @@
...
+const ROUTE_MIDDLEWARE = __NAMESPACE__ . '\Middleware\RouteMiddleware';
+
+/**
+ * Virtual service name that should resolve to a service returning a response
+ * based on a `Throwable` argument produced when generating the application
+ * request.
+ *
+ * @var string
+ */
+const SERVER_REQUEST_ERROR_RESPONSE_GENERATOR = __NAMESPACE__ . '\Expressive\ServerRequestErrorResponseGenerator';
```
webimpress: `__NAMESPACE__` is `Zend\Expressive`, should we have `\Expressive` part twice here?

src/constants.php
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * Virtual service name that should resolve to a service capable of producing
+ * a PSR-7 ServerRequestInterface instance for the application.
+ *
+ * @var string
+ */
+const SERVER_REQUEST_FACTORY = __NAMESPACE__ . '\Expressive\ServerRequestFactory';
```
webimpress: As above.

test/Router/IntegrationTest.php
```diff
@@ -103,6 +106,7 @@ private function createApplicationWithGetPost($adapter, $getName = null, $postNa
         $router = new $adapter();
         $app = $this->createApplicationFromRouter($router);
         $app->pipe(new RouteMiddleware($router, $this->response));
```
webimpress: Here and in many other usages - the second parameter should be remove, `RouteMiddleware::__construct` has now one parameter.

src/Container/ApplicationConfigInjectionDelegator.php
weierophinney: This one hasn't had one defined ever. I can add some later, but in most cases, we'll have different messages for every single factory, making it less viable.

src/constants.php
weierophinney: Good catch; fixing!

test/Router/IntegrationTest.php
weierophinney: Another nice catch! Fixing!

PR #557, opened 2018-02-15
$serviceName is not used.

harikt:

- [x] Are you fixing a bug?
  - [ ] Detail how the bug is invoked currently.
  - [ ] Detail the original, incorrect behavior.
  - [ ] Detail the new, expected behavior.
  - [ ] Base your feature on the `master` branch, and submit against that branch.
  - [ ] Add a regression test that demonstrates the bug, and proves the fix.
  - [ ] Add a `CHANGELOG.md` entry for the fix.

The `$serviceName` is never used.

I was trying to test the aplha version of zend-expressive on https://github.com/harikt/dms-expressive-example . I came across a strange issue that there is no longer `injectRoutesFromConfig` in `Application` and recommends using `Zend\Expressive\Container\ApplicationConfigInjectionDelegator`.

I missed to see how to use it, so this is how I modified 

```php
    // config/routes.php
    $delegator = new \Zend\Expressive\Container\ApplicationConfigInjectionDelegator();
    $delegator($container, null, function () use ($app) {
        return $app;
    });
```
I have not seen anything searching on the code base `ApplicationConfigInjectionDelegator` whether it is used. Only one instance of file is seen currently.

So the parameter seems not needed here.

Thank you.

### DIFF
--- a/src/Container/ApplicationConfigInjectionDelegator.php
+++ b/src/Container/ApplicationConfigInjectionDelegator.php
@@ -28,7 +28,7 @@ class ApplicationConfigInjectionDelegator
      *     something other than an `Application` instance, as the delegator cannot
      *     proceed with its operations.
      */
-    public function __invoke(ContainerInterface $container, string $serviceName, callable $callback) : Application
+    public function __invoke(ContainerInterface $container, callable $callback) : Application
     {
         $application = $callback();
         if (! $application instanceof Application) {


xtreamwayz:
It's there for compatibility with zend-servicemanager:

https://github.com/zendframework/zend-servicemanager/blob/master/src/Factory/DelegatorFactoryInterface.php#L23-L39

Implementing the DelegatorFactoryInterface directly would cause an extra zend-servicemanager dependency which is not really needed if you use another container.

harikt:
@xtreamwayz hm... .So do you think we can keep something like `$serviceName = null` here ?

Also I am wondering if the usage I shown regarding this is right or not. Can you confirm the same ?

xtreamwayz:
No you can't use null as you will always have a callback. Without the callback the delegator doesn't work. See: https://github.com/harikt/zend-expressive/blob/1e2bc4bd0fdb9ace75acd2c3f5fa62405d17f845/src/Container/ApplicationConfigInjectionDelegator.php#L33

harikt:
Why do you want to pass the application itself when the container can get the application ?

Looking at the `public/index.php`

```
$app = $container->get(\Zend\Expressive\Application::class);
```

I am not really sure about the need for the callback itself. I have not ever looked at zend service manager nor used it , may be this is only there for the zend service manager ?

As I mentioned the reason I am questioning is the `ApplicationConfigInjectionDelegator` is called inside the project only once by me. There is no other usages I noticed in the search ( Either the search is giving wrong result ) .

xtreamwayz:
I haven't used this myself and I can't find it in the docs. But I think you have to register this in your config:

```php
public function getDependencies()
{
    return [
        'delegators' => [
            \Zend\Expressive\Application::class => [
                \Zend\Expressive\Container\ApplicationConfigInjectionDelegator::class,
            ],
        ],
    ];
}
```

If you register it like that, then it should inject all the configured routes and the pipeline for you. This is also where the DelegatorFactoryInterface is needed if you use zend-servicemanager.

weierophinney:
There are a number of reasons for the argument order and for each argument being present.

The container is present so that the delegator can pull _additional_ dependencies as needed.

The service name is provided to allow users to determine what service was _originally_ requested. This can be used to prevent cyclical lookups, or to compare the service provided by the `$callback` (more on that) to what it was requested for (e.g., to see if the service returned is an instance of the service requested).

The callback is present because of when and how delegators are executed.

Delegators are only executed during initial creation of a service. You provide a _list_ of delegators, and the container is expected to pass the results of one to the next. However, in actual usage, you may not want a delegator to execute if a previous one has raised an exception; alternately, you may want to be able to catch such exceptions and provide a substitute. As  such, we pass a callback to each delegator. Invoking that callback returns the service. 

But it doesn't, actually. It returns the return value _of the previous callback_. In the case of the first delegator, this means it returns the result of a factory or abstract factory. In the case of the second, it returns whatever the first delegator returns.

You can actually think of delegators as _middleware for containers_. The last delegator on the list is executed first, and it calls on the callback to get the result of the previous, and so on. If one of the delegators returns an alternate service without ever calling on the callback passed to it, then those previous layers, including the original factory, are never executed.

This is why you cannot call `$container->get(Application::class)` within the delegator; you're already _within_ such a call, which would lead to a recursive cycle. It's also because a delegator may choose to return a different instance entirely.

xtreamwayz:
> You can actually think of delegators as middleware for containers. 

That's a nice description.

PR #562, opened 2018-02-22
Provide a dedicated service for server request error response generation

weierophinney:
Provide a dedicated service for the server request error response generator

Instead of re-using the existing `ErrorResponseGenerator`, this patch does the following:

- Separates functionality for generating the error response into a trait, `Zend\Expressive\Response\ErrorResponseGeneratorTrait`. The method `prepareTemplatedResponse()` now expects the error, the response, and an array of data to provide to the template.
- Updates `Zend\Expressive\Middleware\ErrorResponseGenerator` to use the new trait.
- Creates a new class, `Zend\Expressive\Response\ServerRequestErrorResponseGenerator`, which expects a response factory, and optionally the debug flag, renderer, and template to render to. It composes the trait, and generates a response on-the-fly to pass to the trait methods.

The `RequestHandlerRunnerFactory` has been updated to use this new service, and the `ServerRequestErrorResponseGeneratorFactory` was updated to generate an instance of the new class.

These changes mean the constant SERVER_REQUEST_ERROR_RESPONSE_GENERATOR is no longer needed, and was thus removed.

### DIFF
--- a/src/Response/ServerRequestErrorResponseGenerator.php
+++ b/src/Response/ServerRequestErrorResponseGenerator.php
@@ -45,10 +45,10 @@ class ServerRequestErrorResponseGenerator
         $this->template  = $template;
     }
 
-    public function __invoke(Throwable $e)
+    public function __invoke(Throwable $e) : ResponseInterface
     {
-        $response = ($this->responseFactory)()
-            ->withStatus(Utils::getStatusCode($e, $response));
+        $response = ($this->responseFactory)();
+        $response = $response->withStatus(Utils::getStatusCode($e, $response));
 
         if ($this->renderer) {
             return $this->prepareTemplatedResponse($e, $response, [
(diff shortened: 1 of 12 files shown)


weierophinney:
@webimpress Please re-review; this removes the constant, and provides a dedicated service.

src/Response/ServerRequestErrorResponseGenerator.php
```diff
@@ -0,0 +1,63 @@
...
+
+        $this->debug     = $isDevelopmentMode;
+        $this->renderer  = $renderer;
+        $this->template  = $template;
+    }
+
+    public function __invoke(Throwable $e)
+    {
+        $response = ($this->responseFactory)()
+            ->withStatus(Utils::getStatusCode($e, $response));
```
webimpress: `$response` is not yet defined so can't be used here.

weierophinney: Good catch - and demonstrates I need to fully test this class. I'll push more commits momentarily.

PR #564, opened 2018-02-22
ensure $path to be string on path() call

samsonasik:

The `Zend\Stratigility\path()` function require 1st parameter to string, while `$path` can be `array|callable|Psr\Http\Server\MiddlewareInterface|Psr\Http\Server\RequestHandlerInterface|string` with compare to `/` string.

### DIFF
--- a/test/ApplicationTest.php
+++ b/test/ApplicationTest.php
@@ -16,6 +16,7 @@ use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Server\MiddlewareInterface;
 use Psr\Http\Server\RequestHandlerInterface;
+use TypeError;
 use Zend\Expressive\Application;
 use Zend\Expressive\MiddlewareFactory;
 use Zend\Expressive\Router\Middleware\PathBasedRoutingMiddleware as RouteMiddleware;
@@ -128,19 +129,18 @@ class ApplicationTest extends TestCase
         $this->assertNull($this->app->pipe('/foo', $middleware));
     }
 
-    public function testPipeNonSlashPathOnNonStringPipe()
+    public function testPipeNonSlashPathOnNonStringPipeProduceTypeError()
     {
         // @codingStandardsIgnoreStart
         $middleware1 = function ($request, $response) {};
         // @codingStandardsIgnoreEnd
         $middleware2 = $this->createMockMiddleware();
 
-        $preparedMiddleware = $this->createMockMiddleware();
-        $this->factory
-            ->prepare($middleware2)
-            ->willReturn($preparedMiddleware);
-
-        $this->assertNull($this->app->pipe($middleware1, $middleware2));
+        try {
+            $this->app->pipe($middleware1, $middleware2);
+        } catch (TypeError $t) {
+            $this->assertInstanceOf(TypeError::class, $t);
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


webimpress:
@samsonasik Can you add test case please?

samsonasik:
@webimpress I've added the test

webimpress:
Hm... in general it works, but I think we should change it. Because passing two middlewares has no sense. I would add another check then - if `$middleware !== null` then `$middlewareOrPath` must be a string, and that test case makes more sens for me.

@samsonasik what do you think?

weierophinney:
`path()` will produce a `TypeError` if `$path` is not a string, which I think is likely far better than silently ignoring the type problem.

webimpress:
@weierophinney so you just suggest to change the test case that it's desired behaviour and not changing the code? It's fine to me 👍

weierophinney:
> so you just suggest to change the test case that it's desired behaviour and not changing the code? Sounds good to me

Yes - add a test case demonstrating the `TypeError` occurs in such a situation.

samsonasik:
@weierophinney Ok, I've updated the test that shows it got `TypeError` that demonstrate the condition.

test/ApplicationTest.php
```diff
@@ -128,6 +129,20 @@ public function testPipeCanAcceptAPathArgument($middleware)
...
+    public function testPipeNonSlashPathOnNonStringPipeProduceTypeError()
+    {
+        // @codingStandardsIgnoreStart
+        $middleware1 = function ($request, $response) {};
+        // @codingStandardsIgnoreEnd
+        $middleware2 = $this->createMockMiddleware();
+
+        try {
+            $this->app->pipe($middleware1, $middleware2);
+        } catch (TypeError $t) {
```
webimpress: Can't we use `$this->expectedException`?

test/ApplicationTest.php
```diff
@@ -128,6 +129,20 @@ public function testPipeCanAcceptAPathArgument($middleware)
         $this->assertNull($this->app->pipe('/foo', $middleware));
     }
 
+    public function testPipeNonSlashPathOnNonStringPipeProduceTypeError()
+    {
+        // @codingStandardsIgnoreStart
+        $middleware1 = function ($request, $response) {};
+        // @codingStandardsIgnoreEnd
```
webimpress: Let's remove the CS ignore and format it in two lines. It is just 2 lines instead of 3.

test/ApplicationTest.php
samsonasik: updated.

test/ApplicationTest.php
samsonasik: updated.

test/ApplicationTest.php
```diff
@@ -128,6 +129,18 @@ public function testPipeCanAcceptAPathArgument($middleware)
         $this->assertNull($this->app->pipe('/foo', $middleware));
     }
 
+    public function testPipeNonSlashPathOnNonStringPipeProduceTypeError()
+    {
+        $this->expectException(TypeError::class);
```
webimpress: Please move this just before line which should raise the exception. Otherwise it's not that clear. I can see that the exception can't be throw in other place, but it's good convention to have it just before the place where it should be thrown.

samsonasik: updated.

weierophinney:
Thanks, @samsonasik!

PR #565, opened 2018-02-22
Minor fixes detected by phpstan

belgattitude:

Few modifications for fixes detected by phpstan

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -104,7 +104,7 @@ class Application implements MiddlewareInterface, RequestHandlerInterface
         $middleware = $middleware ?: $middlewareOrPath;
         $path = $middleware === $middlewareOrPath ? '/' : $middlewareOrPath;
 
-        $middleware = $path !== '/'
+        $middleware = ( is_string($path) && $path !== '/' )
             ? path($path, $this->factory->prepare($middleware))
             : $this->factory->prepare($middleware);
 
--- a/src/Application.php
+++ b/src/Application.php
@@ -104,7 +104,7 @@ class Application implements MiddlewareInterface, RequestHandlerInterface
         $middleware = $middleware ?: $middlewareOrPath;
         $path = $middleware === $middlewareOrPath ? '/' : $middlewareOrPath;
 
-        $middleware = ( is_string($path) && $path !== '/' )
+        $middleware = $path !== '/'
             ? path($path, $this->factory->prepare($middleware))
             : $this->factory->prepare($middleware);
 
--- a/src/Handler/NotFoundHandler.php
+++ b/src/Handler/NotFoundHandler.php
@@ -21,7 +21,7 @@ class NotFoundHandler implements RequestHandlerInterface
     public const LAYOUT_DEFAULT = 'layout::default';
 
     /**
-     * @var TemplateRendererInterface
+     * @var TemplateRendererInterface|null
      */
     private $renderer;
 
@@ -59,15 +59,14 @@ class NotFoundHandler implements RequestHandlerInterface
      * Creates and returns a 404 response.
      *
      * @param ServerRequestInterface $request Passed to internal handler
-     * @param RequestHandlerInterface $handler Ignored.
      */
     public function handle(ServerRequestInterface $request) : ResponseInterface
     {
-        if (! $this->renderer) {
+        if ($this->renderer === null) {
             return $this->generatePlainTextResponse($request);
         }
 
-        return $this->generateTemplatedResponse($request);
+        return $this->generateTemplatedResponse($this->renderer, $request);
     }
 
     /**
@@ -91,11 +90,14 @@ class NotFoundHandler implements RequestHandlerInterface
      *
      * Template will receive the current request via the "request" variable.
      */
-    private function generateTemplatedResponse(ServerRequestInterface $request) : ResponseInterface
-    {
+    private function generateTemplatedResponse(
+        TemplateRendererInterface $renderer,
+        ServerRequestInterface $request
+    ) : ResponseInterface {
+
         $response = ($this->responseFactory)()->withStatus(StatusCodeInterface::STATUS_NOT_FOUND);
         $response->getBody()->write(
-            $this->renderer->render($this->template, ['request' => $request, 'layout' => $this->layout])
+            $renderer->render($this->template, ['request' => $request, 'layout' => $this->layout])
         );
 
         return $response;
--- a/src/MiddlewareFactory.php
+++ b/src/MiddlewareFactory.php
@@ -76,11 +76,11 @@ class MiddlewareFactory
             return $this->pipeline(...$middleware);
         }
 
-        if (is_string($middleware)) {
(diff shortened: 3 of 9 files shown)


src/Application.php
```diff
@@ -104,7 +104,7 @@ public function pipe($middlewareOrPath, $middleware = null) : void
         $middleware = $middleware ?: $middlewareOrPath;
         $path = $middleware === $middlewareOrPath ? '/' : $middlewareOrPath;
 
-        $middleware = $path !== '/'
+        $middleware = ( is_string($path) && $path !== '/' )
```
webimpress: see #564

src/Handler/NotFoundHandler.php
```diff
@@ -59,15 +59,14 @@ public function __construct(
      * Creates and returns a 404 response.
      *
      * @param ServerRequestInterface $request Passed to internal handler
-     * @param RequestHandlerInterface $handler Ignored.
      */
     public function handle(ServerRequestInterface $request) : ResponseInterface
     {
-        if (! $this->renderer) {
+        if ($this->renderer === null) {
```
webimpress: why?

src/Handler/NotFoundHandler.php
```diff
@@ -59,15 +59,14 @@ public function __construct(
...
             return $this->generatePlainTextResponse($request);
         }
 
-        return $this->generateTemplatedResponse($request);
+        return $this->generateTemplatedResponse($this->renderer, $request);
```
webimpress: what's the point as it is property?

src/Middleware/ErrorResponseGenerator.php
```diff
@@ -62,14 +62,15 @@ public function __invoke(
     ) : ResponseInterface {
         $response = $response->withStatus(Utils::getStatusCode($e, $response));
 
-        if ($this->renderer) {
-            return $this->prepareTemplatedResponse($e, $request, $response);
+        if ($this->renderer !== null) {
+            return $this->prepareTemplatedResponse($this->renderer, $e, $request, $response);
```
webimpress: As above, it looks quite weird to pass class property to the private class method

src/MiddlewareFactory.php
```diff
@@ -76,11 +76,11 @@ public function prepare($middleware) : MiddlewareInterface
             return $this->pipeline(...$middleware);
         }
 
-        if (! is_string($middleware) || ! $middleware) {
-            throw Exception\InvalidMiddlewareException::forMiddleware($middleware);
+        if (is_string($middleware)) {
```
webimpress: wrong: https://3v4l.org/DIbno

src/Application.php
belgattitude: Sorry wasn't clear. for me. The TypeError looks good. I'll update

src/Handler/NotFoundHandler.php
belgattitude: Not a bug of course, the type is `null|TemplateRendererInterface`. This is just a proposal to include phpstan as q&a tool without adding too much exclusions (ignore patterns in phpstan.neon). 

The same idea applies to the next comments. It's just a proposal feel free to ignore the P/R. 

I'll update with phpstan, so you can see by yourself and decide if it  helps or not.

webimpress: Fine, check my last comment: https://github.com/zendframework/zend-expressive/pull/565#discussion_r170020405

The condition is replaced wrong, so not sure if the tool works correctly?

Ocramius: > The same idea applies to the next comments. It's just a proposal feel free to ignore the P/R.

Having a static analysis tool is very valuable, so I'm fairly sure this will go in :+1:

src/Handler/NotFoundHandler.php
Ocramius: Property is defined as nullable, whereas here it is not nullable (call-time)

src/Middleware/ErrorResponseGenerator.php
Ocramius: Same thing: can't rely on the property due to it being documented as nullable, so it is added as a call-time parameter once it is checked

src/MiddlewareFactory.php
belgattitude: Ooops, my mistake...

I can revert and add an exclusion in phpstan. Or simply test empty string in a way that It won't complain: 

```
// many options exists depending on where we want
// to catch the error (if empty the middleware cannot be found anyway)
if (is_string($middleware) && $middleware !== '') {        
            return $this->lazy($middleware);
}
throw Exception\InvalidMiddlewareException::forMiddleware($middleware);
```

Let me know

webimpress: tbh I don't like it, this is PHP not Java, nor any strongly typed programming language, and making changes like that are covering php flexibility. I know many people will disagree with me...

belgattitude:
@webimpress thanks for your comments... 

to give a little context I've included (temporarily) the phpstan config, see https://github.com/zendframework/zend-expressive/pull/565/commits/0feb7b566c6690a82b8ddda20bfb4984f223a0fc

You can have a look at the current ignored patterns. and run a check with `composer phpstan`.

Don't want to promote phpstan, but it helped me a point errors quickly while migrating from 2 to 3 alpha... 

Just a base for discussion.

PS: I'll have a look to https://github.com/zendframework/zend-expressive/pull/565#discussion_r170020405 later on.

src/MiddlewareFactory.php
belgattitude: I'm not purist neither, so I agree with you 100%. The idea behind this P/R is to propose phpstan as Q&A tool.  

So I tried to figure out where it would complain and tried to minimize ignored patterns. Not to make it Java like ;) 

The goal is more Q&A for people like me who would gladly contribute from times to times. PHPstan can be a good safety belt. 

I'll check tomorrow to see what's best to do and probably revert this one.

belgattitude: > The condition is replaced wrong, so not sure if the tool works correctly?

without the change phpstan 0.9 will complain about

```
------ --------------------------------------------------------------- 
  Line   src/MiddlewareFactory.php                                      
 ------ --------------------------------------------------------------- 
  79     Only booleans are allowed in a negated boolean, string given.  
 ------ --------------------------------------------------------------- 
```

`array|callable|MiddlewareInterface|RequestHandlerInterface` have been well discarded by previous tests (is_array, is_callable...). The remaining possible type is `string`... In `strict_mode` it will complain. We can choose of course

weierophinney:
@belgattitude I took the liberty of rebasing your work off of the current release-3.0.0, and updating the `MiddlewareFactory` to reflect our intentions; everything currently passes phpstan!

weierophinney:
@belgattitude I've also added phpstan checks to Travis under the PHP 7.2 locked environment.

weierophinney:
Thanks, @belgattitude! This is quite useful, and I learned a few things along the way.

PR #570, opened 2018-02-26
Incorrect class name in changelog

PowerKiKi:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -531,8 +531,8 @@ All notable changes to this project will be documented in this file, in reverse 
 - [#543](https://github.com/zendframework/zend-expressive/pull/543) removes the
   following methods from `Zend\Expressive\Application`:
 
-  - `pipeRoutingMiddleware()`: use `pipe(\Zend\Expressive\Router\PathBasedRoutingMiddleware::class)` instead.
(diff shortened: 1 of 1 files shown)


webimpress:
@PowerKiKi This is in release notes for expressive `3.0.0alpha3` and at this point all was good, check `zend-expressive-router:3.0.0alpha1`:
https://github.com/zendframework/zend-expressive-router/tree/3.0.0alpha1/src

PowerKiKi:
This is no longer correct in [alpha3](https://github.com/zendframework/zend-expressive-router/tree/3.0.0alpha3/src), exactly since https://github.com/zendframework/zend-expressive-router/commit/84a0f54e0c7567c35473ea83b0664014865a1df7#diff-25d902c24283ab8cfbac54dfa101ad31. I think doc should be kept up to date as much as possible.

webimpress:
@PowerKiKi still I think it's changelog and at time of this release all was fine, this is not a migration guide so we shouldn't change changelog as it was correct and became incorrect because one of dependencies changed.

weierophinney:
@PowerKiKi — @webimpress is correct. The changelog entry is correct _for the release in which it was made_.

When I prepare the final 3.0.0 changelog, I'll be merging all of the alpha entries, and, at that time, these will be corrected to the final versions.

PR #571, opened 2018-02-26
Make all implementation final

snapshotpl:
IMO we should mark final also factories and config provider. What are you think?

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -19,7 +19,7 @@ use Zend\Stratigility\MiddlewarePipeInterface;
 
 use function Zend\Stratigility\path;
 
-class Application implements MiddlewareInterface, RequestHandlerInterface
+final class Application implements MiddlewareInterface, RequestHandlerInterface
 {
     /**
      * @var MiddlewareFactory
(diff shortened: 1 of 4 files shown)


webimpress:
The problem is with testing - mocking. For example `MiddlewareContainer` implements only `ContainerInterface` but we have RTH on `get` method - `MiddlewareInterface`. If so we should have then separate interface - like `MiddlewareContainerInterface` with this method signature and everywhere use that typehint - and then in tests use mock from the interface....
Unless you know another solution?

snapshotpl:
If mocking is a problem that means problem with tests.

I need to fix this cases, however I need to be sure that change is welcomed.

weierophinney:
We're in feature freeze, @snapshotpl, and this change represents breakage. We can consider again for v4.

That said, each of the classes you marked final, with the exception of the `LazyLoadingMiddleware`, are ones I'd reasonably expect could be extended. The `Application` class, in particular, is one where users might extend it to provide additional functionality (more routing methods, etc.). The `MiddlewareContainer` might be extended to add support for double-pass middleware decoration. etc. etc. etc.

PR #572, opened 2018-02-26
Avoid object creation by class name

snapshotpl:
Why?
* it's error-prone,
* hard to debug,
* it add to much responsibility to one class

We should move responsibility of service creation to decorated container and user. I make this PR because I had a lot of real life problems with this magic.

### DIFF
--- a/src/MiddlewareContainer.php
+++ b/src/MiddlewareContainer.php
@@ -27,22 +27,16 @@ class MiddlewareContainer implements ContainerInterface
     }
 
     /**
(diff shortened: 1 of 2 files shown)


webimpress:
@snapshotpl Please have a look here:
https://github.com/zendframework/zend-expressive/pull/543#issuecomment-360724150
and next comments :)

weierophinney:
I appreciate the patch; however, we're on feature freeze, and this is a feature that helps migrations (as it's the previous behavior in v2). We can re-evaluate for v4, but for now, the feature stays.

PR #575, opened 2018-02-28
Versions documentation

weierophinney:
This patch versions the documentation.

V1 docs were imported from the v1.1.1 tag, and pushed under a v1 subdirectory of `docs/book/`. A sibling `v2/` subdirectory was also created, and the existing v2 docs were moved under that directory. The v1-specific migration documentation was removed from the v2 tree, leaving us with a single migration document for that tree.

v2 documentation was scrubbed of references to "version 1" and/or "since v2"; users can go to the v1 documentation for that information.

All existing pages were updated to render as redirects, using a combination of `<noscript>` tags and JS redirects.

The mkdocs configuration was updated:

- to rewrite all existing mappings to the v2 subdirectory
- to add links for the v1 docs
- to add hidden mappings for the existing redirect pages

I've tested builds locally, and all works as expected. Existing URIs redirect to the correct v2 or v1 variants (v1 if the doc only exists in that tree; v2 otherwise), and the navigation now uses v2 links.


docs/book/cookbook/autowiring-routes-and-pipelines.md
```diff
@@ -1,210 +1,8 @@
...
-- We recommend using delegator factories for the purpose of autowiring routes,
-  and, with caveats, pipeline middleware:
-    - The pipeline should be created exactly once, so calls to `pipe()` should
-      occur in exactly _one_ delegator factory.
-- Distributable packages should create a delegator factory for _routes only_,
-  but _should not_ register the delegator factory by default.
+<noscript><meta http-equiv="refresh" content="0; url=v2/cookbook/autowiring-routes-and-pipelines/"></noscript>
+<script>
+  document.addEventListener("DOMContentLoaded", function (event) {
+    var uri = new URL(window.location.href);
```
froschdesign: The problem still exists: it will not work in IE.

You can simplify the script:

```javascript
document.addEventListener('DOMContentLoaded', function (event) {
    window.location.pathname = 'v2/cookbook/autowiring-routes-and-pipelines/';
});
```

weierophinney: In which versions of IE will it not work?

Based on our metrics, IE users are a sliver of a percent of users; most are on (in decreasing order) Chrome, Firefox, Safari, or Edge. The majority of IE users are on 11. If the syntax will work on IE 11, the above will work.

Additionally, window.location.pathname is _read-only_. We would need to use `location.replace`, which expects a fully qualified URI... which the proposed solution accomplishes using the URL class.

froschdesign: > In which versions of IE will it not work?

All:

* https://caniuse.com/#feat=url
* https://developer.mozilla.org/en-US/docs/Web/API/URL#Browser_compatibility

froschdesign: > Additionally, window.location.pathname is read-only.

But works.

> We would need to use `location.replace`…

Which is correct for HTTP redirects.

```javascript
document.location.replace(
    document.location.protocol + '//' + document.location.host +  '/v2/cookbook/autowiring-routes-and-pipelines/'
);
```

Issue #474, opened 2017-03-26
LazyLoadingMiddleware suggestion


n3vrax:
Hi guys,

I was wondering if it wouldn't be good to have a getter in `LazyLoadingMiddleware` class to the the middleware service's name, as I'm often put in the situation of needing that, and I guess I'm not the only one that would need that, and I don't like the idea of using reflection just for that.

Situations when you need the middleware names are for example when writing configuration driven libraries that work on a per middleware basis(like globally configured authorization, content validation etc.)

Version ^1.0 had that in the RouteResult -> getMatchedMiddleware() even if lazy loaded
Or better, a new method for the RouteResult that would always return the middleware stack names.

I'm open to suggestions on achieving this, maybe I am missing something and there is already a method for it

Thanks!
Tiberiu

Ocramius:
Could you please make a usage example first? What's the use-case scenario?

n3vrax:
I needed this when developing a content validation library for example.
The setup is:
* i have routed middleware(for api endpoints)
* i need to configure a content validation middleware, after routing and before dispatching
* the configuration is per routed middleware, so in the configuration file I use the middleware class name as key, to associate an input filter to it(this is much like what apigility is doing)

```php
'dot_content_validation' => [
        'controllers' => [
            FooResource::class => [
                // the route param name used to identify a resource
                'identifier_name' => 'id',

                // the below flag is not used yet, reserved for future use
                'validate_uploaded_files' => false,

                // return api error when request data has unknown fields
                'allow_only_validated_fields' => true,

                // configure input filters to apply
                'validation' => [
                    'input_filter' => 'FooInputFilter',
                    // 'GET' => 'GetFooQueryParamsInputFilter'
                    // 'POST' => 'CreateFooInputFilter',
                    //...
                ],
            ],
        ],
    ],
```

Now, in the content validation middleware I need to get the input filter from the manager, as configured
For now, I solved it by using reflection like this
```php
protected function getMiddlewareName($middleware): string
    {
        if (is_string($middleware)) {
            return $middleware;
        }

        if ($middleware instanceof LazyLoadingMiddleware) {
            $class = new \ReflectionObject($middleware);
            $middlewareProperty = $class->getProperty('middlewareName');
            $middlewareProperty->setAccessible(true);
            return $middlewareProperty->getValue($middleware);
        }

        if (is_object($middleware)) {
            return get_class($middleware);
        }

        return '';
    }
```

[...]

PR #577, opened 2018-03-06
Made middlewareName somwhat accessible in LazyLoadingMiddleware

Eloar:

- [x] Are you creating a new feature?
  - [x] Why is the new feature needed? What purpose does it serve?

    I needed to access middleware for current rote in my StrategyInterface implementation. I have some information provided as annotations on action (it provides some context and is easy to be found). The problem is, middleware is set as LazyLoadingMiddleware in action. If I could have class name, I could access annotations.

  - [ ] How will users use the new feature?

    Anyware user will get `LazyLoadingMiddleware` instance he may get underlying middleware name to get it from container by himself.

  - [ ] Base your feature on the `develop` branch, and submit against that branch.
  - [x] Add only one feature per pull request; split multiple features over multiple pull requests
  - [ ] Add tests for the new feature.

    It is just simple getter so test would act just as cover.

  - [x] Add documentation for the new feature.

    method is documented

  - [ ] Add a `CHANGELOG.md` entry for the new feature.

### DIFF
--- a/src/Middleware/LazyLoadingMiddleware.php
+++ b/src/Middleware/LazyLoadingMiddleware.php
@@ -79,4 +79,13 @@ class LazyLoadingMiddleware implements ServerMiddlewareInterface
             return $delegate->process($request);
         });
     }
+
+    /**
+     * Getter for name of underlying Middleware
+     * @return string name of underlying middleware
+     */
+    public function getMiddlewareName()
+    {
+        return $this->middlewareName;
+    }
 }


Ocramius:
> I needed to access middleware for current rote in my StrategyInterface implementation

This seems like the wrong prerequisite. This happens every time reading data from services is attempted too late (after the service is instantiated), and should instead be performed in the bootstrap phase of the app.

> It is just simple getter so test would act just as cover.

Tests have to be written for every change in order to verify that the change is valid, as well as making sure it stays valid: no exclusions, no matter how trivial the code is.

IMO this isn't needed, and it is a fix for a different issue that comes from the configuration: patch seems to solve a symptom rather than the root cause.

weierophinney:
Agreed with @Ocramius. The issue goes deeper: what if you create a pipeline of middleware, either when piping or routing? At that point, you do not even have a `LazyLoadingMiddleware` instance, but rather a generic `Zend\Stratigility\MiddlewarePipe`.

What exact problem are you trying to solve? You mention a strategy pattern, but does that mean:

- you're trying to add strategies to a given middleware?
- you're determining what strategy to use based on the middleware?
- something else?

If the first point is what you're accomplishing, use a [delegator factory](https://docs.zendframework.com/zend-expressive/v2/features/container/delegator-factories/) to inject strategies at the time that the middleware is being pulled from the container.

If the second is what you're attempting to accomplish, I'd build the strategies based on the _route matched_ instead. That way the rules will work even if you end up decorating your middleware, providing a middleware pipeline, or providing a substitution.

The chances that you'll have access to the specific middleware instance are pretty slim. Typically you can solve most problems using one of the above approaches.

Regardless, this is not a feature we wish to implement at this time.

Eloar:
@weierophinney I'm determining how to parse body based on annotation on action. I moved my `BodyParamsMiddleware` implementation to place in pipe where I'm sure to have proper route with middleware, but it is `LazyLoadingMiddleware` so I may only process it (which would be wrong).

This is code for solving not yet closed issue #474 which is almost year old. I would appreciate suggestions to improve this piece of code to be marege ready. My own code might be usefull but it is not the only case when getter on name in `LazyLoadingMiddleware` would be used.

@Ocramius ok, I will try to provide good unit test for this simple getter.

weierophinney:
@Eloar There are multiple solutions presented there and here that allow you to accomplish what both you and the originator of that issue want to do, without adding the getter method.

You have ignored the point I made: you may or may not have a `LazyLoadingMiddleware` instance here. You could have:

- The middleware instance itself.
- A middleware pipeline.
- LazyLoadingMiddleware.
- A CallableInteropMiddlewareWrapper.
- A CallableMiddlewareWrapper.
- In upcoming versions, a PathMiddlewareDecorator, CallableMiddlewareDecorator, or DoublePassMiddlewareDecorator.

There are other approaches you can also take. One thing we've done with the upcoming v3 and v2.2 releases is adding _utility functions_ for generating middleware wrappers. These make it possible for you to create context-specific decorators directly where you need to use them.

As an example, you know already the action class when routing to it. For your custom BodyParamsMiddleware, have the service return a _factory_ that you will instead invoke _with a runtime value_ in order to create a discrete instance. Then create a callable wrapping the functionality:

```php
// in config/routes.php:
$container = $app->getContainer();
$bodyParamsMiddleware = function (string $forClass) use ($container) {
    $factory = $container->get(BodyParamsMiddleware::class);
    return $factory($forClass);
};

$app->post('/api/something', [
    $bodyParamsMiddleware(YourAction::class);
    YourAction::class,
]);
```

This approach allows you to define services that themselves have dependencies contextually.

My main point is: the middleware type _should not matter_. Composition within the pipeline and request artifacts present at call time _do_. Use those to accomplish your goals.

Eloar:
@weierophinney  ok, I might try to create some kind of wrapping middleware for each of actions where body parsing should be changed. I don't yet see how I would have this wrapped middleware served to `BodyParamsMiddleware`. About that information that there might be no `LazyLoadingMiddleware`, in my case I know there is and will be. I will try to use it different way... If #474 issue is product of misunderstanding and misuse of framework should be closed.

Or maybe there is other way to get all registered routes with its exact actions as those hold some code, aspect and documentation.

weierophinney:
> Or maybe there is other way to get all registered routes with its exact actions as those hold some code, aspect and documentation?

No, there's not - because of the reasons outlined above (auto-decoration of known types, creation of middleware pipelines, etc.). The point of using a _standard interface_ is to make things _interchangeable_. As such, you need to rely on either declarative approaches like I outlined in my previous comment, or runtime information (such as matched route name).

Eloar:
@weierophinney so isn't `Zend\Expressive\Application` a _standard interface_? Ok, with MiddlewarePipe I am able to solve problem with body parsing, but other problem is generation of documentation where I tried to generate list of routes with description from actions. I tried to do it in one of actions where application object were injected. There is just no way to read routes the way those were declared. They are always changed, right?

weierophinney:
@Eloar At this point, I'm going to ask you to direct your question to the forums, where we may be able to get more eyes on it, and where you can provide more detail:

- https://discourse.zendframework.com/c/questions/expressive/

PR #579, opened 2018-03-06
Updated to the latest dependencies and fixed integration tests

webimpress:
These need update after releasing fixes in `Implicit*Middlewares`

See: https://github.com/zendframework/zend-expressive-router/pull/58

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,22 +28,22 @@
         "psr/http-message": "^1.0.1",
         "psr/http-server-middleware": "^1.0",
         "zendframework/zend-diactoros": "^1.3.10",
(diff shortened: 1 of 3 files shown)


webimpress:
Two tests are failing:
> ZendTest\Expressive\Router\IntegrationTest::testAllowedMethodsWhenNoHttpMethodsSet with data set "fast-route-options" ('Zend\Expressive\Router\FastRouteRouter', 'OPTIONS')
Error: Undefined class constant 'METHOD_ANY'

- because of issue in zend-expressive-fastroute which is fixed in https://github.com/zendframework/zend-expressive-fastroute/pull/50

PR #580, opened 2018-03-07
zend-diactoros is an optional dependency

webimpress:
Everywhere where we are using diactoros we check first for class existence, so it should be an optional dependency.

Added conflict with version prior to 1.7.1 - see [release notes for 1.7.1](https://github.com/zendframework/zend-diactoros/releases/tag/1.7.1)

Added to suggest section in composer.json.

**Build is failing because of other issues on current release-3.0.0 branch** - see #579


weierophinney:
Merged with 75af35c.

PR #581, opened 2018-03-08
Backports and deprecations for 2.2 release

weierophinney:
This is a work-in-progress branch implementing the Expressive 2.2 roadmap, which is primarily around deprecations and backports, and detailed here:

- https://discourse.zendframework.com/t/roadmap-expressive-2-2/504

TODO
----

- [x] Update to zend-stratigility 2.2
- [x] Update to zend-expressive-router 2.4
- [x] Update the `MarshalMiddlewareTrait`:
  - [x] When the routing/dispatch middleware constants are encountered:
    - [x] Raise a deprecation notice, detailing proper usage.
    - [x] Create and use instances of the zend-expressive-router middleware, pulling them from the container.
  - [x] When double-pass middleware is encountered, raise a deprecation notice, detailing how to modify your code to use the appropriate zend-stratigility decorator, and then do so internally.
  - [x] Mark the trait as deprecated.
- [x] Update `Application`
  - [x] Mark `pipeRoutingMiddleware()` deprecated, and have it trigger a deprecation notice.
  - [x] Mark `pipeDispatchMiddleware()` deprecated, and have it trigger a deprecation notice.
  - [x] Mark `getContainer()` deprecated
  - [x] Mark `getDefaultDelegate()` deprecated
  - [x] Mark `getEmitter()` deprecated
- [x] Create `ResponseFactoryFactory`, which should create a PHP closure around creation of a response instance, and mapped to the `ResponseInterface` service. Existing factories that require an actual instance should pull this and invoke it immediately. This can be pulled from the release-3.0.0 branch.
- [x] Create a `StreamFactoryFactory`, which should create a PHP closure around creation of a PSR-7 stream instance; this can be pulled from the release-3.0.0 branch.
- [x] Create `Zend\Expressive\Handler\NotFoundHandler`, based on the current `NotFoundDelegate`. Map the `DefaultDelegate` service to this instead.
- [x] Create a `ConfigProvider` with default service mappings as currently present in the skeleton; alias deprecated functionality to new functionality as it is created, and add any new services as needed.
- [x] Create `ApplicationConfigInejctionDelegator`, moving the instance methods from the existing trait into static methods of the class that require access to an `$application` instance.
- [x] Mark deprecated classes and traits
  - [x] `AppFactory`. Emit a deprecation notice at the file level.
  - [x] `Container\NotFoundDelegatorFactory`
  - [x] `Delegate\NotFoundDelegate`
  - [x] `Emitter\EmitterStack`
  - [x] `IsCallableInteropMiddlewareTrait`
  - [x] `Middleware\DispatchMiddleware` (extend from zend-expressive-router implementation, and have constructor raise a deprecation notice. Its factory should extend the canonical zend-expressive-router factory, and also be marked deprecated.)
  - [x] `Middleware\ImplicitHeadMiddleware` (extend from zend-expressive-router implementation, and have constructor raise a deprecation notice. Its factory should extend the canonical zend-expressive-router factory, and also be marked deprecated.)
  - [x] `Middleware\ImplicitOptionsMiddleware` (extend from zend-expressive-router implementation, and have constructor raise a deprecation notice. Its factory should extend the canonical zend-expressive-router factory, and also be marked deprecated.)
  - [x] `Middleware\NotFoundHandler`
  - [x] `Middleware\RouteMiddleware` (extend from zend-expressive-router implementation, and have constructor raise a deprecation notice. Its factory should extend the canonical zend-expressive-router factory, and also be marked deprecated.)


weierophinney:
Ready to review. Coverage decreases are expected:

- Classes that now extend classes in zend-expressive-router
- AppFactory no longer gets coverage (due to the fact that xdebug halts as soon as trigger_error occurs in that class file)

src/Container/ApplicationConfigInjectionDelegator.php
```diff
@@ -0,0 +1,373 @@
...
+     *
+     * @return callable
+     * @throws InvalidArgumentException
+     */
+    private static function createCollectionMapper()
+    {
+        $appMiddleware = [
+            Application::ROUTING_MIDDLEWARE,
+            Application::DISPATCH_MIDDLEWARE,
+        ];
```
weierophinney: This differs from release-3.0.0, but is necessary for BC purposes in version 2.

src/Container/ApplicationConfigInjectionDelegator.php
```diff
@@ -0,0 +1,373 @@
...
+    /**
+     * Pipe routing middleware into the application.
+     *
+     * Attempts to pull both the router and response interface from the application
+     * container, falling back to reflection of the application instance
+     * when unable to find either one.
+     *
+     * @return void
+     */
+    private static function pipeRoutingMiddleware(Application $application)
```
weierophinney: This method is necessary in v2 to allow `pipeRoutingMiddleware()` and usage of the `ROUTING_MIDDLEWARE` constant to work; it is not present in v3.

src/Container/ApplicationConfigInjectionDelegator.php
```diff
@@ -0,0 +1,373 @@
...
+     * Pipe dispatch middleware into the application.
+     *
+     * If the DispatchMiddleware is present in the application's container, it
+     * pipes that service. Otherwise, instantiates it directly and pipes it.
+     *
+     * @return void
+     * @throws MissingDependencyException if the RouterInterface service is not
+     *     found when the RouteMiddleware is not present.
+     */
+    private static function pipeDispatchMiddleware(Application $application)
```
weierophinney: This method is necessary in v2 to allow `pipeDispatchMiddleware()` and usage of the `DISPATCH_MIDDLEWARE` constant to work; it is not present in v3.

src/Container/ApplicationConfigInjectionDelegator.php
```diff
@@ -0,0 +1,373 @@
...
+
+    /**
+     * Get the router.
+     *
+     * If not available in the container, uses reflection to pull it from the
+     * application.
+     *
+     * @return RouterInterface
+     */
+    private static function getRouter(Application $application, ContainerInterface $container)
```
weierophinney: This method is only necessary in v2, as the router may be injected only in the application, and not the container (most likely only true when using `AppFactory`, however).

src/Container/ApplicationConfigInjectionDelegator.php
```diff
@@ -0,0 +1,373 @@
...
+     *
+     * If not available in the container, uses reflection to pull it from the
+     * application.
+     *
+     * If in the container, fetches it. If the value is callable, uses it as
+     * a factory to generate and return the response.
+     *
+     * @return ResponseInterface
+     */
+    private static function getResponsePrototype(Application $application, ContainerInterface $container)
```
weierophinney: This method is only necessary in v2, as the response prototype may be injected only in the application, and not the container.

src/Container/ApplicationFactory.php
```diff
@@ -91,7 +91,7 @@ public function __invoke(ContainerInterface $container)
      */
     private function injectRoutesAndPipeline(Application $app, array $config)
     {
-        $app->injectRoutesFromConfig($config);
-        $app->injectPipelineFromConfig($config);
+        ApplicationConfigInjectionDelegator::injectRoutesFromConfig($app, $config);
+        ApplicationConfigInjectionDelegator::injectPipelineFromConfig($app, $config);
```
weierophinney: I did this to prevent triggering the deprecation notices in the trait when using this functionality.

src/Container/ResponseFactoryFactory.php
```diff
@@ -0,0 +1,29 @@
...
+
+use Psr\Container\ContainerInterface;
+use Psr\Http\Message\ResponseInterface;
+use Zend\Diactoros\Response;
+
+/**
+ * Produces a callable capable of producing a response prototype for use with
+ * services that need to produce a response.
+ */
+class ResponseFactoryFactory
```
weierophinney: This class is new in v2, and we did not previously define a `ResponseInterface` service. I've ensured that all factories in zend-expressive and zend-expressive-router that might possibly inject a response instance will pull the service, and, if it's a factory, invoke it to get the response prototype to inject. I will do so with other components as well.

src/Container/StreamFactoryFactory.php
```diff
@@ -0,0 +1,29 @@
...
+
+use Psr\Container\ContainerInterface;
+use Psr\Http\Message\StreamInterface;
+use Zend\Diactoros\Stream;
+
+/**
+ * Produces a callable capable of producing an empty stream for use with
+ * services that need to produce a stream for use with a request or a response.
+ */
+class StreamFactoryFactory
```
weierophinney: This factory is new to v2, and no previous factories used a stream prototype, making it BC.

src/Middleware/DispatchMiddleware.php
```diff
@@ -28,65 +21,10 @@
...
-        $middleware = $routeResult->getMatchedMiddleware();
-
-        if (! $middleware instanceof ServerMiddlewareInterface) {
-            $middleware = $this->prepareMiddleware(
-                $middleware,
-                $this->router,
-                $this->responsePrototype,
-                $this->container
-            );
-        }
```
weierophinney: This logic was only used previously if using the `ApplicationConfigInjectionTrait`, as it would not first marshal the middleware passed to it before piping our routing it. Since the `ApplicationConfigInjectionDelegator` _does_ do that at this point, this logic would never be hit, allowing it to be removed in the zend-expressive-router variant.

src/Container/ApplicationConfigInjectionDelegator.php
```diff
@@ -0,0 +1,373 @@
...
+    {
+        if (empty($config['middleware_pipeline'])) {
+            if (! isset($config['routes']) || ! is_array($config['routes'])) {
+                return;
+            }
+
+            self::pipeRoutingMiddleware($application);
+            self::pipeDispatchMiddleware($application);
+            return;
+        }
```
weierophinney: I'm thinking the above clause should likely be removed. It creates a potentially dangerous situation if the user is creating a programmatic pipeline, but using this delegator, as the above will pipe the routing and dispatch middleware _before_ the user has had a chance to setup the pipeline.

I'm thinking the above should only trigger if the "programmatic pipeline" flag is _not_ enabled.

src/Container/ApplicationFactory.php
```diff
@@ -85,13 +90,87 @@ public function __invoke(ContainerInterface $container)
     /**
      * Injects routes and the middleware pipeline into the application.
      *
-     * @param Application $app
-     * @param array $config
      * @return void
      */
-    private function injectRoutesAndPipeline(Application $app, array $config)
-    {
+    private function injectRoutesAndPipeline(
```
weierophinney: I've added this method, and the other new methods in this factory, as this is the only place where _implicitly_ injecting the routing and dispatch middleware makes sense. Doing it from the `Application::pipeRoutingMiddleware()` (via the `ApplicationConfigInjectionTrait`) or the `ApplicationConfigInjectionDelegator` does not make sense, as these will typically execute _before_ the pipeline is fully built, leaving the pipeline in an incorrect state if any middleware are added via `config/pipeline.php` later.

In this particular factory, however, the methods only execute if the user has not opted into a programmatic pipeline; i.e., if they have specifically chosen to use configuration-driven pipelines, or not updated their pre-2.0 app to use programmatic pipelines. This is the only place implicit registration makes sense.

PR #582, opened 2018-03-09
Fixed redirection links in versioned documentation

webimpress:
For example:
https://docs.zendframework.com/zend-expressive/features/router/piping/
redirects to:
https://docs.zendframework.com/v2/features/router/piping/
instead of:
https://docs.zendframework.com/zend-expressive/v2/features/router/piping/

Also removed index.md as we have there index.html.

**There is another issue:**
- https://docs.zendframework.com/zend-expressive/v1/ - works, loads index.md from `v1/`
- https://docs.zendframework.com/zend-expressive/v2/ - shows 404, why? There is index.md in `v2/`.

Maybe this is not an issue, because we don't have this link anywhere, but I think we should see there something, especially the file is there.

/cc @weierophinney @froschdesign


weierophinney:
> shows 404, why? There is index.md in v2/.
> 
> Maybe this is not an issue, because we don't have this link anywhere, but I think we should see there something, especially the file is there.

This is because there's not an entry for it in the mkdocs.yml. I'll add it when merging.

weierophinney:
Also: had to re-instate `index.md` in the root tree; this is _required_ in order to build, even though we swap it with the `index.html`.

PR #592, opened 2018-03-16
Fix DispatchMiddlewareFactory duplicate

xtreamwayz:

In version 2.2.0 The [`DispatchMiddleware::class`](https://github.com/zendframework/zend-expressive/blob/2.2.0/src/ConfigProvider.php#L46-L48) is an invokable, but it also has a factory. The factory is used in [zend-expressive-router](https://github.com/zendframework/zend-expressive-router/blob/2.4.1/src/ConfigProvider.php#L30) and is causing a conflict. The DispatchMiddlewareFactory should be used instead.

- [x] Are you fixing a bug?
  - [x] Detail how the bug is invoked currently.
  - [x] Detail the original, incorrect behavior.
  - [x] Detail the new, expected behavior.
  - [x] Base your feature on the `master` branch, and submit against that branch.
  - [x] ~~Add a regression test that demonstrates the bug, and proves the fix.~~
  - [x] Add a `CHANGELOG.md` entry for the fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,32 @@
 
 All notable changes to this project will be documented in this file, in reverse chronological order by release.
 
+## 2.2.1 - TBD
+
+### Added
+
+- Nothing.
+
+### Changed
+
+- Nothing.
+
+### Deprecated
+
+- Nothing.
+
+### Removed
+
+- Nothing.
+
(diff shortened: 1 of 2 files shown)


xtreamwayz:
Only merge it into 2.2.0?

PR #593, opened 2018-03-16
Docs update Psr\ instead of Interop\

gsteel:
Reading the new docs - middleware-types.md for v3 references deprecated Interop namespace instead of Psr


weierophinney:
Thanks, @gsteel!

PR #596, opened 2018-03-17
allow config-route string key as name

pine3ree:
Hello..just a thought...

Preamble: I still prefer configuration driven pipelines and routes for non trivial projects. 

The current implementation searches for a 'name' key inside each defined route spec, but since the name must be unique we could use the routes array keys to define the name of each route. 
I believe this is also a little more readable: 1 line less per route spec and, at least for me, it's easier to search among array-keys.
 
Anyway, this PR does not impose a new  "unique key as name" spec... it just allows it by using a string key AND omitting the name inside the spec. If a 'name' is still defined inside the spec that would be the one used, so there is no backward incompatibility.

kind regards

### DIFF
--- a/docs/book/v3/cookbook/autowiring-routes-and-pipelines.md
+++ b/docs/book/v3/cookbook/autowiring-routes-and-pipelines.md
@@ -40,6 +40,15 @@ return [
                 'the'   => 'underlying router',
             ],
         ],
+        'another.route.name' => [
+            'path' => '/another/path/to/match',
+            'middleware' => 'Middleware service or pipeline',
+            'allowed_methods' => ['GET', 'POST'],
+            'options' => [
+                'more'    => 'router',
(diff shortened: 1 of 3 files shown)


webimpress:
@pine3ree I like the idea and I think we do the same in some other ZF component (hm... zend-form for defining forms and field names? or maybe inputfilter?).

I'd suggest only adding some test to cover these changes 👍

pine3ree:
@webimpress Hello, Yes, of course. I just wanted to make sure this could be interesting for other developers first so I decided to wait for a reviewer response/opinion.  btw a bit of documentation would be needed as well. Another small detail: in the pr I used a neutral `$key => $spec` variable for the array key, but in my opinion a more significative nomenclature would be `$name => $spec`. 
kind regards

weierophinney:
Thanks, @pine3ree!

PR #597, opened 2018-03-19
Update MiddlewareFactory.php

wesperinteractive:
Re-use the `handler()` method instead of directly instantiating a `RequestHandlerMiddleware` instance within the `prepare()` method.

### DIFF
--- a/src/MiddlewareFactory.php
+++ b/src/MiddlewareFactory.php
@@ -71,7 +71,7 @@ class MiddlewareFactory
         }
 
         if ($middleware instanceof RequestHandlerInterface) {
-            return new RequestHandlerMiddleware($middleware);
+            return $this->handler($middleware);
         }
 
         if (is_callable($middleware)) {


Ocramius:
Requires test additions and better PR/Commit description

xtreamwayz:
Not sure if this needs additional tests. All it does is using the internal method to return a new instance of `RequestHandlerMiddleware` instead of returning `RequestHandlerMiddleware` directly.

xtreamwayz:
Also the tests are already in place:

https://github.com/zendframework/zend-expressive/blob/a7bd1e9ea02778291718503775071fb81a4513bb/test/MiddlewareFactoryTest.php#L203-L217

Ocramius:
Gotcha: not having any context on the nature of the patch besides the diff, I asked for a test, assuming a functional change.

weierophinney:
Thanks, @wesperinteractive

Issue #588, opened 2018-03-15
Missing Zend Expressive 3 dependencies config format


oshdev:
As per suggestion #372 it seems that Zend Expressive uses the following format: https://github.com/gabbydgab/zend-expressive-config-provider/blob/develop/src/ConfigProvider.php#L29
Can we please have those documented or even better - if possible - create an interface(s) so we could create adapters for various container implementations with ease and perhaps include more of the popular ones to the https://github.com/zendframework/zend-expressive-skeleton installer.
For more information see https://discourse.zendframework.com/t/zend-expressive-3-dependencies-config-format/531

weierophinney:
@oshdev I can certainly document what configuration we expect to work across all implementations. However, _how the implementations consume that configuration_ is really up to each.

We've started splitting that task into separate packages; you can see those here:

- https://github.com/zendframework/zend-auradi-config
- https://github.com/zendframework/zend-pimple-config

Ideally, we should expose these as composer plugins or to a composer plugin in order to copy artifacts into the project (e.g., to copy in the `config/container.php` file); we'll likely do that in the future.

In the meantime, I'll add a document detailing what config features implementors need to be compatible with.

PR #600, opened 2018-03-19
Provide a specification for container configuration

weierophinney:
Per #588 and https://discourse.zendframework.com/t/zend-expressive-3-dependencies-config-format/531, this patch provides a document with the specification for container configuration that implementations MUST support in order to work with Expressive applications.

### DIFF
--- a/docs/book/v3/features/container/config.md
+++ b/docs/book/v3/features/container/config.md
@@ -1,0 +1,213 @@
+# Container configuration
+
+> This chapter is primarily written for container providers, so that they know
+> what configuration features must be compatible, and what compatibility
+> ultimately means within the project.
+
+[PSR-11](https://www.php-fig.org/psr/psr-11/) defines an interface for
+dependency injection containers, and that interface is geared towards
+_consumption_ of the container &mdash; not _population_ of it.
+
+Expressive _consumes_ a PSR-11 container, but also provides _configuration_ for
+a container: it defines what services it needs, and how to create them.
+
+As such, any container consumed by Expressive must also understand its
+configuration format, and deliver consistent understanding of that format when
+providing services based on it.
+
+This document describes the configuration format, and details expectations for
+implementations.
+
+## The format
+
+Container configuration is provided within the `dependencies` key of
+configuration. That key is structured as follows:
+
+```
+return [
+    'dependencies' => [
+        'services' => [
+            // name => instance pairs
+        ],
+        'aliases' => [
+            // alias => target pairs
+        ],
+        'factories' => [
+            // service => factory pairs
+        ],
+        'invokables' => [
+            // service => instantiable class pairs
+        ],
+        'delegators' => [
+            // service => array of delegator factories
+        ],
+    ],
+];
+```
+
+## Services
+
+_Services_ are actual instances you want to retrieve later from the container.
+These are generally provided at initial creation; the `config` service is
+populated in this way.
+
+When retrieving a service mapped in this way, you will always receive the
+initial instance.
+
+## Aliases
+
+_Aliases_ map a service _alias_ to another service, and are provided as
+key/value pairs. As an example:
+
+```php
+'aliases' => [
+    'Zend\Expressive\Delegate\DefaultDelegate' => \Zend\Expressive\Handler\NotFoundHandler::class,
+],
+```
+
+In this case, if the service named "Zend\\Expressive\\Delegate\\DefaultDelegate"
+is requested, the container should resolve that to the service
+`Zend\Expressive\Handler\NotFoundHandler` and return that instead.
+
+When returning an aliased service, the container MUST return the same instance
+as if the target service were retrieved.
+
+Aliases may reference other aliases, as well. In such cases, the above rules
+apply to the final resolved service, and not any intermediary aliases.
+
+## Factories
+
+_Factories_ map a service name to the factory capable of producing the instance.
+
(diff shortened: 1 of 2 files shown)


docs/book/v3/features/container/config.md
```diff
@@ -0,0 +1,213 @@
...
+that defines `__invoke()`. Generally, this latter convention is used, as class
+names are serializable, while closures are not.
+
+Factories are guaranteed to receive the PSR-11 container as an argument,
+allowing you to pull other services from the container as necessary to fulfill
+dependencies of the class being created and returned. Additionally, containers
+SHOULD pass the service name requested as the second argument; factories can
+determine whether that argument is necessary.
+
+A typical container will generally ignore the second argument:
```
webimpress: 'container' or 'factory'?

docs/book/v3/features/container/config.md
```diff
@@ -0,0 +1,213 @@
...
+
+```php
+use Psr\Container\ContainerInterface;
+use Zend\Expressive\Template\TemplateRendererInterface;
+
+function (ContainerInterface $container)
```
webimpress: Hm... Wouldn't be better to have here invokable class example?

docs/book/v3/features/container/config.md
```diff
@@ -0,0 +1,213 @@
...
+
+In this case, if the service named "Zend\\Expressive\\Delegate\\DefaultDelegate"
+is requested, the container should resolve that to the service
+`Zend\Expressive\Handler\NotFoundHandler` and return that instead.
+
+When returning an aliased service, the container MUST return the same instance
+as if the target service were retrieved.
+
+Aliases may reference other aliases, as well. In such cases, the above rules
+apply to the final resolved service, and not any intermediary aliases.
```
webimpress: I'm updating zend-container-test library and still for me it's not clear for what I can define aliases. I mean if it is possible to define here alias which reference to:
- service
- invokable
- factory
- delegator
- alias
?

docs/book/v3/features/container/config.md
```diff
@@ -0,0 +1,213 @@
...
+
+## Factories
+
+_Factories_ map a service name to the factory capable of producing the instance.
+
+A _factory_ is any PHP callable capable of producing the instance. They may also
+be the _class name_ of a directly instantiable class (no constructor arguments)
+that defines `__invoke()`. Generally, this latter convention is used, as class
+names are serializable, while closures are not.
```
webimpress: So.. does it mean we should support:
- invokable class: `['service' => Factory::class]` (when Factory has `__invoke` method)
- invokable array: `['service' => [FactoryStatic::class, 'create']]` (class is not invokable, `create` method is static),
- invokable string: `['service' => FactoryStatic::class . '::create']` (class is not invokable, `create` method is static), 
- callback: `['service' => function() { return new Service(); }]`

Asking because here we have some inconsistencies between our supported containers, and it needs clarification for our tests, see: https://github.com/webimpress/zend-container-test/blob/master/src/FactoryTestTrait.php

docs/book/v3/features/container/config.md
```diff
@@ -0,0 +1,213 @@
...
+This document describes the configuration format, and details expectations for
+implementations.
+
+## The format
+
+Container configuration is provided within the `dependencies` key of
+configuration. That key is structured as follows:
+
+```
+return [
```
oshdev: I think it'd be nice to provide samples as well, for instance:
```php
return [
    'dependencies' => [
        'services' => [
            // name => instance pairs
            FooService::class => new FooService(),
        ],
        'aliases' => [
            // alias => target pairs
            FooInterface::class => FooImplementation::class,
        ],
        'factories' => [
            // service => factory pairs
            FooImplementation::class => FooImplementationFactory::class,
        ],
        'invokables' => [
            // service => instantiable class pairs
            InvokableFoo::class => InvokableFoo::class,
        ],
        'delegators' => [
            // service => array of delegator factories
            FooInterface::class => [
                FooDelegator::class,
                FooDecorator::class,
            ],
        ],
    ],
];
```
Alternatively, maybe we could add an example for each of the configuration types below with more detail and alternatives.

docs/book/v3/features/container/config.md
```diff
@@ -0,0 +1,213 @@
...
+
+Configuration for delegator factories is using the "delegators" sub-key of the
+"dependencies" configuration. Each entry is a service name pointing to an
+_array_ of delegator factories. Delegator factories are called in the order they
+appear in configuration; when you call the `$callback` argument, it effectively
+calls the previous delegator to get the instance to work on; the first
+delegator's `$callback` argument invokes whatever functionality is required to
+create the initial instance (be it as a [service](#services),
+[invokable](#invokables), or [factory](#factories), after [alias](#aliases)
+resolution).
```
webimpress: Not sure if I understand it correctly, please see tests:
https://github.com/webimpress/zend-container-test/blob/master/src/DelegatorTestTrait.php

weierophinney: Your tests are (mostly) correct. (One thing to note: you have a number of tests supposedly testing that delegators are running when retrieving aliased services, but, in fact, you're calling `get()` on the non-aliased versions.)

When you call `get()` for a service that has delegator factories defined, you will receive the value returned by the last delegator in the list. The first delegator's `$callback()` is essentially the result of calling `get()` as if there were no delegators in place.

I'll see if I can reword this paragraph to make it more clear.

PR #603, opened 2018-03-21
Correct ZendViewRendererFactory feature description

shadyan:
See here:
https://docs.zendframework.com/zend-expressive/v3/features/container/factories/#zendviewrendererfactory

Says "no additional services are required", but then lists a dependency.

I made a separate pull request for this since this case might be a bit more complex. My understanding is that the actual dependency is the `UrlHelper`.
The `ZendViewRendererFactory` comments added some more confusion:
> Requires the Zend\Expressive\Router\RouterInterface service (for creating the UrlHelper instance).

https://github.com/zendframework/zend-expressive-zendviewrenderer/blob/master/src/ZendViewRendererFactory.php

In any case, I think this small change should suffice to relieve most of the confusion.

  - [x] Is it a typographical and/or grammatical fix?
   - [ ] Is it new documentation?


weierophinney:
Thanks, @shadyan! I reviewed the actual requirements, and made a few other changes on update to reflect how the factory works. Thanks for prodding me via this patch!

PR #619, opened 2018-06-04
Add middlewareName getter to LazyLoadingMiddleware

robopuff:
In some cases it is necessary to get a middleware name that is corresponding to the route and since that was possible to do in previous versions of Expressive, it would be nice to keep that functionality.
We are currently using input filters written in config files and they are maintained by middleware class name and currently only possible way of doing that is by overriding closure.

### DIFF
--- a/src/Middleware/LazyLoadingMiddleware.php
+++ b/src/Middleware/LazyLoadingMiddleware.php
@@ -45,7 +45,7 @@ class LazyLoadingMiddleware implements MiddlewareInterface
         $middleware = $this->container->get($this->middlewareName);
         return $middleware->process($request, $handler);
     }
-    
+
     public function getMiddlewareName() : string
     {
         return $this->middlewareName;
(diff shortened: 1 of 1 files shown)


weierophinney:
I don't think this is going to do what you think it will.

What happens if:

- a delegator wraps the middleware in a `MiddlewarePipe`?
- a delegator _replaces_ the middleware entirely?

Additionally, where exactly would you have access to the lazy-loading middleware? Generally speaking, the only classes that have direct access to it are the router, and potentially the dispatch middleware. You _might_ have access to it from the `RouteResult`, _but only if neither of the previous two scenarios occur_.

I honestly think you would be better served by using the route _name_ matched, and not the middleware; alternately, use a construct like this:

```php
$app->post('/api/some/endpoint', [
    inputValidatorMiddlewareFactory(Endpoint::class),
    Endpoint::class,
], 'endpoint');
```

(Where the `inputValidatorMiddlewareFactory` likely has access to the input filter plugin manager and/or the `MiddlewareFactory`.)

The above approach ensures the input filter is created and consumed regardless of what substitutions occur for `Endpoint`.

robopuff:
```php
        if ($middleware instanceof LazyLoadingMiddleware) {
            $middlewareName = (function () {
                return $this->middlewareName;
            })->call($middleware);
        }
```

Gives me precisely information that I need but it's suuper ugly hack.

PR #620, opened 2018-06-04
Syntax fixes

JustSteveKing:
A quick fix to the documentation, there were some rogue quotation marks which would cause things to crash and burn. Just doing my bit .... 

- [ X ] Is this related to documentation?
 Syntax fixes, there was 2 unneeded quotation marks .... doing my bit


weierophinney:
Thanks, @JustSteveKing.

PR #625, opened 2018-07-12
Documentation: Fix configuration key name

adrium:
There is a wrong key name printed in the documentation. This PR fixes this.


weierophinney:
Thanks, @adrium!

Issue #626, opened 2018-07-21
It was debug difficult of an error occurred by after handle pipeline


Moln:
### Code to reproduce the issue

It was difficult for us to debug error/exception, when an error/exception occurred by after handle pipeline.

These are situations that may occur.

1. Output has been emitted previously. 
    - When I have a config file and there is no `declare(strict_types=1)` and exists UTF8+BOM.
    - Echo something in middleware or handler.
    - etc..
2. Throw an errors/exceptions in `ErrorHandler`'s listeners.
    - When I attach a logger listener, and logging error.
    - etc..

There will output an empty response (status 200). Then, we don't know what happened.

### Expected results

Output messages by Whoops.

### Actual results

Response OK (status code 200). And no error/exception outputed.

### Suggestions

Remove `$whoops->writeToOutput(false)` in `WhoopsFactory`.

https://github.com/zendframework/zend-expressive/blob/491da1e365d54e467026641b5984c0dd181ebe90/src/Container/WhoopsFactory.php#L56

And add disable/enable send on both sides of get `handleException` result.

https://github.com/zendframework/zend-expressive/blob/491da1e365d54e467026641b5984c0dd181ebe90/src/Middleware/WhoopsErrorResponseGenerator.php#L82-L84

```php
        $sendOutput = $this->whoops->writeToOutput();
        $this->whoops->writeToOutput(false);
        $response
            ->getBody()
            ->write($this->whoops->handleException($e));
        $this->whoops->writeToOutput($sendOutput); //Restore to write to output.
```

weierophinney:
@Moln would you be willing to create a pull request based on your suggestion?

Moln:
@weierophinney  Yeah, I've created.

PR #627, opened 2018-07-25
`WhoopsErrorResponseGenerator` change to restore the default output of Whoops

Moln:
For resolve issue #626


weierophinney:
Thanks, @Moln!

PR #631, opened 2018-08-20
Hotfix/docs getting started

xtreamwayz:

During the conversion of v2 to v3 docs the quick start text for creating a hello template was changed in such way that it unclear what code to use for which renderer. This PR adds example code for plates, zend-view and twig.

### DIFF
--- a/docs/book/v3/getting-started/quick-start.md
+++ b/docs/book/v3/getting-started/quick-start.md
@@ -590,18 +590,26 @@ The above looks for a query string parameter "target", and uses its value to
 provide to the template, which is then rendered and returned in an HTML
 response.
 
-Now, let's edit the template file to have the one of the following header lines (use the one for your chosen template renderer):
+Now, let's edit the template file to have the one of the following header lines
+(use the one for your chosen template renderer):
 
-```html
-<!-- plates -->
-<h1>Hello <?= $this->e($target) ?></h1>
+- Plates
+  ```php
+  <!-- plates -->
+  <h1>Hello <?= $this->e($target) ?></h1>
+  ```
 
-<!-- zend-view -->
-<h1>Hello <?= $this->target ?></h1>
+- zend-view
+  ```php
+  <!-- zend-view -->
+  <h1>Hello <?= $this->target ?></h1>
+  ```
 
-<!-- twig -->
-<h1>Hello {{ target }}</h1>
-```
+- Twig
+  ```twig
+  <!-- twig -->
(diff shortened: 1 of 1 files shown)


docs/book/v3/getting-started/quick-start.md
```diff
@@ -590,15 +590,16 @@ The above looks for a query string parameter "target", and uses its value to
 provide to the template, which is then rendered and returned in an HTML
 response.
 
-Now, let's edit the template file to have the following contents:
+Now, let's edit the template file to have the one of the following header lines (use the one for your chosen template renderer):
 
-```php
-<h1>Hello <?= $this->target ?></h1>
-```
+```html
```
froschdesign: PHP was correct, because the interesting part of this code sample is PHP and not HTML.

xtreamwayz: Changed it to php but that doesn't seem to render nice when previewing the markdown file.

docs/book/v3/getting-started/quick-start.md
```diff
@@ -590,15 +590,16 @@ The above looks for a query string parameter "target", and uses its value to
 provide to the template, which is then rendered and returned in an HTML
 response.
 
-Now, let's edit the template file to have the following contents:
+Now, let's edit the template file to have the one of the following header lines (use the one for your chosen template renderer):
 
 ```php
-<h1>Hello <?= $this->target ?></h1>
-```
+// plates
```
froschdesign: Sorry, a misunderstanding. The code example is HTML _and_ PHP (PHTML), but the relevant part is PHP. Therefore we use "```php" in the documentation for this kind of code examples.
It would be better if we can use "```php+html", but I don't know if this work.

So please revert to HTML comment syntax.

xtreamwayz: Reverted back to html.

docs/book/v3/getting-started/quick-start.md
weierophinney: Well, the target is mkdocs, so not really an issue. :)

docs/book/v3/getting-started/quick-start.md
froschdesign: @xtreamwayz 
I found this:

> There has to be at least one <?php ?> tag in the source code, for Prism to detect if it is a HTML page with PHP tags included or simple plain PHP code.

https://github.com/PrismJS/prism/issues/1238#issuecomment-348593975

PR #632, opened 2018-08-24
Use component provided config providers.

kevin2244:
The provided zend-form component config provider can configure the HelperPluginManager in Expressive v2, as in v3.  It is not essential to replace/delegate/extend the HelperPluginManager factory.

     Edited existing documentation


kevin2244:
The change I made was to add the same existing advice as in the V3 version of the docs to the V2 version. It (seemed to me) that the existing V2 instructions implied that you have some extra work to do when you don't need to. I tried and tested this on Expressive v2 with the skeleton and zend-form.

weierophinney:
Thanks, @kevin2244.

PR #633, opened 2018-08-25
improve the "setting-locale w/o" cookbook example

pine3ree:
[x]  Is this related to documentation?

- the current regex allows `/eng-us/`, I believe only `en`, `eng`, `en-us`, `en-US`, `en_us`, `en_US` (well, i would remove the last 2 as well) should be allowed
- the current regex requires a trailing `/` after the locale, we should allow base paths like `/en`, `/de` as well
- the stripped part length is variable and should be computed from $locale (adding 1 for the root `/` char)
- provide an example for using a default locale from app configuration

### DIFF
--- a/docs/book/v3/cookbook/setting-locale-without-routing-parameter.md
+++ b/docs/book/v3/cookbook/setting-locale-without-routing-parameter.md
@@ -40,9 +40,17 @@ class SetLocaleMiddleware implements MiddlewareInterface
 {
     private $helper;
 
-    public function __construct(UrlHelper $helper)
+    private $defaultLocale;
+    private $fallbackLocale = 'en_US';
+
+    const REGEX_LOCALE = '#^/(?P<locale>[a-z]{2,3}|[a-z]{2}[-_][a-zA-Z]{2})(?:/|$)#';
+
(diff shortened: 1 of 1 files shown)


docs/book/v3/cookbook/setting-locale-without-routing-parameter.md
```diff
@@ -51,17 +59,19 @@ class SetLocaleMiddleware implements MiddlewareInterface
...
 
         $locale = $matches['locale'];
         Locale::setDefault(Locale::canonicalize($locale));
         $this->helper->setBasePath($locale);
 
+        $path = substr($path, strlen($locale) + 1);
+
         return $handler->handle($request->withUri(
-            $uri->withPath(substr($path, 3))
+            $uri->withPath(path ?: '/')
```
xtreamwayz: `$uri->withPath($path ?: '/')`

pine3ree: @xtreamwayz :-) thanks

weierophinney:
Thanks, @pine3ree.

PR #634, opened 2018-08-25
individual checks before calling `inject*` methods

pine3ree:
This could be considered just as a tiny refactoring / code-style change: no bugs, no new features

1. don't call an `inject*` method with missing/empty related config values
2. check for empty routes in `injectRoutesFromConfig` before checking is_array and running the loop on an empty array (this make more sense if the method is called outside this delegator, which, now, already checks for emptiness before calling it)

### DIFF
--- a/src/Container/ApplicationConfigInjectionDelegator.php
+++ b/src/Container/ApplicationConfigInjectionDelegator.php
@@ -55,10 +55,10 @@ class ApplicationConfigInjectionDelegator
 
         $config = $container->get('config');
 
-        if (!empty($config['middleware_pipeline'])) {
+        if (! empty($config['middleware_pipeline'])) {
             self::injectPipelineFromConfig($application, (array) $config);
         }
-        if (!empty($config['routes'])) {
+        if (! empty($config['routes'])) {
             self::injectRoutesFromConfig($application, (array) $config);
         }
 
--- a/src/Container/ApplicationConfigInjectionDelegator.php
+++ b/src/Container/ApplicationConfigInjectionDelegator.php
@@ -54,12 +54,13 @@ class ApplicationConfigInjectionDelegator
         }
 
         $config = $container->get('config');
-        if (! isset($config['routes']) && ! isset($config['middleware_pipeline'])) {
-            return $application;
-        }
 
-        self::injectPipelineFromConfig($application, (array) $config);
-        self::injectRoutesFromConfig($application, (array) $config);
+        if (!empty($config['middleware_pipeline'])) {
+            self::injectPipelineFromConfig($application, (array) $config);
+        }
+        if (!empty($config['routes'])) {
+            self::injectRoutesFromConfig($application, (array) $config);
+        }
 
         return $application;
     }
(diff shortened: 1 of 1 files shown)


src/Container/ApplicationConfigInjectionDelegator.php
```diff
@@ -172,7 +173,7 @@ public static function injectPipelineFromConfig(Application $application, array
      */
     public static function injectRoutesFromConfig(Application $application, array $config) : void
     {
-        if (! isset($config['routes']) || ! is_array($config['routes'])) {
+        if (empty($config['routes']) || ! is_array($config['routes'])) {
```
xtreamwayz: I'm not sure about this change. It's a public function and when used directly, you will get an undefined index notice.

pine3ree: Hello @xtreamwayz , why?
empty acts like ! isset() plus not-empty value. I am just using a stronger checking. I just intercept empty arrays (or whatever) as well...

before: if not set or not array return
my-mod: if not set or empty (whatever) or not array return

I am just making the function skip the for loop in case of an empty array

kind regards

xtreamwayz: Huh... for some reason I thought it would through a notification but it doesn't. Ignore my comment :)

pine3ree: @xtreamwayz ,no problem :-)
I was also thinking about storing the resulting locale in a request attribute instead of using `Locale::setDefault()` in order to make it compatible with async-php setups. What do you reckon?

pine3ree: @xtreamwayz , oh ...another thing I forgot... the `$config` var is casted to an array because the `inject*` methods requires it, so I left it alone...but since I now check for key existence `$config` should be an `array` or at least implement `ArrayAccess`. Usually if a `config` service exists in a zend-expressive application, it is an array, so we could also remove those argument castings unless we want to support a non array ArrayAccess configuration object.

kind regards

xtreamwayz: > I was also thinking about storing the resulting locale in a request attribute instead of using Locale::setDefault() in order to make it compatible with async-php setups. What do you reckon?

This is what I do usually. So yes, go ahead.

>  oh ...another thing I forgot... the $config var is casted to an array because the inject* methods requires it, so I left it alone...but since I now check for key existence $config should be an array or at least implement ArrayAccess. Usually if a config service exists in a zend-expressive application, it is an array, so we could also remove those argument castings unless we want to support a non array ArrayAccess configuration object.

Config should always be an array or ArrayAccess object for Aura.Di.

Usually what we do is `$config = $container->get('config') ?? [];`.

pine3ree: @xtreamwayz oh gosh...yes..right....i have a "few" repos open in so many browser tabs and too little sleep...good, I am glad I didn't remove the array castings since we need to support ArrayAccess for Aura.Di. best regards.

weierophinney:
Thanks, @pine3ree.

PR #639, opened 2018-09-28
Add support for lazy loading request handlers

azjezz:

- [x] Are you creating a new feature?
  - [x] Why is the new feature needed? What purpose does it serve?

  Ability to lazy load request handlers 
  ```php
  $app->post('/u/:username', $factory->lazy(App\Handler\ProfileViewHandler::class), 'user.profile');

  ```
  - [x] How will users use the new feature?
  - [x] Base your feature on the `develop` branch, and submit against that branch.
  - [x] Add only one feature per pull request; split multiple features over multiple pull requests
  - [ ] Add tests for the new feature.
  - [ ] Add documentation for the new feature.
  - [ ] Add a `CHANGELOG.md` entry for the new feature.

### DIFF
--- a/src/Middleware/LazyLoadingMiddleware.php
+++ b/src/Middleware/LazyLoadingMiddleware.php
@@ -43,6 +43,11 @@ class LazyLoadingMiddleware implements MiddlewareInterface
     public function process(ServerRequestInterface $request, RequestHandlerInterface $handler) : ResponseInterface
     {
         $middleware = $this->container->get($this->middlewareName);
+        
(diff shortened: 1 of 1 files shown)


xtreamwayz:
@azjezz I'm not sure why you started this PR. FYI, request handlers are only instantiated when a specific route is called. They are not created if the route isn't requested.
```php
$app->post('/u/:username', App\Handler\ProfileViewHandler::class, 'user.profile');
```

azjezz:
@xtreamwayz yes, i noticed that after digging into the code and i found that the middleware container already converts handlers into middlewares so this is kinda useless.

PR #642, opened 2018-10-13
Added PHP 7.3 support

webimpress:
Updated Travis CI configuration to run all tests on PHP 7.3 with lowest/locked/latest dependencies

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,15 @@ matrix:
     - php: 7.2
       env:
         - DEPS=latest
+    - php: 7.3
+      env:
+        - DEPS=lowest
+    - php: 7.3
+      env:
+        - DEPS=locked
+    - php: 7.3
+      env:
+        - DEPS=latest
 
 before_install:
   - if [[ $TEST_COVERAGE != 'true' ]]; then phpenv config-rm xdebug.ini || return 0 ; fi
--- a/composer.json
+++ b/composer.json
@@ -49,7 +49,8 @@
     },
     "conflict": {
         "container-interop/container-interop": "<1.2.0",
-        "zendframework/zend-diactoros": "<1.7.1"
+        "zendframework/zend-diactoros": "<1.7.1",
+        "zendframework/zend-stdlib": "<3.2.1"
     },
     "suggest": {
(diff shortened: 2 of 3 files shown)


xtreamwayz:
To fix the build, you might want to look into this:

http://php.net/manual/en/control-structures.continue.php
php/php-src#3364
https://wiki.php.net/rfc/continue_on_switch_deprecation

webimpress:
@xtreamwayz 

The problem is only with __lowest__ dependencies and problem is already fixed in zend-stdlib 3.2.1.
But here, zend-stdlib is required implicitly:

```console
$ composer why zendframework/zend-stdlib

zendframework/zend-expressive-fastroute  3.0.2   requires  zendframework/zend-stdlib (^2.0 || ^3.1)    
zendframework/zend-http                  2.8.2   requires  zendframework/zend-stdlib (^3.1 || ^2.7.7)  
zendframework/zend-router                3.2.0   requires  zendframework/zend-stdlib (^2.7.7 || ^3.1)  
zendframework/zend-servicemanager        3.3.2   requires  zendframework/zend-stdlib (^3.1)            
zendframework/zend-validator             2.10.2  requires  zendframework/zend-stdlib (^2.7.6 || ^3.1)  
```

Now I think the best solution will be to add conflict with zend-stdlib < 3.2.1 into composer.json and I am going to do in a sec :)

Xerkus:
I disagree with adding zend-stdlib to conflicts.

webimpress:
@Xerkus 

> I disagree with adding zend-stdlib to conflicts.

Why? What the would you suggest? Add it as explicit dependency?

Xerkus:
Sorry, accidentally sent the message.
@webimpress I need to take a look into this. I'll get to it in a bit after I'll make a release of zend-file

Xerkus:
There is no clear solution here.
One option is bump minimum stdlib version in dependencies and then bump minimum on those dependencies in turn.
Another option is to declare conflict.
I don't see what else can be done here. Proper solution would have been to specify upper minor php versions we support instead of open range. No going back on that one tho.

I am favouring first one slightly more as it allows us to avoid conflict on package we don't even use here.

I asked @Ocramius and he is OK with declaring conflict here.

Issue #624, opened 2018-07-10
composer require zendframework/zend-expressive broken


ksemeks:
Provide a narrative description of what you are trying to accomplish.
I'm trying to include zend-expressive in my project, but unsuccessfully.
When i require zend-expressive via console ( composer require zendframework/zend-expressive ), i get the following error: 

```
Your requirements could not be resolved to an installable set of packages.
  Problem 1
    - zendframework/zend-expressive 3.0.0 requires zendframework/zend-httphandlerrunner ^1.0.1 -> satisfiable by zendframework/zend-httphandlerrunner[1.0.1].
    - zendframework/zend-expressive 3.0.1 requires zendframework/zend-httphandlerrunner ^1.0.1 -> satisfiable by zendframework/zend-httphandlerrunner[1.0.1].
    - zendframework/zend-expressive 3.0.2 requires zendframework/zend-httphandlerrunner ^1.0.1 -> satisfiable by zendframework/zend-httphandlerrunner[1.0.1].
    - zendframework/zend-httphandlerrunner 1.0.1 requires psr/http-message-implementation ^1.0 -> no matching package found.
    - Installation request for zendframework/zend-expressive ^3.0 -> satisfiable by zendframework/zend-expressive[3.0.0, 3.0.1, 3.0.2].

Potential causes:
 - A typo in the package name
 - The package is not available in a stable-enough version according to your minimum-stability setting
   see <https://getcomposer.org/doc/04-schema.md#minimum-stability> for more details.
 - It's a private package and you forgot to add a custom repository to find it

Read <https://getcomposer.org/doc/articles/troubleshooting.md> for further common problems.
```

### Code to reproduce the issue

```
composer require zendframework/zend-expressive
```

### Expected results
Expected composer to install the package

### Actual results
Composer returned an error regarding package dependencies

I managed to install the zend-expressive package by adding 
```
"zendframework/zend-expressive": "^3.0.1",
```
directly to the composer.json and running composer update.

JoeBengalen:
It's not broken, you are missing a `psr/http-message-implementation`, e.g. `zendframework/zend-diactoros`

So I guess this would do:
`composer require zendframework/zend-expressive zendframework/zend-diactoros`

weierophinney:
@ksemeks One new tool for you: `composer why-not <package>`. This will give you detailed information on what blocks installation of a package.

That said, your error message indicates the issue:

```text
[...]

PR #643, opened 2018-10-13
zendframework/zend-expressive#624: add installation note

dmanners:
Add note to installation steps with regards to not being supported in existing projects.

This is adding a short note to the main installation and the getting started documentation to note that installation in existing projects is not currently supported. See https://github.com/zendframework/zend-expressive/issues/624#issuecomment-403832747 for details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,10 @@ Additionally, you may optionally want to install a template renderer
 implementation, and/or an error handling integration. These are covered in the
 documentation.
 
+Installation of Expressive in an existing project is not currently supported.
+Installation steps are covered in detail in the documentation. Please check
+for more information.
+
 ## Documentation
 
 Documentation is [in the doc tree](docs/book/), and can be compiled using [mkdocs](http://www.mkdocs.org):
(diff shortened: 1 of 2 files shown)


dmanners:
Happy to update the content or style here if needed.

Xerkus:
Expressive can easily be installed in existing projects, when its dependency is satisfied as indicated by composer:
`- zendframework/zend-httphandlerrunner 1.0.1 requires psr/http-message-implementation ^1.0 -> no matching package found.`
https://packagist.org/providers/psr/http-message-implementation

It is in `suggest` section of `composer.json` but composer does not show those on install problems:
https://github.com/zendframework/zend-expressive/blob/4e551cd318584ddba4fdc529264e0037cac11a6a/composer.json#L56

dmanners:
Hi @Xerkus yes it is possible but I think that @weierophinney wanted to note that it is not support but possible. Happy to update things to make the wording a bit clearer on this point if needed.

Xerkus:
@dmanners I believe you misunderstood him. He was saying we should document how to use it with existing project. Documentation needs quickstart standalone section that covers all steps needed to install and wire expressive without skeleton.

PR #648, opened 2019-01-08
Add note about needing to quote or escape backslashes.

jbh:

It took me awhile to figure out that I needed to escape the backslashes on command line while trying to generate a handler. It seems obvious now that I tried it and it worked, but it'd be nice to have it noted in the documentation regardless.

### DIFF
--- a/docs/book/v3/reference/cli-tooling.md
+++ b/docs/book/v3/reference/cli-tooling.md
@@ -73,7 +73,10 @@ Commands supported include:
 - **`handler:create [options] <handler>`**: Create a request handler named after
   `<handler>`. By default, the command will also generate a factory, register
   both with the application container, and, if a template renderer is
-  discovered, generate a template in an appropriate location.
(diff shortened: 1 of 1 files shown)


docs/book/v3/reference/cli-tooling.md
```diff
@@ -73,7 +73,10 @@ Commands supported include:
 - **`handler:create [options] <handler>`**: Create a request handler named after
   `<handler>`. By default, the command will also generate a factory, register
   both with the application container, and, if a template renderer is
-  discovered, generate a template in an appropriate location.
+  discovered, generate a template in an appropriate location. **Note:**
+  `<handler>` will either need to be quoted or the backslashes escaped.
+  **Example:** `handler:create ModuleName\\Handler\\IndexHandler` or
```
froschdesign: The examples should be turned over, then they would match the description.

jbh: Good eye. :)

PR #658, opened 2019-03-22
Namespace inaccuracy

samlikins:
The namespaces provided in the migration document for the `RouteMiddleware` and `DispatchMiddleware` are incorrect.


weierophinney:
Thanks, @samlikins!

